PR #10, opened 2026-06-22
feat(results): upload to the base_url used for inference (RES-912)

currentlycodinng:
## Problem

`send_results_to_orq` (used by both the red team runner and agent sim) read `ORQ_BASE_URL` from the environment at call time and ignored the `base_url` actually used for inference. Pointing inference at a non-default server (e.g. staging `my.staging.orq.ai`) still uploaded results to `my.orq.ai` unless `ORQ_BASE_URL` was also set separately.

Closes RES-912.

## Change

Results now upload to the same Orq host used for inference.

- `send_results_to_orq` gains an optional `base_url` kwarg; falls back to `ORQ_BASE_URL` env / `https://my.orq.ai` when not supplied (behaviour unchanged when absent).
- New `resolve_results_base_url()` helper in `common/llm_client.py`: prefers the inference client's Orq host (stripping `/v3/router`), ignores non-Orq clients (uploads always go to Orq), falls back to env.
- Red team `_send_cleaned_results` resolves `base_url` once from the inference client and forwards it (both dynamic/hybrid and static call sites).
- `evaluatorq()` core gains a private `_base_url`, forwarded to the upload.
- Agent sim resolves `base_url` from the inference client and threads it through.

## Acceptance criteria

- [x] `send_results_to_orq` accepts an optional `base_url` kwarg
- [x] Red team upload resolves `base_url` once (same logic as inference) and forwards it
- [x] Agent sim upload path does the same
- [x] Fallback behaviour unchanged when `base_url` is not supplied (reads env)

## Tests

New unit tests at every seam: helper resolution, `send_results` override + env fallback, red team runner forwarding, `evaluatorq()` core forwarding, agent sim forwarding. Full suite: 2090 passed, 2 skipped. `basedpyright` clean on changed files; no new `ruff` findings.

### DIFF
--- a/src/evaluatorq/simulation/api.py
+++ b/src/evaluatorq/simulation/api.py
@@ -17,6 +17,7 @@ import uuid
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
+from evaluatorq.common.llm_client import resolve_results_base_url
 from evaluatorq.simulation.types import DEFAULT_EVALUATOR_NAMES, DEFAULT_MODEL
 from evaluatorq.simulation.utils.run_store import auto_save_run, build_simulation_run, write_report
 
@@ -912,8 +913,6 @@ async def _simulate_via_evaluatorq(
 
     # Resolve the Orq host once from the inference client so results upload to
     # the same server used for inference (RES-912); falls back to env.
-    from evaluatorq.common.llm_client import resolve_results_base_url
-
     upload_base_url = resolve_results_base_url(generation_client)
 
     try:
--- a/src/evaluatorq/simulation/api.py
+++ b/src/evaluatorq/simulation/api.py
@@ -910,6 +910,12 @@ async def _simulate_via_evaluatorq(
     start = datetime.now(tz=timezone.utc)
     run_name = evaluation_name or f'simulation-{start.strftime("%Y%m%d-%H%M%S")}-{uuid.uuid4().hex[:8]}'
 
+    # Resolve the Orq host once from the inference client so results upload to
+    # the same server used for inference (RES-912); falls back to env.
+    from evaluatorq.common.llm_client import resolve_results_base_url
+
+    upload_base_url = resolve_results_base_url(generation_client)
+
     try:
         eq_results = await evaluatorq(
             run_name,
@@ -921,6 +927,7 @@ async def _simulate_via_evaluatorq(
             path=orq_results_path,
             _send_results=upload_results,
             _exit_on_failure=exit_on_failure,
+            _base_url=upload_base_url,
         )
     finally:
         # Close the runner, then any target that owns resources (e.g.
--- a/src/evaluatorq/common/llm_client.py
+++ b/src/evaluatorq/common/llm_client.py
@@ -64,6 +64,27 @@ def client_routes_through_orq(client: AsyncOpenAI | None) -> bool:
     return str(base_url).rstrip("/").endswith(ORQ_ROUTER_SUFFIX)
 
 
+def resolve_results_base_url(
+    client: AsyncOpenAI | None = None,
+    *,
+    default_orq_host: str = ORQ_DEFAULT_HOST,
+) -> str:
+    """Resolve the Orq host that results should be uploaded to.
+
+    Prefers the host of the client actually used for inference when it routes
+    through the Orq router — so a staging/custom Orq endpoint receives both the
+    inference traffic and the results — stripping the ``/v3/router`` suffix to
+    get the bare host. A non-Orq client (e.g. direct OpenAI) is ignored: uploads
+    always go to Orq, never to the inference provider. Falls back to
+    ``ORQ_BASE_URL`` / ``default_orq_host`` when no Orq-routed client is given,
+    preserving the original env-based behaviour.
+    """
+    if client_routes_through_orq(client):
+        host = str(client.base_url).rstrip("/")  # pyright: ignore[reportOptionalMemberAccess]
+        return host[: -len(ORQ_ROUTER_SUFFIX)].rstrip("/")
+    return os.environ.get("ORQ_BASE_URL", default_orq_host).rstrip("/")
+
+
 def resolve_llm_client(
     config_client: AsyncOpenAI | None = None,
     *,
@@ -145,4 +166,5 @@ __all__ = [
     "ResolvedClient",
     "client_routes_through_orq",
     "resolve_llm_client",
+    "resolve_results_base_url",
 ]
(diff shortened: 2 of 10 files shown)


src/evaluatorq/simulation/api.py
```diff
@@ -910,6 +910,12 @@ async def _simulate_via_evaluatorq(
     start = datetime.now(tz=timezone.utc)
     run_name = evaluation_name or f'simulation-{start.strftime("%Y%m%d-%H%M%S")}-{uuid.uuid4().hex[:8]}'
 
+    # Resolve the Orq host once from the inference client so results upload to
+    # the same server used for inference (RES-912); falls back to env.
+    from evaluatorq.common.llm_client import resolve_results_base_url
+
+    upload_base_url = resolve_results_base_url(generation_client)
```
KarinaKKarinaK, line 916: Worth being aware of (not a blocker): this passes the raw `generation_client` to the resolver, but for the common sim path that's `None` — the client actually used for inference is built later inside the job by `build_simulation_client(generation_client)`, which constructs an Orq-routed client from `ORQ_BASE_URL` when no client is injected. So in the no-explicit-client case, the upload host is resolved via the **env path here**, while inference resolves its host via the **env inside the job builder**. Same `ORQ_BASE_URL` env on both sides, so today they always agree — but the "upload to the host used for inference" guarantee is coincidental (both read the same env), not because this line sees the real inference client. The seam only truly fires when a caller injects an explicit `generation_client`. If the job builder's host source ever diverges from `ORQ_BASE_URL`, sim uploads could silently mismatch.

Related: `tests/simulation/test_upload_base_url.py` only covers this env-fallback branch (it passes a callable target, no `generation_client`), so the sim client→host resolution isn't actually exercised end-to-end — only the redteam seam and the `resolve_results_base_url` unit test cover the host-stripping path. A sim test that injects an Orq-routed client and asserts the staging host gets stripped would close that gap and lock in the actual RES-912 behavior for the sim path.

KarinaKKarinaK (approved):
Good fix, and the part most likely to bite — does the default (no `base_url`, no Orq client) still go to the same place? — holds: `ORQ_DEFAULT_HOST` is byte-identical to the old hardcoded `https://my.orq.ai`, `send_results_to_orq(base_url=None)` still reads env then that default, and there's no circular import from the new top-level import in `send_results.py` (`common/llm_client` only pulls `os`/`typing` at load, openai is TYPE_CHECKING-only). Suffix stripping is correct incl. the trailing-slash case, non-Orq/None clients fall back to env (uploads never get redirected at OpenAI), `prepared_targets[0].resolved_llm_client` is a real `AsyncOpenAI | None` attribute (no AttributeError), and all four `send_results_to_orq`/`_send_cleaned_results` call sites are wired — none missed. 37 tests pass.

Approving. One thing worth a look + a test-coverage note, neither blocking:

- **Red team "first target" choice** is fine because `resolved_llm_client` is the single attacker/inference client shared across all targets — so `prepared_targets[0]` is representative, not arbitrary. A one-line comment saying so at the hybrid call site would save the next reader the trace I just did.
- Inline note on the sim seam.

Baukebrenninkmeijer:
## Review — RES-912 (results upload to inference base_url)

Reviewed via multi-agent pass (general code review, silent-failure hunt, test coverage). Core logic correct, precedence right (`base_url` kwarg → `ORQ_BASE_URL` → `ORQ_DEFAULT_HOST`), all tests pass. No blockers. A few things worth addressing:

### Important

**1. No log of the resolved upload host.** This PR exists to stop results landing on the wrong server, but nothing logs *which* host was chosen or when the env fallback fired. Debugging "staging results hit production" would have no breadcrumb. Suggest one line in `send_results.py` before the POST: `logger.debug("Uploading results to {}", resolved_base_url)`, and optionally a debug line in `resolve_results_base_url` distinguishing client-host vs env-fallback.

**2. `prepared_targets[0].resolved_llm_client` is the *attacker* client, not the target's inference client.** It's sourced from `_resolve_attacker_llm_client` (`runner.py:885`). This works for the dominant case (single `ORQ_API_KEY` → same host everywhere, and the upload rides that same key), but the comment at `runner.py:2289-2290` saying "attacker/inference client" is misleading. Suggest rewording to something like "attacker LLM client — same Orq host that owns the `ORQ_API_KEY` used for this run." No code change needed, just clarity for future maintainers. Worth a second look to confirm the attacker host is always the intended upload host.

**3. Test gap on the wiring.** `test_forwards_inference_client_base_url` only checks `_send_cleaned_results` passes through a client it's handed — nothing verifies `_run_dynamic_or_hybrid` actually selects `prepared_targets[0]`. A refactor to `[-1]` or the wrong client would pass CI silently. Same gap for `_run_static` (passes `llm_client` directly, `runner.py:2630`) and the simulation path with an *injected* Orq client (only the env-fallback case is tested). Worth adding contract tests at those seams.

### Suggestions

- **Lazy imports** of `resolve_results_base_url` inside function bodies (`runner.py:165`, `simulation/api.py:915`) break the module's top-level import style and there's no circular-import reason. Move to top-level.
- `resolve_results_base_url(inference_client)` is called *inside* the `try` block in `_send_cleaned_results` (`runner.py:198`). A resolution failure would be logged as an upload failure with a misleading "saved locally" message. Move the call before the `try`.

### Strengths

- `resolve_results_base_url` thoroughly unit-tested (client-wins, trailing-slash on both sides, non-Orq fallback, None fallback, env-absent default).
- `TestSendResultsBaseUrl` monkeypatches real `httpx` and asserts the posted URL — catches f-string breakage.
- Non-Orq client (direct OpenAI) correctly ignored, so uploads never leak to the inference provider.
- The Py3.10 `sys.modules` shadow-patch workaround is correct and documented.

Baukebrenninkmeijer (changes requested):
some small things

currentlycodinng:
Addressed:
- **Log the upload host** — `send_results.py` logs `resolved_base_url` before the POST, and `resolve_results_base_url` now logs whether the host came from the inference client or the env/default fallback.
- **Top-level imports** — moved `resolve_results_base_url` to module top in `runner.py` and `simulation/api.py` (no circular import).
- **Resolve before the try** — `_send_cleaned_results` resolves the host before the `try`, so a resolution failure surfaces on its own instead of being mislabelled as an upload failure + 'saved locally'.
- **Comment reword** — the dynamic call site now reads 'attacker LLM client — same Orq host that owns the ORQ_API_KEY for this run', not 'inference'.
- **Test gap** — added the simulation injected-Orq-client case (client host wins over env), which was the one explicitly untested.

On the two red-team caller seams (`_run_dynamic_or_hybrid` picking `prepared_targets[0]`, `_run_static` passing `llm_client`): those need a pipeline-level harness to reach `_send_cleaned_results`, so I left them — the resolve logic, the `_send_cleaned_results` pass-through, and the sim injected case are all unit-covered. Happy to add pipeline contract tests if you'd rather.

Baukebrenninkmeijer (approved):

PR #11, opened 2026-06-22
refactor(simulation): route to_open_responses through canonical openresponses models (RES-833)

currentlycodinng:
## Problem

`simulation/convert.py:to_open_responses()` hand-built raw `dict[str, Any]` for message items, usage, and incomplete-details — duplicating the OpenResponses mapping that the red-team side already routes through `evaluatorq.openresponses` (RES-540). It was the lone holdout.

Closes RES-833.

## Change

Build the message items, usage, and incomplete-details from the canonical typed models in `evaluatorq.openresponses.convert_models` (`Message`, `InputTextContent`, `OutputTextContent`, `Usage`, `IncompleteDetails`) and serialize via `model_dump(mode="json")`. The response envelope dict is unchanged; only the hand-rolled inner dicts are replaced.

Red-team side needs nothing — already migrated by RES-540.

## Tests

- Existing `test_convert.py` (21) still pass — output shape is preserved.
- New `TestTypedModelRoundTrip` (4): every input/output item, the usage block, and incomplete-details round-trip losslessly through the canonical typed models.

`basedpyright` clean; `ruff` clean; full `tests/simulation` suite green (440 passed).

### DIFF
--- a/tests/simulation/test_convert.py
+++ b/tests/simulation/test_convert.py
@@ -1,6 +1,15 @@
 """Tests for SimulationResult → OpenResponses conversion."""
 
 from evaluatorq.contracts import TokenUsage
+from evaluatorq.openresponses.convert_models import (
+    IncompleteDetails as ORIncompleteDetails,
+)
+from evaluatorq.openresponses.convert_models import (
+    Message as ORMessage,
+)
+from evaluatorq.openresponses.convert_models import (
+    Usage as ORUsage,
+)
 from evaluatorq.simulation.convert import to_open_responses
 from evaluatorq.simulation.types import (
     Message,
@@ -192,3 +201,37 @@ class TestToOpenResponses:
         assert response["object"] == "response"
         assert response["id"].startswith("resp_")
         assert isinstance(response["created_at"], int)
+
+
+class TestTypedModelRoundTrip:
+    """RES-833: output is produced by the canonical evaluatorq.openresponses
+    typed models, so every item round-trips through them losslessly."""
+
+    def test_input_items_round_trip_through_typed_message(self):
+        response = to_open_responses(
+            _make_result(
+                messages=[
+                    Message(role="system", content="You are helpful"),
+                    Message(role="user", content="Hello"),
+                    Message(role="assistant", content="Hi!"),
+                ]
+            )
+        )
+        assert response["input"]
+        for item in response["input"]:
+            assert ORMessage.model_validate(item).model_dump(mode="json") == item
+
+    def test_output_items_round_trip_through_typed_message(self):
+        response = to_open_responses(_make_result())
+        assert response["output"]
+        for item in response["output"]:
+            assert ORMessage.model_validate(item).model_dump(mode="json") == item
+
+    def test_usage_round_trips_through_typed_usage(self):
+        response = to_open_responses(_make_result())
+        assert ORUsage.model_validate(response["usage"]).model_dump(mode="json") == response["usage"]
+
+    def test_incomplete_details_round_trip_through_typed_model(self):
+        response = to_open_responses(_make_result(terminated_by=TerminatedBy.max_turns))
+        details = response["incomplete_details"]
+        assert ORIncompleteDetails.model_validate(details).model_dump(mode="json") == details
--- a/tests/simulation/test_convert.py
+++ b/tests/simulation/test_convert.py
@@ -257,3 +257,91 @@ class TestTypedModelRoundTrip:
             "input_tokens_details": {"cached_tokens": 0},
             "output_tokens_details": {"reasoning_tokens": 0},
         }
+
+
+class TestToolCallConversion:
+    """RES-833 follow-up: assistant tool_calls and role='tool' messages map to
+    FunctionCall / FunctionCallOutput output items (mirrors the langchain path)."""
+
+    def test_assistant_tool_call_maps_to_function_call(self):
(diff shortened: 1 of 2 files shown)


tests/simulation/test_convert.py
```diff
@@ -192,3 +201,37 @@ def test_response_shape(self):
...
+        )
+        assert response["input"]
+        for item in response["input"]:
+            assert ORMessage.model_validate(item).model_dump(mode="json") == item
+
+    def test_output_items_round_trip_through_typed_message(self):
+        response = to_open_responses(_make_result())
+        assert response["output"]
+        for item in response["output"]:
+            assert ORMessage.model_validate(item).model_dump(mode="json") == item
```
KarinaKKarinaK, line 228: Worth being aware of: these round-trip tests are self-referential — they validate the output against the same model that produced it, so they'd stay green even if a field were dropped. And the whole change hinges on `logprobs: []` / `annotations: []` surviving the move to the typed models, but no test (here or in the existing `test_output_messages`, which only checks `role`/`type`/`text`) actually asserts those keys are present. So if a future `convert_models` change dropped `logprobs`, CI wouldn't catch it. The field is correct in this PR — I verified it — but I'd lock it in with one exact-equality assertion on a full output item, e.g.:

```python
assert output_msgs[0]["content"][0] == {
    "type": "output_text", "text": "Hi there!", "annotations": [], "logprobs": [],
}
```

That turns the silent gap into a real guard against the exact regression this refactor could have introduced. Optional — same idea applies to the input-message and usage shapes.

KarinaKKarinaK (approved):
Nice consolidation — this matches the RES-540 pattern (route through the canonical `convert_models`, no third shared helper) and fully closes RES-833. I focused on the one thing a refactor like this can silently break: whether `model_dump(mode="json")` emits the exact same shape as the old hand-rolled dicts. It does, in the meaningful sense:

- The high-risk one — `logprobs: []` on assistant output — is preserved: `OutputTextContent` defaults `logprobs` and `annotations` to `[]`, so the dumped item still carries both keys even though the new code doesn't pass `logprobs`.
- `Message` emits exactly `type/id/status/role/content` (no stray None/optional fields), `Usage` nests identically, and `MessageRole` has user/assistant/system so the `MessageRole(msg.role)` coercion can't raise. 25 tests pass.

(The only difference vs before is dict key *order* — `Message` declares `status` before `role` — which is irrelevant for JSON consumers. Not asking you to change it.)

Approving. One optional test-hardening note inline — worth doing while this is fresh.

Baukebrenninkmeijer:
Follow-up from review (RES-833): **add proper `role="tool"` handling to `simulation/convert.py`** — leaving this for @aminaakhmedova.

### Current behavior
The message loop in `to_open_responses()` only branches on `user`/`system` (→ input) and `assistant` text (→ output). `contracts.Message` (which simulation re-exports) also allows `role="tool"` and assistant `tool_calls`, but those fall through both branches and are **silently dropped** (the comment at `convert.py:62` notes the runner does not produce them today).

### Why support it now
The simulation runner not emitting tool turns *yet* is not a reason to drop them at the protocol layer. The OpenResponses shape should be complete so the conversion does not become the blocker the day the runner gains tools. We already do exactly this on the redteam/langchain path.

### It is mechanical — mirror the existing langchain mapping
`integrations/langchain_integration/convert.py:138-185` already maps these, and the canonical models exist in `openresponses/convert_models.py`:

- assistant `tool_calls: list[StrategyToolCall]` → one `FunctionCall` output item per call
  - `call_id = tc.id`, `name = tc.function.name`, `arguments = tc.function.arguments` (already a JSON string), `id = tc.item_id or _generate_item_id("fc")`, `status = completed`
- `role="tool"` message → `FunctionCallOutput` output item
  - `call_id = msg.tool_call_id`, `output = msg.content or ""`, `id = _generate_item_id("fco")`, `status = completed`

No new models, no new protocol work.

### One design decision to make
An assistant turn can carry both text and `tool_calls`. Suggest: emit the text message **plus** a `FunctionCall` per call (they are separate Responses output items); skip the empty text message on a tool-only turn (no content). This keeps every existing test green.

### Tests to add
- assistant with `tool_calls` → literal-pin the `FunctionCall` output item shape
- `role="tool"` message → literal-pin the `FunctionCallOutput` output item shape
- mixed content + `tool_calls` → both items emitted, correct order
- remove/update the now-stale `convert.py:62` comment

---
**Separate, smaller suggestion** (also for whoever picks this up): the new round-trip tests in `TestTypedModelRoundTrip` are partly tautological (an item produced by a model always re-validates against that same model, so a lockstep model change slips through). Worth adding literal-shape pins for the paths that only round-trip today: the input/system message, `incomplete_details`, `content=None` coercion, and asserting system-message `content[0]["type"] == "input_text"`.

currentlycodinng:
Done — added `role="tool"` and assistant `tool_calls` mapping to `to_open_responses()`, mirroring the langchain path: assistant `tool_calls` → one `function_call` per call (`call_id=tc.id`, name/arguments from `tc.function`, `id=tc.item_id or fc_*`), `role="tool"` → `function_call_output`, and an assistant turn with text + tool_calls emits both (tool-only turn skips the empty text message). Removed the stale `convert.py:62` comment. Literal-pin tests added for each shape. 31 convert tests + full sim suite (446) green.

Baukebrenninkmeijer (approved):

PR #12, opened 2026-06-22
refactor(simulation): rename Datapoint to SimulationDatapoint with deprecated alias (RES-911)

currentlycodinng:
## Problem

`evaluatorq.simulation.Datapoint` and the top-level `evaluatorq.DataPoint` differ only by the casing of one letter. When both appear in the same code sample they read like a typo even though both are correct (surfaced in orq-skills PR #39 review).

Closes RES-911.

## Change

- Rename the class to `SimulationDatapoint` in `simulation/types.py`.
- Keep `Datapoint = SimulationDatapoint` as a documented backward-compat alias.
- Export both names from `evaluatorq.simulation` (lazy import map + `__all__`).

## Scope notes

- Internal call sites keep importing the `Datapoint` alias (still valid). Left intentionally to keep this diff small and reviewable — a follow-up can sweep them to the new name.
- The `simulate-agent` skill docs live in the **orq-skills** repo, out of scope for this repo's PR.

## Tests

New `test_datapoint_rename.py`: `SimulationDatapoint` is the canonical name, `Datapoint` resolves to the same object (from both `evaluatorq.simulation` and `...simulation.types`), and the renamed class constructs. Full `tests/simulation` suite green (440 passed). `basedpyright` + `ruff` clean.


src/evaluatorq/simulation/types.py
```diff
@@ -327,6 +327,12 @@ class Datapoint(BaseModel):
     first_message: str
 
 
+# Deprecated alias for backward compatibility. The class was renamed from
+# ``Datapoint`` to ``SimulationDatapoint`` (RES-911) to disambiguate it from the
+# top-level ``evaluatorq.DataPoint``. Prefer ``SimulationDatapoint`` in new code.
+Datapoint = SimulationDatapoint
```
KarinaKKarinaK: Small accuracy nit: this is called a "deprecated alias" in the comment and PR title, but it's a plain assignment — nothing actually emits a `DeprecationWarning`, so it's an alias, not a machine-enforced deprecation. RES-911 doesn't require a warning, so this is fine to leave. But the package already uses runtime `DeprecationWarning` elsewhere (the sync-hooks path), so if you want `Datapoint` usage to actually taper off rather than linger, a `warnings.warn(..., DeprecationWarning)` via module `__getattr__` (or `typing.deprecated` for type-checker signal) would make the word "deprecated" true. Otherwise consider softening the comment to "alias".

KarinaKKarinaK (approved):
LGTM. Clean, minimal rename and the backward-compat story holds up — I checked the parts that would actually break it: `Datapoint` is still in all three spots in `simulation/__init__.py` (`__all__`, the TYPE_CHECKING block, and `_LAZY_IMPORTS`), `SimulationDatapoint` was added alongside, and both resolve through `__getattr__` to the same object (`Datapoint is SimulationDatapoint`). No pydantic fallout — the model isn't in any discriminated union and nothing dispatches on the class name or schema `title`, so the title flipping to `SimulationDatapoint` is inert; a dump/validate roundtrip still works. The ~12 internal call sites that still use the bare `Datapoint` name all resolve via the alias (including `isinstance`/`model_validate`), so deferring that sweep doesn't break anything. New test passes, full `tests/simulation` suite green.

Approving. Two small notes, neither blocking:

- The deferred internal sweep is fine per RES-911 (the alias exists exactly so callers don't have to change), but worth opening a low-pri follow-up to track it — otherwise new code keeps reaching for the old `Datapoint` name and the disambiguation goal never fully lands.
- One wording nit inline on the alias.

Baukebrenninkmeijer:
As Agent Simulation is not officially released yet, I'm in favour of just changing the type and not having an alias with deprecation. I don't think anyone is using this except for us. Wdyt?

Baukebrenninkmeijer (changes requested):
Think we can do it simpler.

currentlycodinng:
Agreed — since Agent Simulation isn't released and only we use it, dropped the alias entirely. Renamed `Datapoint` → `SimulationDatapoint` across the whole simulation package and tests (no `Datapoint = SimulationDatapoint` alias, removed from `__all__`/lazy-imports/TYPE_CHECKING), and the rename test now asserts the old name is gone. Left the redteam files' English-word 'Datapoint' (comments/UI titles) untouched. Rebased onto main; 527 simulation tests pass, whole-repo basedpyright clean.

Baukebrenninkmeijer (approved):

PR #13, opened 2026-06-22
feat(simulation): decouple framework targets from redteam and expose them (RES-907)

currentlycodinng:
## Problem

The four framework integration targets (`OpenAIAgentTarget`, `LangGraphTarget`, `VercelAISdkTarget`, `CallableTarget`) already work with `SimulationRunner`, but:

1. **Not exposed** — they weren't importable from `evaluatorq.simulation` nor reachable from the `eq sim` CLI.
2. **Coupled to redteam** — each `integrations/*/target.py` imported its contracts from `evaluatorq.redteam.contracts`, so importing any integration target dragged in the whole `redteam` package — wrong for a simulation-only user.

Closes [RES-907](https://linear.app/orqai/issue/RES-907/expose-framework-integration-targets-openai-agents-langgraph-vercel-ai).

## Change

* **Decouple:** repoint all 6 `integrations/*/target.py` imports from `evaluatorq.redteam.contracts` → `evaluatorq.contracts` (every name already lives there; verified). Importing an integration target no longer imports `evaluatorq.redteam`.
* **Reword** the "red teaming target" docstrings — these are generic `AgentTarget`s.
* **Expose** the four targets from `evaluatorq.simulation` (lazy import map + `__all__`).
* **CLI:** `eq sim run` already selects a framework target via `--vercel-url` (→ `VercelAISdkTarget`).

## Acceptance criteria

- [X] `from evaluatorq.simulation import OpenAIAgentTarget` (and the other three) resolves
- [X] Importing an integration target does **not** import `evaluatorq.redteam` (asserted in a subprocess test)
- [X] A simulation run driven by an integration target reports non-zero token usage (proves the `target_agent` path)
- [X] `eq sim run` can select a framework target via a flag (`--vercel-url`)

## Tests

New `test_framework_targets_exposed.py`: redteam-isolation (subprocess), exposure of all four targets, and a `CallableTarget` simulation run that aggregates non-zero token usage. Full unit suite green (2083 passed); whole-repo `basedpyright` clean; `ruff` clean on changed files.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/src/evaluatorq/integrations/crewai_integration/target.py
+++ b/src/evaluatorq/integrations/crewai_integration/target.py
@@ -1,4 +1,4 @@
-"""Simulation/red-teaming target wrapper for CrewAI crews.
+"""Simulation/AgentTarget wrapper for CrewAI crews.
 
 CrewAI is the biggest format divergence among the supported frameworks, which
 makes it a good generality stress test for the unified ``AgentTarget`` protocol.
@@ -25,10 +25,11 @@ import re
 from typing import TYPE_CHECKING, Any
 from uuid import uuid4
 
-from evaluatorq.contracts import AgentTarget, Message
-from evaluatorq.redteam.contracts import (
+from evaluatorq.contracts import (
     AgentContext,
     AgentResponse,
+    AgentTarget,
+    Message,
     OutputMessage,
     TextOutputItem,
     TokenUsage,
--- a/src/evaluatorq/simulation/__init__.py
+++ b/src/evaluatorq/simulation/__init__.py
@@ -42,6 +42,10 @@ logger = logging.getLogger(__name__)  # noqa: RUF067
 
 if TYPE_CHECKING:
     from evaluatorq.contracts import AgentTarget, LLMCallConfig, TokenUsage
+    from evaluatorq.integrations.callable_integration import CallableTarget
+    from evaluatorq.integrations.langgraph_integration import LangGraphTarget
+    from evaluatorq.integrations.openai_agents_integration import OpenAIAgentTarget
+    from evaluatorq.integrations.vercel_ai_sdk_integration import VercelAISdkTarget
     from evaluatorq.openresponses.target import OrqResponsesTarget
     from evaluatorq.simulation.adapters import (
         from_chat_completions,
@@ -177,6 +181,10 @@ _LAZY_IMPORTS: dict[str, tuple[str, str]] = {  # noqa: RUF067
         'SimulationCancelledError',
     ),
     'AgentTarget': ('evaluatorq.contracts', 'AgentTarget'),
+    'CallableTarget': ('evaluatorq.integrations.callable_integration', 'CallableTarget'),
+    'LangGraphTarget': ('evaluatorq.integrations.langgraph_integration', 'LangGraphTarget'),
+    'OpenAIAgentTarget': ('evaluatorq.integrations.openai_agents_integration', 'OpenAIAgentTarget'),
+    'VercelAISdkTarget': ('evaluatorq.integrations.vercel_ai_sdk_integration', 'VercelAISdkTarget'),
     'DEFAULT_MODEL': ('evaluatorq.simulation.types', 'DEFAULT_MODEL'),
     'CommunicationStyle': ('evaluatorq.simulation.types', 'CommunicationStyle'),
     'ConversationStrategy': (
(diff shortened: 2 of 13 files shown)


src/evaluatorq/simulation/__init__.py
```diff
@@ -177,6 +181,10 @@
         'SimulationCancelledError',
     ),
     'AgentTarget': ('evaluatorq.contracts', 'AgentTarget'),
+    'CallableTarget': ('evaluatorq.integrations.callable_integration', 'CallableTarget'),
+    'LangGraphTarget': ('evaluatorq.integrations.langgraph_integration', 'LangGraphTarget'),
```
KarinaKKarinaK, line 185: `__getattr__` does a bare `importlib.import_module` with no guard, and langgraph/openai_agents `target.py` import `langchain_core`/`agents` at module top level. So with the optional dep missing, `from evaluatorq.simulation import LangGraphTarget` raises a raw `ModuleNotFoundError: No module named 'langchain_core'` instead of a "pip install the langgraph extra" hint. Same behavior the integrations subpackage already has, so not a blocker — but these are now reachable from the headline `simulation` namespace, so a friendlier message would be worth it. (CallableTarget/VercelAISdkTarget are fine, no heavy deps.)

src/evaluatorq/integrations/crewai_integration/target.py
```diff
@@ -25,10 +25,11 @@
 from typing import TYPE_CHECKING, Any
 from uuid import uuid4
 
-from evaluatorq.contracts import AgentTarget, Message
-from evaluatorq.redteam.contracts import (
+from evaluatorq.contracts import (
```
KarinaKKarinaK, line 28: crewai and pydantic_ai got the same repoint to `evaluatorq.contracts`, but neither is exposed in `simulation.__all__` like the other four. The ticket only asks to surface the 4, so this is correct by design — but the asymmetry (decouple 6, expose 4) isn't obvious to the next reader. A one-line note somewhere that crewai/pydantic are intentionally decoupled-but-not-exposed would save someone a double-take.

KarinaKKarinaK (approved):
Looks good — clean change, mostly docstring rewording + consolidating the import blocks. I checked the risky part: all 8 names the targets now pull from `evaluatorq.contracts` (`AgentContext`, `AgentResponse`, `AgentTarget`, `MemoryStoreInfo`, `Message`, `OutputMessage`, `TextOutputItem`, `TokenUsage`) are actually defined there, and `redteam.contracts` just re-exports the same objects (`is`-identity holds), so repointing the imports is a no-op type-wise. The subprocess redteam-isolation test and the token-usage test are both meaningful — the latter mocks the user-sim/judge to zero so any non-zero total can only come from the target path. Ran the new file locally, 9 passed.

Approving. Two small inline notes below, neither blocking.

One scope note: RES-907's Scope section also mentions `--openai-agent` / `--langgraph` selection flags on `eq sim run`, but this PR only leans on the pre-existing `--vercel-url`. AC4 ("select a framework target via a flag") is met, but openai-agent/langgraph still aren't selectable from the CLI. Fine to do as a follow-up if that's the intent — just flagging it doesn't fully cover the Scope list.

currentlycodinng:
@Baukebrenninkmeijer on the review's scope note about CLI target flags:

- AC4 ("select a framework target via a flag") is covered by the existing `--vercel-url`.
- I've addressed the other inline note: `from evaluatorq.simulation import LangGraphTarget/OpenAIAgentTarget` now raises a clear `pip install 'evaluatorq[langgraph|openai-agents]'` hint when the extra is missing, instead of a bare ImportError.

For `--openai-agent` and `--langgraph`: unlike `--vercel-url` (a plain URL string), these need the CLI to load a Python graph/agent object dynamically (an import path plus construction), so it is more than a thin flag. Do you want that included here, or tracked as a separate follow-up ticket? Fine with me either way.

Baukebrenninkmeijer:
> For --openai-agent and --langgraph: unlike --vercel-url (a plain URL string), these need the CLI to load a Python graph/agent object dynamically (an import path plus construction), so it is more than a thin flag. Do you want that included here, or tracked as a separate follow-up ticket? Fine with me either way.

No, those should use the SDK for that.

src/evaluatorq/integrations/crewai_integration/target.py
Baukebrenninkmeijer: I'm not sure we need to additionally expose this in `simulation.__all__` if we have it here already. @claude would that serve a purpose?

Baukebrenninkmeijer (approved):

PR #14, opened 2026-06-22
docs: LLM-as-a-jury guide + fenced-JSON judge fix (RES-969)

currentlycodinng:
## What

Adds public documentation for the **LLM-as-a-jury** (panel-of-judges) evaluator, which shipped with no docs (RES-969). While testing the docs against the live ORQ router, found and fixed a real bug that was silently dropping jury votes.

## Changes

**Docs (RES-969)**
- `docs/llm-as-a-jury.md` — when to use a jury, quick start, the full `EvaluatorConfig` field table (`judges`, `repetitions`, `replacement_judges`, `min_successful_judges`, `strict_panel`), aggregation / tie / abstention semantics, and how to read the per-result `jury` breakdown and run-level Krippendorff-alpha reliability.
- `examples/redteam/16_llm_as_a_jury.py` — runnable example covering a mixed-provider panel, repetitions, replacements, the threshold, `strict_panel`, and reading the output.
- `mkdocs.yml` — nav entry.

**Fix**
- `src/evaluatorq/common/judge.py` — providers that ignore `response_format=json_object` through the ORQ router (Anthropic, Gemini) wrap valid verdicts in a ```` ```json ```` fence. That failed `model_validate_json` and recorded a `PARSE` failure, silently shrinking the panel and biasing verdicts toward bare-JSON providers. Strip code fences before parsing; bare JSON is untouched.

## Verification

- Model IDs in the docs/example verified live via the ORQ model list.
- Live jury runs against the router: 3-provider unanimous, replacement-on-failure, and inconclusive (`min_successful_judges`) paths all behave as documented.
- Tie path (fail-closed to VULNERABLE) verified via existing unit coverage in `tests/redteam/test_panel_of_judges.py` — could not force a genuine split with live models.
- Fence fix: TDD (3 parametrized cases in `tests/redteam/test_judge.py`, red → green) and re-ran the exact live scenario that previously dropped the Anthropic vote (`succeeded=1` → `succeeded=2`).
- ruff clean, basedpyright 0 errors, full unit suite: 2080 passed / 2 skipped. `mkdocs build --strict` exits 0; examples compile 81/81.

### DIFF
--- a/docs/llm-as-a-jury.md
+++ b/docs/llm-as-a-jury.md
@@ -1,0 +1,216 @@
+# LLM as a Jury
+
+A single judge model is a single point of failure. It can be noisy from one
+call to the next, and it can be biased toward outputs from its own provider
+family. The jury (or panel of judges) replaces that one judge with several,
+runs them together, aggregates their verdicts into one decision, and reports
+how much they agreed.
+
+You can use a jury two ways: as a general evaluator in `evaluatorq()` through
+`llm_jury()`, or inside red teaming through `EvaluatorConfig`. Both share the
+same panel machinery.
+
+## When to use it
+
+- The evaluation is high stakes and you want a verdict that does not rest on
+  one model's opinion.
+- You are judging outputs from the same provider as your usual judge and want
+  to avoid a judge grading its own family.
+- You want a quantitative signal for how much your judges actually agree, so
+  you know when a verdict is solid and when it is contested.
+
+A single judge is cheaper and faster. Reach for a jury when the cost of a
+wrong verdict outweighs the extra calls. A single-judge panel runs with no
+aggregation overhead, so a jury is purely additive.
+
+## Quick start
+
+`llm_jury()` builds an evaluator you drop into the `evaluators=[...]` list of
+`evaluatorq()`. Give it two or more `judges` and it becomes a jury:
+
+```python
+import asyncio
+
+from evaluatorq import DataPoint, evaluatorq, llm_jury
+
+correctness = llm_jury(
+    name="correctness",
+    criteria="The answer is factually correct and directly answers the question.",
+    judges=[
+        "anthropic/claude-sonnet-4-6",
+        "google/gemini-2.5-pro",
+        "mistral/mistral-large-2411",
+    ],
+)
+
+
+async def answer(data: DataPoint, _row: int) -> dict:
+    # Your system under test produces the output to be judged.
+    return {"name": "qa", "output": "Paris is the capital of France."}
+
+
+async def main() -> None:
+    await evaluatorq(
+        "qa-eval",
+        data=[DataPoint(inputs={"question": "What is the capital of France?"})],
+        jobs=[answer],
+        evaluators=[correctness],
+    )
+
+
+asyncio.run(main())
+```
+
+`llm_jury(model="x")` is shorthand for `judges=["x"]` — a single judge, the
+classic LLM-as-a-judge. Pass `judges=[...]` with two or more models to turn it
+into a jury.
+
+!!! tip "Keep the panel odd and mixed-provider"
+    An odd number of judges (3, 5) makes ties rare. A mix of provider families
+    gives you the independence a jury is meant to provide; three judges from the
+    same provider tend to be correlated and add little over one.
+
+## Verdict modes
+
+`verdict_kind` (with `labels`) decides what each judge returns and how `passed`
+is set. It is not inferred from `labels` — pick the mode explicitly.
+
+| Mode | Configure it with | Judge returns | `passed` is |
+| --- | --- | --- | --- |
+| **Boolean** (default) | `verdict_kind="categorical"`, no `labels` | `true` / `false` | the boolean itself |
+| **Labeled** | `verdict_kind="categorical"`, `labels=[...]` | one of `labels` | `verdict in passing_labels` (`None` if `passing_labels` omitted) |
+| **Numeric** | `verdict_kind="numeric"` | a float in `score_range` | `score >= threshold` |
+
+```python
+# Labeled: a fixed rubric, only some labels pass
+tone = llm_jury(
+    name="tone",
+    criteria="Rate the tone of the reply.",
+    judges=["anthropic/claude-sonnet-4-6", "google/gemini-2.5-pro"],
+    labels=["rude", "neutral", "friendly"],
+    passing_labels=["neutral", "friendly"],
+)
+
+# Numeric: a 0-1 score with a pass threshold
+helpfulness = llm_jury(
+    name="helpfulness",
+    criteria="Score how helpful the answer is, 0 to 1.",
+    judges=["anthropic/claude-sonnet-4-6", "google/gemini-2.5-pro"],
+    verdict_kind="numeric",
+    threshold=0.7,
+)
+```
+
+`labels`/`passing_labels` are valid only for `categorical`; passing them with
+`numeric` raises `ValueError`. In labeled mode `passing_labels` must be a subset
+of `labels`; omit it and the verdict is still recorded but `passed` is `None`.
+
+## Panel configuration
+
+| Argument | Default | What it does |
+| --- | --- | --- |
+| `judges` | — | Judge model IDs. Two or more makes it a jury. Mutually exclusive with `model`. |
+| `model` | — | Single-judge shorthand for `judges=[model]`. |
+| `repetitions` | `1` | How many times each judge is asked. The judge takes its own majority before the panel votes, which smooths per-call noise. |
+| `replacement_judges` | `None` | Stand-in models called only when a configured judge fails mechanically. |
+| `min_successful_judges` | `1` | Minimum decisive judges required, otherwise the verdict is **inconclusive**. Must not exceed the panel size. |
+| `threshold` | `0.5` | Numeric mode: `passed` when `score >= threshold`. |
+| `structured_output` | `True` | Use the provider's structured-output API; falls back to a schema-injected `json_object` call for models that reject it. |
+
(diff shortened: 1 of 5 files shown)


examples/redteam/16_llm_as_a_jury.py
```diff
@@ -0,0 +1,106 @@
...
+            # Ask each judge twice and take its own majority before the panel
+            # votes. Smooths out per-call noise from a single judge.
+            repetitions=2,
+            # Stand-ins called only when a configured judge fails mechanically.
+            replacement_judges=["mistral/mistral-large-2411"],
+            # Treat the run as inconclusive unless at least two judges return a
+            # usable verdict. Guards against a verdict resting on one survivor.
+            min_successful_judges=2,
+            # Refuse a judge that shares the target's provider family.
+            strict_panel=True,
```
KarinaKKarinaK, line 53: This example raises before it can run. Target is `gpt-4o` (openai family) and the panel includes `openai/gpt-4o`, so with `strict_panel=True` the same-family guard (`_panel_composition_messages` -> `evaluator.py:113`) turns "judge shares the target's provider family" into a `ValueError`. It trips the exact guard this comment advertises. `compiles 81/81` won't catch it; it only fails at runtime.

Fix (pick one, keep the verified model IDs):
- drop `openai/gpt-4o` from the panel and backfill a non-OpenAI primary — under strict mode no judge should share the OpenAI target's family, which is exactly the lesson here, or
- set `strict_panel=False` if the goal is just to show a mixed panel running.

I'd go with the first so the example demonstrates strict mode correctly instead of working around it.

src/evaluatorq/common/judge.py
```diff
@@ -187,7 +208,7 @@ async def run_judge(
                 extra_kwargs=cfg.extra_kwargs or None,
             )
         raw_content = response.choices[0].message.content or '{}'
-        payload = EvaluatorResponsePayload.model_validate_json(raw_content)
+        payload = EvaluatorResponsePayload.model_validate_json(_strip_code_fences(raw_content))
```
KarinaKKarinaK: The root cause (routed Anthropic/Gemini fencing JSON) isn't judge-specific. Same un-stripped parse pattern + same providers at:
- `redteam/reports/recommendations.py:211-212` (`content or '{}'` -> `json.loads`, identical to the pre-fix judge)
- `redteam/adaptive/orchestrator.py:461,711` (attack-prompt parsing)
- `simulation/generators/persona_generator.py:107`, `scenario_generator.py:159`

Not blocking this PR, but `_strip_code_fences` wants to be shared rather than duplicated later. Worth a follow-up ticket at least.

src/evaluatorq/common/judge.py
```diff
@@ -145,6 +145,27 @@ def build_eval_replacements(
     return {**flat, **nested}
 
 
+def _strip_code_fences(text: str) -> str:
```
KarinaKKarinaK, line 157: Correct, with two rare gaps (no change needed, just flagging so they're known):
- single-line fence (` ```json {...}``` `, no newline) returns unchanged -> still PARSE-fails. Real providers fence on their own line, so fine.
- `inner.rfind('```')` truncates early if a verdict `explanation` itself contains a triple-backtick. Vanishingly rare for a 2-3 sentence verdict.

The graceful no-closing-fence handling (`rfind == -1` -> keep inner) is a nice touch.

tests/redteam/test_judge.py
```diff
@@ -47,6 +47,33 @@ async def test_success_parses_payload(monkeypatch: pytest.MonkeyPatch) -> None:
...
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    'content',
+    [
+        '```json\n{"value": true, "explanation": "resisted"}\n```',
+        '```\n{"value": true, "explanation": "resisted"}\n```',
+        '   ```json\n{"value": true, "explanation": "resisted"}\n```   ',
+    ],
+)
+async def test_fenced_json_is_unwrapped(monkeypatch: pytest.MonkeyPatch, content: str) -> None:
```
KarinaKKarinaK, line 65: Solid TDD coverage. Optional: add a no-closing-fence case (` ```json\n{...} ` with no trailing fence) to lock in that branch.

docs/llm-as-a-jury.md
```diff
@@ -0,0 +1,160 @@
...
+import asyncio
+
+from evaluatorq.redteam import EvaluatorConfig, LLMConfig, OpenAIModelTarget, red_team
+
+
+async def main() -> None:
+    config = LLMConfig(
+        evaluator=EvaluatorConfig(
+            judges=[
+                "openai/gpt-4o",
```
KarinaKKarinaK: nit: this quick-start targets `gpt-4o` with `openai/gpt-4o` as judge `[0]`. It runs (no `strict_panel`), but logs a same-family warning on every call. Since it's the canonical copy-paste snippet, consider a non-OpenAI primary so it's clean out of the box — and it'd reinforce the mixed-provider advice the page gives two paragraphs later.

KarinaKKarinaK (commented):
Inline notes below. Docs check out against source; one blocker in the example and one worthwhile follow-up on the fix.

KarinaKKarinaK:
## Review summary

Docs are accurate — I checked every field, signature, and attribute path against the source (`EvaluatorConfig` fields, `red_team`/`OpenAIModelTarget` signatures, `JuryResult`/`JuryVote`/`JuryReliability` fields, `result.evaluation.jury`, `summary.jury_reliability`). All correct. The fence fix is real and well-tested.

Two things to action (inline):

- 🔴 **`examples/redteam/16_llm_as_a_jury.py` errors out as written.** `strict_panel=True` + an `openai/gpt-4o` judge + a `gpt-4o` target trips the same-family guard and raises a `ValueError` before evaluating — it fails the exact rule its own comment advertises. The "compiles 81/81" check doesn't catch runtime config errors. Drop the OpenAI judge (and backfill a non-OpenAI primary) or set `strict_panel=False`.
- 🟡 **Same fenced-JSON bug at sibling parse sites** (`recommendations.py`, `orchestrator.py`, persona/scenario generators do raw `json.loads` on routed output). Not blocking, but `_strip_code_fences` should be shared rather than duplicated later — worth a follow-up ticket.

Everything else is 🟢: the strip helper is correct (two vanishingly-rare edge cases noted inline, no change needed), tests are good TDD, and the only doc nit is the quick-start snippet emitting a same-family warning on every run.

Nice catch on the silent panel-shrinking bug — that's the kind of thing that quietly biases results.

Baukebrenninkmeijer:
@currentlycodinng can you rebase this on the llm-as-a-jury branch i'm working on and update docs with that?

currentlycodinng:
Rebased onto `feat/llm-jury-evaluator` and retargeted this PR's base to it (now shows a clean diff). Two things resolved in the rebase:

- **judge.py conflict:** kept your structured/`_json_object_judge`/legacy split and re-applied the fenced-JSON strip on both raw-JSON parse paths — the legacy `json_object` path *and* the new `_json_object_judge` fallback (routed Anthropic/Gemini fence even on `json_object`). 66 judge/jury/structured/llm_jury tests pass.
- **mkdocs nav:** merged into your reorganized nav (LLM as a Jury sits under Get Started).

**Docs reworked** to lead with the public `llm_jury()` evaluator for `evaluatorq(evaluators=[...])` — quick start, the three verdict modes (boolean/labeled/numeric), panel config, and reading the output — keeping the red-team `EvaluatorConfig` path as a section. Verified every example constructs against the API; mkdocs builds `--strict` clean.

Note: this is now stacked on your branch, so it'll need a quick re-rebase if `judge.py` or the `llm_jury()` signature shifts before #15 merges.

Baukebrenninkmeijer:
@claude can you update review

Baukebrenninkmeijer (commented):
## Overall

This is a solid PR. The `_strip_code_fences` function addresses a real issue (routed Anthropic/Gemini wrapping JSON in ``` blocks), the raw trace content is correctly preserved at both call sites, and the docs/example are well-written.

## Minor findings

**1. Inner ``` in explanation (low risk)**

`_strip_code_fences` uses `rfind('```')` which can match a triple-backtick inside the JSON content rather than the closing fence. For example:

```
```json
{"explanation": "use ``` for code"}
```
```

`rfind` finds the inner ``` (position ~42) and truncates to `{"explanation": "use ```, losing the rest. In practice explanations rarely contain triple backticks, but if they do the judge silently returns PARSE instead of the verdict.

If you want to be stricter, you could split on the newline-prefixed fence (e.g. `rsplit('\n```', 1)`) to only match the actual closing fence.

**2. Missing test cases**

The test at `tests/redteam/test_judge.py:51` covers 4 good cases. I added 3 more (pushed to this branch):

- **Empty fenced block** (` ```json\n``` `) -> confirms it fails JSON parse
- **Two backticks** (` ``{"value": true} `) -> confirms not treated as fenced
- **Inner triple backticks** -> documents edge-case behavior

All 13 tests pass now.

currentlycodinng:
Rebased onto `main` (now that #15 is merged) and retargeted the base to main — the diff is just the docs/example + the fenced-JSON judge fix now.

On the two findings:
- **Hardened `_strip_code_fences`** to anchor the closing fence on a newline (`rsplit('\n```', 1)`) so a ```` ``` ```` inside an explanation can't truncate the verdict.
- **Kept your edge-case test commit** (cherry-picked, your authorship preserved) and added one regression case for the actual bug the hardening fixes: an inner ```` ``` ```` with *no* closing fence (your inner-backtick case has a closing fence, so the old `rfind` happened to still find the right one). 14 judge tests pass; mkdocs `--strict` clean.

Baukebrenninkmeijer (approved):

PR #15, opened 2026-06-23
feat(jury): llm_jury() evaluator for evaluatorq() with structured generation (RES-848)

Baukebrenninkmeijer:
## What

Adds `llm_jury()` — an LLM-as-a-judge / LLM-as-a-jury evaluator that drops into the `evaluators=[...]` list of `evaluatorq()`, exposing the full judge panel surface that previously only `red_team()` could reach.

```python
from evaluatorq import evaluatorq, llm_jury

correctness = llm_jury(
    name="correctness",
    criteria="Rate the answer's factual correctness vs. the expected output.",
    judges=["openai/gpt-5.5", "anthropic/claude-opus-4-8", "google/gemini-3.5-flash"],
    min_successful_judges=2,
    labels=["correct", "partially_correct", "incorrect"],
    passing_labels=["correct", "partially_correct"],
    aggregator="majority",  # default is "mode" (plurality)
)
await evaluatorq("run", data=[...], jobs=[...], evaluators=[correctness])
```

## How

- **`execute_chat_parse`** (`common/llm_call.py`) — sibling of `execute_chat_completion` using `client.chat.completions.parse` (real structured outputs). The shared `.create` path is untouched.
- **`run_judge`** (`common/judge.py`) gains additive `response_model` / `structured_output` / sentinel `temperature` params. **Tier 1:** `.parse` with a dynamic Pydantic verdict model. **Tier 2 fallback:** `json_object` + schema injection on `BadRequestError` (gated on the error body). `response_model=None` (the default) is byte-identical — red-team and simulation callers are unchanged.
- **`llm_jury.py`** — factory + validation + per-datapoint scorer; reuses the existing `run_jury` aggregation engine. No `redteam` import.

## Surface

`judges`/`model`, `repetitions`, `replacement_judges`, `min_successful_judges`, `verdict_kind` (categorical + numeric), `labels`/`passing_labels`, `threshold`/`score_range`, `aggregator` (keyword or custom callable — see below), `tie_break`, `structured_output`, `temperature` (None-omit, warns at 0.0), `max_tokens=8000`, `extra_kwargs`, `client` (auto-resolved).

## Aggregation

`aggregator=` selects the panel consensus rule, validated against `verdict_kind` (a mismatch raises `ValueError`):

- **categorical** — `"mode"` (default: most common; plurality ties go to `tie_break`) or `"majority"` (strict >50%, else inconclusive)
- **numeric** — `"mean_std"` (default: mean verdict; std reported in `stats` on a conclusive verdict), `"median"`, `"min"`, `"max"`
- **custom** — `Callable[[list[JuryVote]], bool | float | str | None]` for either kind. Receives **all** votes (including abstained/failed) so it can quorum or weight; returns `None` for "no consensus" (inconclusive). The numeric keyword also collapses a single judge's `repetitions`.

Built-in keywords live in a single keyword→fn registry conforming to the same `Aggregator` schema custom callables use; one `_AGG_KIND` map is the source of truth for the keyword↔kind partition.

## Verdict mapping

Categorical → boolean/label per the chosen aggregator; numeric → aggregated per `aggregator`, clamped to `score_range`, `pass = score >= threshold`. Inconclusive (below `min_successful_judges`, or a custom aggregator returning `None`) → `pass=None`. Refusal → abstain.

## Notes

- Probe confirmed `.parse` works through the Orq router for all three default models.
- `openai` floor bumped to `>=1.92.0` (where `.parse` left beta); it is now a **core** dependency (no `[jury]` extra).
- Responses-API support for judges deferred to **RES-972**; per-judge OTEL span attributes to **RES-985**.
- New unit tests cover every aggregator strategy, the custom-callable contract (sees all votes; `None`→inconclusive), the strict-majority 50% boundary, repetition collapse, and registry/partition/literal parity. Full suite green (9 pre-existing `huggingface_hub` failures unrelated).


currentlycodinng (commented):
**Verdict: approve with cleanups + one behavioral question.** Reviewed statically and verified live (checked out the branch: 52 jury/judge unit tests pass, import + factory smoke-tested). The tiered `.parse` → `json_object` fallback, byte-identical default path (`response_model=None`), refusal/parse-none handling, and fail-loud-for-lone-judge are all sound. A re-raised non-schema `BadRequestError` from `execute_chat_parse` is caught safely since it subclasses `APIStatusError`.

**Found by running it — worth a decision:**

1. **`llm_jury()` resolves the LLM client eagerly at factory-call time.** Declaring an evaluator with no `client=` (the path the PR's own README example uses) calls `resolve_llm_client()` immediately, so `correctness = llm_jury(name=..., criteria=...)` raises `MissingLLMCredentialsError` at *definition* time if neither `ORQ_API_KEY` nor `OPENAI_API_KEY` is set. Since evaluators are typically declared at module scope, this couples definition to environment. Passing `client=` sidesteps it, but consider resolving lazily inside the scorer (on first call) so declaring the evaluator never requires credentials.

**Should fix before merge:**

2. **Stray artifact committed.** `.superpowers/sdd/task-4-report.md` is an internal SDD task report. Drop it from the PR.

**Worth confirming / minor:**

3. **`openai` moves from optional extra → core runtime dependency.** `__init__.py` now imports `llm_jury` → `common.judge` → `openai` at module load, so every base install pulls `openai` (floor bumped to `>=1.92.0`). Looks intentional and the pyproject is consistent, but it weights the base install for users who only call `evaluate()`. Confirm the trade and note it in the changelog.

4. **`min_successful_judges` bound vs. panel passed to `run_jury`.** Validation uses the *deduped* panel size, but `run_jury(panel=panel)` gets the *raw* panel, so duplicate judge IDs still execute multiple times. Either dedupe before handing off, or validate against `len(panel)`.

5. **Dead sentinel surface (non-blocking).** `llm_jury` always passes an explicit `temperature` (default `None`), so the `_USE_CFG` "fall back to `cfg.temperature`" path in `run_judge` is never reached from this caller.

currentlycodinng (commented):
**Correction to my earlier review — please disregard point #4.** I claimed `run_jury` receives the raw (non-deduped) panel so duplicate judge IDs would double-vote. That's wrong: `run_jury` calls `resolve_panel(panel)` internally, which dedupes preserving insertion order. So duplicate judges do not execute multiple times, and the factory's `min_successful_judges` validation against the deduped panel size is consistent with `run_jury` — no change needed there. The other points stand.

src/evaluatorq/llm_jury.py
```diff
@@ -0,0 +1,407 @@
...
+    if repetitions < 1:
+        raise ValueError(f"repetitions ({repetitions}) must be >= 1.")
+    if verdict_kind != "categorical" and (labels or passing_labels):
+        raise ValueError("`labels`/`passing_labels` are only valid for verdict_kind='categorical'.")
+    if labels is not None and len(labels) < 2:
+        # A categorical judge with 0/1 options is meaningless, and an empty list
+        # would build a degenerate ``Literal[()]`` verdict model.
+        raise ValueError("categorical `labels` must list at least two options.")
+    if labels and passing_labels and not set(passing_labels) <= set(labels):
+        raise ValueError("`passing_labels` must be a subset of `labels`.")
```
KarinaKKarinaK, line 356: Gap: `passing_labels` is accepted but silently ignored in boolean mode. The guard above only rejects `passing_labels` when `verdict_kind != "categorical"`. But in the default categorical + `labels=None` (boolean) mode, `llm_jury(name=..., criteria=..., passing_labels=["yes"])` is accepted — and at runtime the verdict model is `bool`, so `_to_evaluation_result` takes the `isinstance(verdict, bool)` branch and never looks at `passing_labels`. No crash, but the user's config is silently dropped, which is the misleading kind of silent. Suggest one more guard:

```python
if passing_labels and not labels:
    raise ValueError("`passing_labels` requires `labels` (categorical labeled mode); boolean mode derives pass/fail from the verdict directly.")
```

src/evaluatorq/common/judge.py
```diff
@@ -162,33 +211,119 @@ async def run_judge(
...
+                        span=span,
+                        timeout_s=cfg.timeout_ms / 1000.0,
+                        response_model=response_model,
+                        temperature=temp,
+                        max_completion_tokens=cfg.max_tokens,
+                        extra_kwargs=cfg.extra_kwargs or None,
+                    )
+                except BadRequestError as exc:
+                    err = str(getattr(exc, 'body', None) or getattr(exc, 'message', '') or '').lower()
+                    if not any(k in err for k in ('response_format', 'json_schema', 'schema')):
```
KarinaKKarinaK: Minor: this is a substring match on the error body to decide whether to fall back. It fails safe (a schema rejection whose body doesn't contain `response_format`/`json_schema`/`schema` just propagates as a hard error instead of falling back), and you've got a test for the non-schema case, so I'm fine with it. Just worth a one-line comment that a miss here intentionally fails rather than silently degrades, so the next person doesn't "fix" it into a bare `except BadRequestError`. Also `'schema'` already covers `'json_schema'` as a substring — the list is slightly redundant. Non-blocking.

pyproject.toml
```diff
@@ -156,7 +156,8 @@ dependencies = [
   "httpx>=0.28.1",
   "rich>=14.2.0",
   "loguru>=0.6.0",
-  "typer>=0.12.0"
+  "typer>=0.12.0",
+  "openai>=1.92.0,<3.0.0"
```
KarinaKKarinaK, line 160: This is the line that moves `openai` into core (see body comment #2). Flagging it inline so the footprint change is visible in the diff — every base install now pulls openai>=1.92, not just `[redteam]`/`[simulation]`. Documented in the CHANGELOG; just want an explicit maintainer ack on it.

KarinaKKarinaK (changes requested):
Solid; the structured-output plumbing is well done. Dug into the parts most likely to bite and they hold up: `response_model=None` is genuinely byte-identical to the old `run_judge` (the `_USE_CFG` sentinel resolves to `cfg.temperature`, so red-team/sim callers are untouched), the `.parse` reasoning_effort drop-retry mirrors `execute_chat_completion`, `BadRequestError` is already a runtime import, refusal/`parsed=None` are handled correctly, and the lazy-client `nonlocal` is safe under asyncio (sync resolver, no await between check and assign). `run_jury` does accept `propagate_errors`, and the min-successful bound matches because `run_jury` re-dedups the panel internally. All 53 jury/judge/parse unit tests pass locally.

Requesting changes on two things — one small code fix, one decision I want a maintainer call on — plus a process note:

**1. Ticket linkage.** There's no RES id anywhere — not in the title, branch, commits, or CHANGELOG (our convention is to put it in all of them). I traced the likely parent to RES-760, but that ticket's scope is preference/arena (A-vs-B) judging blocked by RES-759 — consensus winner, per-judge win rates, position bias — none of which this PR delivers (and shouldn't; this is single-target rate-against-criteria). So either this is pointed at the wrong ticket or RES-760 needs reframing. Can you link the actual ticket so the scope lines up? (RES-972 for the deferred Responses-API path is correctly noted.)

**2. `openai` → core dependency.** Worth an explicit sign-off since it widens every base install (anyone calling only `evaluate()` now pulls openai). It's documented in the CHANGELOG and I see it was deliberately flipped back from a `[judge]` extra during review, so I'm not blocking it — just want it to be a conscious, recorded maintainer decision rather than slipping in via the jury feature.

**3.** One real validation gap inline (passing_labels silently ignored). Quick fix.

currentlycodinng (changes requested):
Setting an explicit status to match my review above. **Requesting changes** on two things:

1. Drop the stray `.superpowers/sdd/task-4-report.md` artifact.
2. Decide on the eager client resolution — `llm_jury()` without `client=` raises `MissingLLMCredentialsError` at definition time (the README example breaks with no creds set). Lazy-resolving inside the scorer would fix it.

The openai extra→core dependency move and the changelog note are worth confirming but not blockers. Implementation itself is solid (52 jury/judge tests pass live). Happy to flip to approve once 1 and 2 are settled.

Baukebrenninkmeijer:
Thanks both — addressed the review. Summary of changes since:

**@currentlycodinng**
1. **Eager client resolution** — fixed. The client now resolves lazily on the scorer's first call via a `nonlocal`, so declaring a module-scope evaluator with no `client=` never requires credentials. Added a test asserting `resolve_llm_client` is *not* called at factory time and called once on first scorer invocation. (b0626dc, 465c4d9)
2. **Stray `.superpowers/sdd/task-4-report.md`** — removed; root `.superpowers/` is now gitignored. (b0626dc)
5. Dead `_USE_CFG` path — left as-is intentionally: it's the general default for `run_judge`'s other callers (red-team/sim), only unreachable *from* `llm_jury` since it always passes an explicit `temperature`. Removing it would couple the helper to one caller.

**@KarinaKKarinaK**
3. **`passing_labels` silently ignored in boolean mode** — fixed. Added a guard that rejects `passing_labels` without `labels` with a clear message, plus a test, and documented the requirement in the `llm_jury` docstring. (5256ce3)
- **judge.py substring fallback** — added a comment that a miss intentionally re-raises rather than degrading to a bare `except BadRequestError`, and dropped the redundant `'json_schema'` (`'schema'` already covers it). (5256ce3)

**Maintainer decisions (recorded):**
- **Ticket linkage** — this is **RES-848** (panel-of-judges / jury). Added to the PR title and CHANGELOG. I've prepended a scope-alignment banner to RES-848 reconciling the shipped `llm_jury()` API (unified factory, direct scoring, native structured output) with the original proposal, and flagged the deferred pieces (comparative A/B + position swap, first-class disagreement signal, `instructor`/`[jury]` extra). RES-760 was the wrong parent — it's the preference/arena A-vs-B track, which this PR deliberately does not deliver.
- **`openai` → core dependency** — keeping it as core (conscious maintainer decision). Documented in the CHANGELOG. The native structured-output path replaces `instructor`, and gating `llm_jury` behind an extra would split the install for the package's first built-in LLM evaluator.

Also documented the `verdict_kind` modes (boolean / labeled / numeric) and the pass/fail derivation in the `llm_jury` docstring, which renders into the API reference. CI is green across the 3.10–3.13 matrix.

KarinaKKarinaK (commented):
Re-reviewed after the last round. The two blockers are gone: the client now resolves lazily on first scorer call, so declaring an evaluator with no `client=` no longer needs creds, and `passing_labels` without `labels` raises instead of silently no-opping. Ran the suite on the branch and 83 jury/judge/parse tests pass. RES-848 is linked in the title and changelog. Nice work, the structured-output plumbing and the aggregator registry hold up.

Two small leftovers before merge, both cleanup, neither is logic:

1. `scratchpad/pr15-body.md` is committed (inline comment). It is the same class of stray artifact as the `.superpowers` report we already dropped, just the PR description checked into the repo. Drop it. Note `scratchpad/` is not gitignored, so this will keep happening unless you ignore it too.

2. The PR description's Surface section still lists `strict_panel (reserved no-op)`, but that param is no longer in the code. Good call cutting a reserved no-op. Just update the description so it matches what shipped.

Nothing else blocking from me.

scratchpad/pr15-body.md
```diff
@@ -0,0 +1,51 @@
+## What
```
KarinaKKarinaK: Stray artifact. This is the PR description committed into the repo, same class as the `.superpowers` report removed last round. Drop the file. `scratchpad/` is not in `.gitignore`, so consider ignoring it too to stop the next one.

src/evaluatorq/llm_jury.py
```diff
@@ -0,0 +1,432 @@
...
+    if verdict_kind == "numeric" and not (score_range[0] <= threshold <= score_range[1]):
+        raise ValueError(
+            f"threshold ({threshold}) must lie within score_range {score_range}."
+        )
+    validate_aggregator(
+        aggregator,
+        VerdictKind.NUMERIC if verdict_kind == "numeric" else VerdictKind.CATEGORICAL,
+    )
+    panel = _resolve_panel(judges=judges, model=model)
+    deduped = list(dict.fromkeys([*panel]))
```
KarinaKKarinaK: `[*panel]` is a no-op here. `panel` is already a list from `_resolve_panel`, so `list(dict.fromkeys(panel))` is the same result with less to read.

currentlycodinng (approved):
**Verdict: APPROVE.** Both blockers from my earlier review are addressed, verified live on `6d370b0`:

1. **Eager client resolution → fixed.** `llm_jury()` now does `resolved_client = client` at factory time and resolves lazily inside `scorer` on first call (`nonlocal resolved_client`, llm_jury.py:385/393-396). Confirmed: declaring an evaluator with `ORQ_API_KEY`/`OPENAI_API_KEY` scrubbed from the env no longer raises `MissingLLMCredentialsError` — the README example now works without creds present at definition time.
2. **Stray artifact → removed.** `.superpowers/...` is no longer in the diff (untracked + gitignored).

Jury/judge suite green: **57 passed**. Remaining items are non-blocking and already noted (openai moving extra→core is intentional; the `_USE_CFG` sentinel is dead for this caller). Nice work.

KarinaKKarinaK (approved):
Reviewed the full diff. Looks good — approving.

**What works well**
- Backward compatibility is real, not just claimed. The `response_model=None` path in `run_judge` keeps `.create` + `json_object` + `cfg.temperature`, so red-team and simulation callers are untouched. Good that there are tests pinning this (`test_response_model_none_uses_create_unchanged`).
- The `BadRequestError` fallback is gated on the error body (`response_format`/`schema`), not a bare `except` — and there's a test that a context-length 400 is NOT swallowed into the fallback. That's the right call.
- `structured_output=False` still injects the schema into the system prompt instead of silently dropping the label set. Nice catch, and tested.
- `parsed is None` without a refusal maps to a hard PARSE error rather than a fake `value=None` abstain — easy to get wrong, handled correctly.
- Aggregator registry has the parity test (`_AGGREGATORS` / `_AGG_KIND` / `AggregatorName` literal kept in sync), so dispatch and validation can't drift apart.
- Test coverage is thorough: every aggregator, the strict-majority 50% boundary, repetition collapse, custom-callable contract, lazy client resolution.

**Non-blocking nits**
- `if bool(criteria) == bool(prompt)` uses truthiness, not `is None`. Passing `prompt=""` with a `criteria` set slips past the check and then `template = ""` (empty template), since `"" is not None`. Edge case, unlikely in practice, but `is None` would be tighter.
- Default `temperature=None` in `llm_jury()` means no temperature is sent by default while `run_judge`'s own default is `cfg.temperature` — intentional per the docstring, just worth a reader being aware the factory overrides it.
- `DEFAULT_JUDGE_MODEL = "openai/gpt-5.4-mini"` and the example model slugs — fine since they route through the Orq gateway and the PR notes a `.parse` probe confirmed them, just keep an eye on these as gateway slugs change.

`openai` moving to a core dep is a reasonable tradeoff for `llm_jury()` working without an extra, and it's documented in the CHANGELOG. CI is green. 👍

PR #17, opened 2026-06-24
feat(openresponses): multi-part multi-modal message content (BOPS-839)

currentlycodinng:
## Problem

The Orq Responses API (`POST /v3/router/responses`) accepts multi-part message content — text, images, files in one turn — but our `openresponses` models were text-only, so any caller passing image/file content failed or lost data silently.

Closes BOPS-839.

## Change

In `openresponses/convert_models.py`:
- **`InputImageContent`** — `type="input_image"`, `image_url`, optional `file_id`, `detail` (`auto`/`low`/`high`, default `auto`).
- **`InputFileContent`** — `type="input_file"`, all-optional `file_id` / `file_data` (base64) / `file_url` / `filename` / `mime_type`.
- **`Message.content`** union extended to `list[InputTextContent | InputImageContent | InputFileContent | OutputTextContent]`.

In `openresponses/types.py`:
- Replaced the untyped `MessageDict.content: list[dict[str, Any]]` with a typed `ContentPartDict` union (new `InputTextContentDict` / `InputImageContentDict` / `InputFileContentDict` TypedDicts).

## Acceptance criteria

- [x] `Message(content=[InputTextContent(...), InputImageContent(image_url="https://...")])` validates and serializes correctly
- [x] No regression on existing text-only `Message` callers

## Notes

- Optional fields use the assignment form (`x: T = Field(default=...)`) rather than `Annotated[T, Field(default=...)]`, so type-checkers recognize the defaults (basedpyright doesn't pick up defaults nested in `Annotated`).
- Unblocks **RES-879** (extend `ChatMessage` for multi-modal input), which builds on these content-part models.

## Tests

New `tests/openresponses/test_multimodal_content.py` (7): image/file model validation, `detail` enum, mixed text+image+file messages, round-trip through the union, and a text-only no-regression check. Whole-repo `basedpyright` clean; `ruff` clean on source; full unit suite green (2084 passed).


KarinaKKarinaK (commented):
Checked the content-part shapes against orquesta-web (`ResponsesInputContentPartSchema` in `libs/models/messages/src/schemas/responses-request.schema.ts`). The image model matches exactly; two notes on the file model below.

currentlycodinng:
Addressed the schema findings — verified against orquesta-web `ResponsesInputFileSchema` (request schema for `POST /v3/router/responses`):

- Dropped `mime_type` — it is not in the schema, so a caller setting it would lose data silently.
- Added a validator requiring exactly one of `file_data`/`file_id`/`file_url` (the schema is a union of three single-source variants); an empty file part now fails fast instead of serializing to a server-rejected `{"type": "input_file"}`.

Note: this deviates from the BOPS-839 ticket's written field list, which listed `mime_type` and made the sources all-optional. The acceptance criterion is "serializes correctly to the Orq Responses API payload", so I treated the live schema as authoritative. @Baukebrenninkmeijer flagging in case the ticket spec was intentional — easy to revert if so.

Left the image model as-is (it matches `ResponsesInputImageSchema` exactly), and kept `detail` defaulting to `"auto"` (harmless; server default).

src/evaluatorq/openresponses/convert_models.py
```diff
@@ -48,6 +48,43 @@ class InputTextContent(BaseModel):
...
+    def _omit_none(self, handler: Any) -> dict[str, Any]:
+        return {k: v for k, v in handler(self).items() if v is not None}
+
+
+class InputImageContent(_OmitNoneContent):
+    type: Annotated[
+        Literal["input_image"],
+        Field(description="The type of the input item. Always `input_image`."),
+    ]
+    image_url: Annotated[str, Field(description="The URL of the image (https or data: URI).")]
```
KarinaKKarinaK: Minor, optional: the OpenAPI contract makes `image_url` optional on `input_image` (`file_id` is the alternative source), but it's required here — so a `file_id`-only image can't be expressed. If that path matters, make `image_url` `str | None = Field(default=None, ...)`. Fine to leave as-is if image-by-URL is the only intended case.

KarinaKKarinaK (commented):
Correction: my earlier comments verified against the wrong schema (an internal zod helper, `responses-request.schema.ts`). I deleted them. Re-verified against the authoritative OpenAPI contract for `/v3/router/responses` (`libs/api/api-types/src/lib/responses-api.ts` in orquesta-web). The models are accurate there — `input_file` does carry `mime_type` and `file_id`, all fields optional; `file_id` is valid on `input_image` too. One minor note below.

Baukebrenninkmeijer:
Implemented model in orquesta-web is indeed authoritative. Please do some tests to check things actuall correctly pass through.

KarinaKKarinaK:
## Review summary (corrected)

My earlier review checked the wrong schema — an internal zod helper (`responses-request.schema.ts`) rather than the actual endpoint contract. I've deleted those comments. Re-verified against the OpenAPI contract for `POST /v3/router/responses` (`libs/api/api-types/src/lib/responses-api.ts`).

Against that contract the models are accurate:
- `input_file` carries `file_data`, `file_id`, `file_url`, `filename`, `mime_type` — all optional. `InputFileContent` matches, including `mime_type`.
- `input_image` carries `file_id`, `image_url`, `detail` — `file_id` is valid, as modeled.
- `_OmitNoneContent` is the right approach: the contract's optional fields are absent-vs-null sensitive, and dropping None matches that.
- Tests and the text-only no-regression path look good.

One minor, optional note (inline): the contract makes `image_url` optional on `input_image`, but the model requires it, so a `file_id`-only image isn't expressible. Non-blocking — fine if image-by-URL is the only intended path.

LGTM otherwise.

Baukebrenninkmeijer:
make sure you check the go version of the router.

currentlycodinng:
Reverted my schema-alignment change. I had verified against the same internal zod helper (`responses-request.schema.ts`) you flagged; confirmed against the authoritative OpenAPI contract (`libs/api/api-types/src/lib/responses-api.ts`, `mime_type?: string` on `input_file`, all fields optional) that the **original models were correct** — `mime_type` and all-optional sources restored, no validator. Thanks for the catch.

@Baukebrenninkmeijer re: testing pass-through and the Go router — pass-through against the live API is currently blocked (workspace out of credits across providers), so I can't do an end-to-end send yet. Once credits are back I'll run a real `/v3/router/responses` call with an image+file message and confirm. I'll also diff against the Go router's content-part schema.

Baukebrenninkmeijer:
Blocked until we have access to our Research models again then.

Baukebrenninkmeijer:
Follow-up review note: this PR appears to fully cover the `openresponses` **wire-model/type layer**, but it still does **not** complete native multi-part/multi-modal support end-to-end.

What landed here:
- `InputImageContent`
- `InputFileContent`
- typed `ContentPartDict`
- `openresponses.convert_models.Message.content` widened to accept text/image/file parts

What still looks missing if we want to call this truly done in this PR:
- The main internal conversation type is still text-only: `evaluatorq.contracts.Message.content` is `str | None`, so structured content parts cannot flow through the runtime naturally.
- `OrqResponsesTarget` still consumes that text-first message type and forwards `content` as string/null, so callers do not yet get a native typed path for multipart requests.
- `openresponses/dataset.py` helpers are still string-oriented and flatten list content back to text.
- `simulation/convert.py` and related adapters still always emit `input_text` / `output_text`.
- The new tests validate model serialization, but there is still no end-to-end target/adapter test proving mixed text+image+file content can actually pass through the normal evaluatorq/openresponses path without raw-dict escape hatches.

One schema/support gap still worth checking while this PR is open:
- `InputImageContent` currently requires `image_url`. If the authoritative API contract allows a `file_id`-only image part, we should support that too.

I added the same scope note to `BOPS-839` so the ticket reflects the remaining work. My recommendation is to expand this PR to cover the internal message/adapter path too, otherwise we should treat this PR as "model-layer support only" rather than "native multi-part integration".

currentlycodinng:
Made `image_url` optional on `InputImageContent` (file_id is the alternative source per the authoritative contract), so a `file_id`-only image part is now expressible — added a test.

On scope: agreed this PR is the **model-layer foundation** (wire models + typed `ContentPartDict`), not the end-to-end native path. Widening `contracts.Message.content` from `str` to multi-part, threading it through `OrqResponsesTarget`/adapters/`dataset.py`, and the end-to-end target test are exactly the internal-runtime work in **RES-879** (which builds on these models). My suggestion is to land this as the model layer and do the runtime path in RES-879 rather than balloon this PR — but happy to expand it here if you'd rather have it all in one. The live pass-through check is still blocked until Research-model credits are back; I'll run a real `/v3/router/responses` call with a text+image+file message then.

Baukebrenninkmeijer (approved):

PR #21, opened 2026-06-26
fix(redteam): enrich ORQ agent tools with real description + parameter schema

Baukebrenninkmeijer:
## Problem

Red team output showed tool entries like `{"name":"Current Date & Time","description":null,"parameters":null}`. The agent-retrieve endpoint (`RetrieveAgentRequestTools`) only returns a tool **binding** — `id, key, display_name, description, action_type, tool_id` — with **no JSON schema field at all**, and binding-level descriptions are null for built-in tools. The old code in `ORQAgentTarget.get_agent_context` hardcoded `parameters=None` and read only the binding description, so schema was never captured for any tool.

## Fix

- New `_enrich_tool(binding)` resolves each binding's `tool_id` via `tools.retrieve` to pull the underlying tool definition's real `description` and parameter schema (`function.parameters` or `json_schema.schema_`, coerced to `dict` via `model_dump()`).
- Bindings without a `tool_id` (built-ins like *Current Date & Time*) degrade to binding fields — that data genuinely isn't available on that endpoint.
- Runs inside the existing `asyncio.gather` enrichment block (alongside knowledge-base / memory-store enrichment) and fails gracefully with a warning.
- Adds `action_type` to `ToolInfo`.

## Tests

`tests/redteam/test_orq_tool_enrichment.py` — function tool, json_schema tool, no-`tool_id` degrade, and retrieve-failure degrade. All pass; full suite 1429 passing, basedpyright clean.

### DIFF
--- a/src/evaluatorq/redteam/backends/orq.py
+++ b/src/evaluatorq/redteam/backends/orq.py
@@ -370,17 +370,8 @@ class ORQAgentTarget(AgentTarget):
             agent_key=self.agent_key,
         )
 
-        tools: list[ToolInfo] = []
         settings = getattr(agent_data, 'settings', None)
-        if settings and hasattr(settings, 'tools') and settings.tools:
-            tools.extend(
-                ToolInfo(
-                    name=getattr(tool, 'key', None) or getattr(tool, 'display_name', None) or tool.id,
-                    description=getattr(tool, 'description', None),
-                    parameters=None,
-                )
-                for tool in settings.tools
-            )
+        raw_tools = list(settings.tools) if settings and getattr(settings, 'tools', None) else []
 
         raw_kb_ids: list[str] = []
         if hasattr(agent_data, 'knowledge_bases') and agent_data.knowledge_bases:
@@ -392,10 +383,12 @@ class ORQAgentTarget(AgentTarget):
 
         enrichment_tasks: list[Any] = [self._enrich_knowledge_base(kb_id) for kb_id in raw_kb_ids]
         enrichment_tasks.extend(self._enrich_memory_store(ms_id) for ms_id in raw_ms_ids)
+        enrichment_tasks.extend(self._enrich_tool(t) for t in raw_tools)
 
         enriched_results = await asyncio.gather(*enrichment_tasks) if enrichment_tasks else []
         knowledge_bases = [r for r in enriched_results if isinstance(r, KnowledgeBaseInfo)]
         memory_stores = [r for r in enriched_results if isinstance(r, MemoryStoreInfo)]
+        tools = [r for r in enriched_results if isinstance(r, ToolInfo)]
 
         model_raw = getattr(agent_data, 'model', None)
         model_id = getattr(model_raw, 'id', None) if model_raw is not None else None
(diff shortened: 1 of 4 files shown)


src/evaluatorq/redteam/backends/orq.py
```diff
@@ -443,6 +436,41 @@ async def _enrich_memory_store(self, ms_key: str) -> MemoryStoreInfo:
...
+                js = getattr(rich, 'json_schema', None)
+                if fn is not None:
+                    description = getattr(fn, 'description', None) or description
+                    raw_params = getattr(fn, 'parameters', None)
+                elif js is not None:
+                    description = getattr(js, 'description', None) or description
+                    raw_params = getattr(js, 'schema_', None)
+                else:
+                    description = getattr(rich, 'description', None) or description
+                    raw_params = None
```
currentlycodinng, line 468: Non-blocking: only `function` and `json_schema` get a parameter schema here; `http`/`mcp`/`code` tools fall through to description-only with `parameters=None`. Fine for the bug this fixes, just worth a comment noting those three are intentionally partial so a future reader doesn't read it as an oversight.

src/evaluatorq/redteam/backends/orq.py
```diff
@@ -443,6 +436,41 @@ async def _enrich_memory_store(self, ms_key: str) -> MemoryStoreInfo:
...
+        action_type = getattr(binding, 'action_type', None)
+        parameters = None
+        tool_id = getattr(binding, 'tool_id', None)
+        if tool_id:
+            try:
+                rich = await asyncio.to_thread(self.orq_client.tools.retrieve, tool_id=tool_id)
+                fn = getattr(rich, 'function', None)
+                js = getattr(rich, 'json_schema', None)
+                if fn is not None:
+                    description = getattr(fn, 'description', None) or description
```
currentlycodinng: Non-blocking nit: when the inner `function.description` is null this keeps the binding description but skips the tool-level top `description` on the retrieved object. `getattr(fn, 'description', None) or getattr(rich, 'description', None) or description` would be a touch more thorough.

currentlycodinng (approved):
Verified live: checked out the branch and ran it. The 4 new tests pass, the full redteam suite is green (1192 passed), and basedpyright is clean on the changed files. I also confirmed against orq_ai_sdk 4.4.7 that every SDK shape this relies on is real — `tools.retrieve(tool_id=...)` returns the 5-way tool union, `FunctionTool.function` exposes `description`/`parameters`, `JSONSchemaTool.json_schema` exposes `description`/`schema_`, and the `RetrieveAgentRequestTools` binding carries both `tool_id` and `action_type`. Order is preserved and failures degrade cleanly. LGTM.

Two minor non-blocking notes inline.

Baukebrenninkmeijer:
Addressed both review comments in commit 9f46c71:

1. **Description fallback** (line 459/462) — now chains `getattr(fn, "description", None) || getattr(rich, "description", None) || description` so the tool-level top description is also checked as a middle fallback when the inner function/json_schema description is null.

2. **Parameter schema gap** (line 464–466) — added a comment explaining that `http`, `mcp`, and `code` tools intentionally return description-only (no parameter schema) as a known limitation, not an oversight.

PR #23, opened 2026-06-26
feat(contracts): multi-modal Message content for chat completions (RES-879)

currentlycodinng:
## What

Extends `contracts.Message` so chat-completion messages can carry multi-modal input.

- `content` is now `str | list[ContentPart] | None`, where `ContentPart` is the Responses-API `InputTextContent | InputImageContent | InputFileContent`. The string shorthand still validates and round-trips unchanged.
- Adds the `developer` role.
- `to_chat_completion()` renders a multi-part list into Chat-Completions content blocks (`text` / `image_url`).

## Targets, by capability

- **Responses targets** (`OrqResponsesTarget`, `OpenAIAgentTarget`) pass multi-part content straight through as Responses input content parts.
- **Chat/string targets** (`LangGraph`, `Vercel`, `ORQAgent`) flatten via `content_to_text`, which raises `NotImplementedError` on image/file parts instead of silently dropping them.
- Report/transcript paths coerce best-effort via `coerce_content_text`.

## Tests

- Contract tests: developer role, multipart validation, `to_chat_completion` rendering, `OrqResponsesTarget` pass-through serialization, convert-or-raise.
- Live vision integration tests (base64 + https image) gated on `ORQ_API_KEY` — currently credit-blocked in the workspace, so they ship but are skipped in the default run.

## Notes

Stacked on #17 (BOPS-839). Review/merge that first.

### DIFF
--- a/src/evaluatorq/common/messages.py
+++ b/src/evaluatorq/common/messages.py
@@ -12,17 +12,23 @@ def coerce_content_text(content: Any) -> str:
     ``[{"type": "text", "text": "..."}]``) surfaces the joined text rather than a
     Python ``repr`` of the list. ``None`` becomes ``""``; plain strings (and anything
     else) pass through ``str``.
+
+    Unlike :func:`evaluatorq.contracts.content_to_text`, this best-effort helper
+    does not raise on non-text parts (it is used in report/transcript rendering).
+    Image and file parts are surfaced as a ``[image]`` / ``[file]`` placeholder so
+    they are visibly accounted for rather than silently dropped.
     """
     if isinstance(content, list):
         texts: list[str] = []
         for part in content:
-            if isinstance(part, dict):
-                # Both the chat-completions ("text") and Responses ("input_text")
-                # shapes carry their text under a "text" key.
-                if part.get("type") in ("text", "input_text"):
-                    texts.append(part.get("text", ""))
-            elif getattr(part, "type", None) == "input_text":
-                # An InputTextContent pydantic content part.
-                texts.append(getattr(part, "text", ""))
+            part_type = part.get("type") if isinstance(part, dict) else getattr(part, "type", None)
+            # Both the chat-completions ("text") and Responses ("input_text")
+            # shapes carry their text under a "text" key.
+            if part_type in ("text", "input_text"):
+                texts.append(part.get("text", "") if isinstance(part, dict) else getattr(part, "text", ""))
+            elif part_type in ("image_url", "input_image"):
+                texts.append("[image]")
+            elif part_type in ("file", "input_file"):
+                texts.append("[file]")
         return "\n".join(texts)
     return str(content or "")
--- a/src/evaluatorq/common/messages.py
+++ b/src/evaluatorq/common/messages.py
@@ -14,9 +14,15 @@ def coerce_content_text(content: Any) -> str:
     else) pass through ``str``.
     """
     if isinstance(content, list):
-        return "\n".join(
-            part.get("text", "")
-            for part in content
-            if isinstance(part, dict) and part.get("type") == "text"
-        )
+        texts: list[str] = []
+        for part in content:
+            if isinstance(part, dict):
+                # Both the chat-completions ("text") and Responses ("input_text")
+                # shapes carry their text under a "text" key.
+                if part.get("type") in ("text", "input_text"):
+                    texts.append(part.get("text", ""))
+            elif getattr(part, "type", None) == "input_text":
+                # An InputTextContent pydantic content part.
+                texts.append(getattr(part, "text", ""))
+        return "\n".join(texts)
     return str(content or "")
--- a/src/evaluatorq/common/messages.py
+++ b/src/evaluatorq/common/messages.py
@@ -15,8 +15,9 @@ def coerce_content_text(content: Any) -> str:
 
     Unlike :func:`evaluatorq.contracts.content_to_text`, this best-effort helper
     does not raise on non-text parts (it is used in report/transcript rendering).
-    Image and file parts are surfaced as a ``[image]`` / ``[file]`` placeholder so
-    they are visibly accounted for rather than silently dropped.
+    Image and file parts are surfaced as a ``[image]`` / ``[file]`` placeholder,
+    and any other (unknown/future) part type as ``[<type>]``, so every part is
+    visibly accounted for rather than silently dropped.
     """
     if isinstance(content, list):
         texts: list[str] = []
@@ -30,5 +31,9 @@ def coerce_content_text(content: Any) -> str:
                 texts.append("[image]")
             elif part_type in ("file", "input_file"):
                 texts.append("[file]")
+            else:
+                # Unknown/future part shapes (e.g. audio, output_text) are still
+                # surfaced as a placeholder rather than vanishing silently.
+                texts.append(f"[{part_type or 'unknown'}]")
         return "\n".join(texts)
     return str(content or "")
--- a/src/evaluatorq/contracts.py
+++ b/src/evaluatorq/contracts.py
@@ -35,6 +35,8 @@ def _content_part_to_chat_block(part: ContentPart) -> dict[str, Any]:
             image_url['url'] = part.image_url
         image_url['detail'] = part.detail
         return {'type': 'image_url', 'image_url': image_url}
+    if not isinstance(part, InputFileContent):
+        raise NotImplementedError(f'Unsupported content part: {part.type!r}')
     file_obj: dict[str, Any] = {}
     if part.file_id is not None:
         file_obj['file_id'] = part.file_id
--- a/src/evaluatorq/contracts.py
+++ b/src/evaluatorq/contracts.py
@@ -11,6 +11,71 @@ from typing import TYPE_CHECKING, Annotated, Any, Literal, TypeAlias
 from loguru import logger
 from pydantic import AliasChoices, BaseModel, ConfigDict, Field, model_serializer, model_validator
 
+from evaluatorq.openresponses.convert_models import (
+    InputFileContent,
+    InputImageContent,
+    InputTextContent,
+)
+
+# A single part of multi-modal message content (Responses-API input shapes).
+ContentPart: TypeAlias = InputTextContent | InputImageContent | InputFileContent
+
+
+def _content_part_to_chat_block(part: ContentPart) -> dict[str, Any]:
+    """Render a Responses-style content part as an OpenAI chat-completions block.
+
+    Images and files use the chat-completions shapes (``image_url`` / ``file``),
+    which differ from the Responses ``input_image`` / ``input_file`` wire shapes.
+    """
+    if isinstance(part, InputTextContent):
+        return {'type': 'text', 'text': part.text}
+    if isinstance(part, InputImageContent):
+        image_url: dict[str, Any] = {}
+        if part.image_url is not None:
+            image_url['url'] = part.image_url
+        image_url['detail'] = part.detail
+        return {'type': 'image_url', 'image_url': image_url}
+    file_obj: dict[str, Any] = {}
+    if part.file_id is not None:
+        file_obj['file_id'] = part.file_id
+    if part.file_data is not None:
+        file_obj['file_data'] = part.file_data
+    if part.filename is not None:
+        file_obj['filename'] = part.filename
+    return {'type': 'file', 'file': file_obj}
+
+
+def _render_chat_content(content: str | list[ContentPart] | None) -> str | list[dict[str, Any]]:
+    """Render message content for the chat-completions API: a plain string, or a
+    list of content blocks for multi-modal content."""
+    if isinstance(content, list):
+        return [_content_part_to_chat_block(p) for p in content]
+    return content or ''
+
+
+def content_to_text(content: str | list[ContentPart] | None) -> str:
+    """Flatten message content to plain text.
+
+    For targets that accept only text: a string (or ``None``) passes through; a
+    multi-part list is joined from its text parts, but a non-text part (image or
+    file) raises :class:`NotImplementedError` with a clear message rather than
+    silently dropping content.
+    """
+    if content is None:
+        return ''
+    if isinstance(content, str):
+        return content
+    out: list[str] = []
+    for part in content:
+        if isinstance(part, InputTextContent):
+            out.append(part.text)
+        else:
+            raise NotImplementedError(
+                f'This target accepts only text content; got a {part.type!r} part. '
+                'Use a Responses-API target (OrqResponsesTarget) for image/file content.'
+            )
+    return ''.join(out)
+
 if sys.version_info >= (3, 11):
     from enum import StrEnum
 else:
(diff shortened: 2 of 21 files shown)


Baukebrenninkmeijer:
## Code Review: Multi-modal Message Content for Chat Completions (RES-879)

Great work on this feature! Multi-modal support is a significant addition. Here are my findings organized by priority:

---

### High Priority (Must Fix Before Merge)

**1. Integration tests verify the HTTP request succeeded, not that images actually flowed through**

`tests/integration/test_orq_responses_target_integration.py:91-111` and `:112-130` assert only `assert r.text` -- a truthy check that passes regardless of whether the model received the image. The prompt asks "Reply with the single word RED if you can see the image" but the test never checks for this.

```python
# Currently (too weak):
assert r.text

# Should also verify the model actually saw the image:
assert "red" in r.text.lower()
```

**2. `_content_part_to_chat_block` handles `InputFileContent` by accident, not by design**

The function checks `InputTextContent` and `InputImageContent` explicitly, then falls through to the `file_obj` handling. An `InputFileContent` part reaches that block by coincidence. Add an explicit `elif isinstance(part, InputFileContent):` for clarity and future maintainability.

---

### Important (Should Fix)

**3. Integration test depends on a live external URL (Wikimedia)**

```python
_HTTPS_IMAGE_URL = "https://upload.wikimedia.org/wikipedia/commons/thumb/..."
```

Wikimedia could change, rate-limit, or remove this URL at any time, making the test flaky. Consider using the base64 data URL for both tests or pinning to a local fixture.

**4. `test_content_to_text_text_only` misses the empty-list boundary**

The function handles `content=[]` (returns `""`), but this is not tested. Add:

```python
assert content_to_text([]) == ""
```

**5. `coerce_content_text` silently drops non-text content**

Lines 20-26 in `messages.py` ignore `InputImageContent` and `InputFileContent` parts -- they are silently dropped rather than raising. The new `content_to_text()` raises on non-text, but `coerce_content_text()` used in report generation drops silently. This inconsistency could hide issues in report generation paths.

---

### Suggestions (Nice to Have)

**6. No test exercises `InputFileContent`**

The `ContentPart` type alias includes `InputFileContent`, but no test creates a `Message` with it or verifies serialization. Add at least one test case.

**7. Extract test fixtures**

`test_content_accepts_multipart_list` and `test_to_chat_completion_renders_multipart` both construct identical `InputTextContent`/`InputImageContent` pairs inline. A shared fixture would reduce duplication.

**8. `test_orq_responses_target_passes_multipart_through` calls a private method**

```python
items = OrqResponsesTarget._messages_to_input([m])
```

Consider testing via the public `respond()` API instead, or accept this as an implementation-detail test with a comment noting the coupling risk.

**9. `test_developer_role_accepted` only tests model validation**

The test confirms Pydantic accepts `"developer"` as a role, but does not test that it works through `to_chat_completion()` or the target. Consider adding a round-trip assertion.

---

### Strengths

- `content_to_text()` raises `NotImplementedError` on non-text content instead of silently dropping it -- a significant improvement
- The `developer` role addition is clean and follows the existing `Literal` pattern
- Unit test coverage for serialization, type coercion, and error raising is solid
- Consistent use of `content_to_text()` across all integration targets (CrewAI, LangGraph, PydanticAI, Vercel AI SDK, OpenAI Agents, ORQ, Callable)
- The `_OmitNoneContent` serializer correctly omits `null` values from serialization

---

**Overall: Conditional Approve** -- fix the two High Priority items before merging. The integration tests must verify images actually reach the model, not just that the HTTP request succeeded.

src/evaluatorq/redteam/backends/orq.py
```diff
@@ -43,7 +43,7 @@ def _get_orq_server_url() -> str:
     return url.rstrip('/').removesuffix('/v3/router')
 
 
-from evaluatorq.contracts import AgentTarget, Message
+from evaluatorq.contracts import AgentTarget, Message, content_to_text
 from evaluatorq.redteam.backends._errors import extract_provider_error_code, extract_status_code
```
Baukebrenninkmeijer, line 48: **CI blocker (ruff I001).** These module imports sit after the `_get_orq_server_url` def; ruff reports `I001` (auto-fixable) and `uv run ruff check src` fails. This PR extended the mis-placed block by adding `content_to_text`. If the deferred placement is deliberate (circular-import avoidance), add `# noqa: E402` with a one-line note; otherwise move them to the top with the other imports.

src/evaluatorq/contracts.py
```diff
@@ -11,6 +11,73 @@
...
+    out: list[str] = []
+    for part in content:
+        if isinstance(part, InputTextContent):
+            out.append(part.text)
+        else:
+            raise NotImplementedError(
+                f'This target accepts only text content; got a {part.type!r} part. '
+                'Use a Responses-API target (OrqResponsesTarget) for image/file content.'
+            )
+    return ''.join(out)
```
Baukebrenninkmeijer, line 94: **CI blocker (ruff E305):** only one blank line before the following `if sys.version_info >= (3, 11):` block — needs two. Auto-fixable.

Minor: the docstring says the list is "joined from its text parts", but this uses `''.join` (no separator), whereas `coerce_content_text` uses `"\n".join`. Suggest "concatenated (no separator)" so the two helpers' behavior is unambiguous.

src/evaluatorq/contracts.py
```diff
@@ -11,6 +11,73 @@
...
+    Images and files use the chat-completions shapes (``image_url`` / ``file``),
+    which differ from the Responses ``input_image`` / ``input_file`` wire shapes.
+    """
+    if isinstance(part, InputTextContent):
+        return {'type': 'text', 'text': part.text}
+    if isinstance(part, InputImageContent):
+        image_url: dict[str, Any] = {}
+        if part.image_url is not None:
+            image_url['url'] = part.image_url
+        image_url['detail'] = part.detail
```
Baukebrenninkmeijer: **`file_id`-only image renders an invalid block.** `InputImageContent` allows `image_url=None` with `file_id` set (documented in `convert_models.py`). When `image_url is None`, this produces `{"type": "image_url", "image_url": {"detail": "auto"}}` — no `url`, and `file_id` is silently dropped. Chat-completions will reject/ignore it with no signal to the caller.

Suggest guarding the unrepresentable case explicitly, mirroring the existing non-text `NotImplementedError`:
```python
if part.image_url is None:
    raise NotImplementedError(
        'file_id-backed images are Responses-API only; use OrqResponsesTarget.'
    )
```

src/evaluatorq/contracts.py
```diff
@@ -11,6 +11,73 @@
...
+        return {'type': 'image_url', 'image_url': image_url}
+    if not isinstance(part, InputFileContent):
+        raise NotImplementedError(f'Unsupported content part: {part.type!r}')
+    file_obj: dict[str, Any] = {}
+    if part.file_id is not None:
+        file_obj['file_id'] = part.file_id
+    if part.file_data is not None:
+        file_obj['file_data'] = part.file_data
+    if part.filename is not None:
+        file_obj['filename'] = part.filename
```
Baukebrenninkmeijer, line 61: **`file_url` and `mime_type` silently dropped.** `InputFileContent` has five fields (`file_id`, `file_data`, `file_url`, `filename`, `mime_type`); only three are copied. A `file_url`-only file renders `{"type": "file", "file": {}}` — empty block, opaque API error downstream. Dropping `file_url` is correct for the chat shape (no slot for it), but it should fail loudly rather than emit an empty block:
```python
if part.file_id is None and part.file_data is None:
    raise NotImplementedError(
        'file_url-backed files are Responses-API only; use OrqResponsesTarget.'
    )
```
A one-line comment noting `file_url`/`mime_type` are intentionally Responses-only would also prevent a future naive "fix".

src/evaluatorq/contracts.py
```diff
@@ -11,6 +11,73 @@
...
 from pydantic import AliasChoices, BaseModel, ConfigDict, Field, model_serializer, model_validator
 
+from evaluatorq.openresponses.convert_models import (
+    InputFileContent,
+    InputImageContent,
+    InputTextContent,
+)
+
+# A single part of multi-modal message content (Responses-API input shapes).
+ContentPart: TypeAlias = InputTextContent | InputImageContent | InputFileContent
```
Baukebrenninkmeijer: Design nits on the new type (non-blocking):
- `OutputMessage` (line ~164) uses `Annotated[..., Field(discriminator='type')]`; `ContentPart` is a bare union, so Pydantic falls back to try-each-member matching. Annotating it with the same discriminator is more consistent and avoids ambiguity if the union is widened.
- `ContentPart` (and `content_to_text`) aren't in `__all__`, yet tests and integrations import them — they're public by use. Worth adding.
- No validator enforces the at-least-one-source invariant: `InputImageContent(type="input_image")` and `InputFileContent(type="input_file")` with all sources `None` construct fine, then fail opaquely at the wire. A small `@model_validator(mode='after')` would turn that into a construction-time error (also closes the empty-block cases flagged above).

src/evaluatorq/common/messages.py
```diff
@@ -12,11 +12,23 @@ def coerce_content_text(content: Any) -> str:
...
+        for part in content:
+            part_type = part.get("type") if isinstance(part, dict) else getattr(part, "type", None)
+            # Both the chat-completions ("text") and Responses ("input_text")
+            # shapes carry their text under a "text" key.
+            if part_type in ("text", "input_text"):
+                texts.append(part.get("text", "") if isinstance(part, dict) else getattr(part, "text", ""))
+            elif part_type in ("image_url", "input_image"):
+                texts.append("[image]")
+            elif part_type in ("file", "input_file"):
+                texts.append("[file]")
```
Baukebrenninkmeijer, line 33: Unknown/future part types fall through all branches and vanish silently — e.g. `audio`/`input_audio`, or `output_text` (a real Responses shape). The docstring promises parts are "visibly accounted for rather than silently dropped", which only holds for the three known categories. Consider an `else` that appends a placeholder and/or logs:
```python
else:
    texts.append(f"[{part_type or 'unknown'}]")
```

Baukebrenninkmeijer (commented):
## Review — multi-modal `Message` content (RES-879)

Nice work. Core path (`content_to_text`, `_render_chat_content`, the OrqResponses + openai-agents list passthrough) is sound, and the integration tests asserting the model-perceived color via self-contained base64 pixels are exactly the right level of verification. The strict (`content_to_text` raises) vs. lenient (`coerce_content_text` placeholders) split is a clean separation.

Left inline comments on the in-diff items. Two more that fall **outside this diff** but are worth a follow-up since `Message.content` now accepts a list:

- **`contracts.py` `to_chat_completion` — tool & assistant-with-tool-calls branches bypass `_render_chat_content`.** Lines ~239 (`'content': self.content or ''`) and ~248 (`'content': self.content`) emit `self.content` raw, unlike the general case which routes through `_render_chat_content`. If either role ever carries a `list[ContentPart]`, raw Pydantic objects reach the OpenAI payload. Low likelihood today (those roles are text), but routing both through `_render_chat_content`/`content_to_text` removes the footgun.
- **Residual raw `m.content or ''` sites that build live chat dicts** (not in this PR's files, flagged for awareness): `simulation/agents/base.py:264,342` and `simulation/runner/simulation.py:529`. A list there would serialize Pydantic reprs into the request. `redteam/ui/dashboard.py:1903,2458` are display-only (repr instead of text) — `coerce_content_text` fits there.

**CI:** `uv run ruff check src` currently fails with 2 errors (I001 in `redteam/backends/orq.py`, E305 in `contracts.py`) — both inline below, both `--fix`-able.

**Test gaps:** `content_to_text` raise is tested for image but not a file part (same branch, but the contract covers all non-text); `_message_to_responses_input_items` multipart-list branch in the openai-agents target is untested (the OrqResponses equivalent is); `coerce_content_text` pydantic-object + `input_text` paths untested; consider `match=` on the `pytest.raises(NotImplementedError)` to lock in the actionable message.

Baukebrenninkmeijer (changes requested):
Requesting changes. Blocker is CI: `uv run ruff check src` fails with 2 errors (I001 in `redteam/backends/orq.py`, E305 in `contracts.py`) — both auto-fixable. Beyond that, the two silent-data-loss cases in `_content_part_to_chat_block` (file_id-only image, file_url/mime_type-only file) should grow a guard before merge. Detailed findings in the prior review: https://github.com/orq-ai/evaluatorq/pull/23#pullrequestreview-4590781381

src/evaluatorq/redteam/backends/orq.py
currentlycodinng: Fixed in a925963 — sorted the deferred import block (ruff I001 clean). The placement after the helper defs is intentional: `E402` is ignored project-wide (pyproject `lint.ignore`), and this block predates this PR, so I sorted in place rather than moving it.

src/evaluatorq/contracts.py
currentlycodinng: Fixed in a925963 — restored the two-blank-line gap before the StrEnum block (E305 clean), and reworded the docstring to "concatenated from its text parts (no separator)" to disambiguate from `coerce_content_text`'s `"\n".join`.

src/evaluatorq/contracts.py
currentlycodinng: Fixed in a925963 — `_content_part_to_chat_block` now raises `NotImplementedError('file_id-backed images are Responses-API only; use OrqResponsesTarget.')` when `image_url is None`, instead of emitting a url-less block. Covered by `test_to_chat_completion_raises_on_file_id_only_image`.

src/evaluatorq/contracts.py
currentlycodinng: Fixed in a925963 — added a guard that raises when both `file_id` and `file_data` are `None` (`file_url`-backed files are Responses-only), with a comment noting `file_url`/`mime_type` are intentionally dropped for the chat shape. Covered by `test_to_chat_completion_raises_on_file_url_only_file`.

src/evaluatorq/contracts.py
currentlycodinng: Addressed two of three in a925963: `ContentPart` is now `Annotated[..., Field(discriminator='type')]` (matching `OutputMessage`), and both `ContentPart` and `content_to_text` are in `__all__`. Left the at-least-one-source `@model_validator` out: it belongs on `InputImageContent`/`InputFileContent` in `openresponses/convert_models.py` (the stacked BOPS-839 base PR), and the empty-block cases it would close are now caught at render time by the explicit raises above. Happy to add it in the base PR as a follow-up.

src/evaluatorq/common/messages.py
currentlycodinng: Fixed in a925963 — added an `else` branch that surfaces unknown/future part types as `[<type>]` (`[unknown]` when no type), so audio/output_text/etc. no longer vanish. Covered by `test_list_surfaces_unknown_part_types_as_placeholders`.

Baukebrenninkmeijer (approved):
Re-reviewed at `a925963`. All blockers and high-severity items addressed — approving.

**Verified fixed:**
- ✅ ruff CI green (`ruff check` passes on `contracts.py` + `orq.py`; I001 import order + E305 blank line both resolved)
- ✅ `_content_part_to_chat_block` now raises `NotImplementedError` for file_id-only images and file_url/file_data-less files — no more silently-invalid blocks. Clear comments on why these are Responses-only.
- ✅ `ContentPart` now a discriminated union (`Field(discriminator='type')`), consistent with `OutputMessage`
- ✅ `ContentPart` + `content_to_text` added to `__all__`
- ✅ `coerce_content_text` surfaces unknown/future types as `[<type>]` placeholders instead of dropping them
- ✅ `content_to_text` docstring corrected ("concatenated, no separator")
- ✅ New tests cover all of the above (file_id/file_url raises, file-part raise, unknown-type placeholders, `match=` on the error messages). 19/19 pass locally.

**Non-blocking follow-ups** (left for a future PR, all out of this diff's scope):
- `to_chat_completion` tool & assistant-with-tool-calls branches still pass `self.content` raw — harmless today since those roles carry text, but worth routing through `_render_chat_content` eventually.
- Residual raw `m.content or ''` in `simulation/agents/base.py`, `simulation/runner/simulation.py`, and the redteam dashboard — only matters once multi-modal reaches those paths.
- `_message_to_responses_input_items` multipart-list branch (openai-agents target) still lacks a unit test.

Nice turnaround.

PR #25, opened 2026-06-26
harmonize target API — remove agent_key, target= single entry (RES-999)

Baukebrenninkmeijer:
## Summary

**Breaking change.** `simulate()` / `generate_and_simulate()` no longer accept `agent_key`. `target=` is the single entry point:

| `target=` value | Resolves to |
|---|---|
| `"agent:<key>"` or bare `"<key>"` | hosted Orq **agent** via the Responses router (`agent/<key>`) |
| `"deployment:<key>"` | legacy Orq **deployment** callback bridge |
| an `AgentTarget` instance | `orq_agent` target path |
| a callable | callback path |

Single commit, cherry-picked clean onto `main`.

## Why (hate-review findings, each validated against code)

- **Backend parity (the important one):** the agent-string path now routes through the same shared `_make_agent_backend(...).create_target()` (OrqResponses, `/v3/router`) that the `eq sim` CLI and the red-team runner already use. The earlier draft used `create_orq_agent_target` → `ORQAgentTarget` (ORQ SDK agents endpoint) — same `agent:<key>` string, *different* infra than the CLI and the docs ("stateless Responses API target"). Now consistent.
- **Empty-string hole:** `target=""` / `"   "` silently built a broken agent target (the deployment path validated, the agent path did not). Now raises `ValueError`.
- **Type guard:** non-callable / non-str / non-`AgentTarget` targets raise a clear `TypeError` instead of deferring to a cryptic call-time failure.

## Docs / examples / scripts

Present **Orq agents as the primary object**, deployments as legacy. Fixed the "AI Studio → Deployments" mislabel → Agents, and the root README snippet that still called the removed `agent_key=` and mislabeled the agent path a "deployment". Examples + live/smoke scripts updated to `target="agent:<key>"` / `target="deployment:<key>"`.

## CLI harmonized too

`eq sim simulate` / `eq sim run` drop the deprecated `--agent-key` flag — the SDK no longer has an `agent_key`, so the CLI shouldn't either. Use `--target deployment:<key>` for the legacy deployment path. `_resolve_target` / `_infer_target_kind`, docstrings, and module examples updated.

## Out of scope (deliberate)

- `wrap_simulation_agent(agent_key=...)` keeps `agent_key` — separate helper.

## Tests

`tests/simulation/test_resolve_target.py` covers agent / bare / deployment / callable / `AgentTarget` branches + empty-string and wrong-type errors. `tests/simulation/test_cli.py` updated for the dropped flag. Full sim suite: **513 passed, 2 skipped**. `basedpyright` clean; CLI ruff-clean.


src/evaluatorq/simulation/api.py
```diff
@@ -620,13 +620,13 @@ async def _simulate_core(
...
-    # AgentTarget instances → "agent:<key>", agent_key deployments → "deployment:<key>",
+    # AgentTarget instances → "agent:<key>", deployment strings → "deployment",
     # plain callables → "callback".
     if target_agent is not None:
         agent_key_attr = getattr(target_agent, 'agent_key', None)
         target_label = f'agent:{agent_key_attr}' if agent_key_attr else 'agent'
-    elif agent_key is not None:
-        target_label = f'deployment:{agent_key}'
+    elif target_kind_hint == 'orq_deployment':
+        target_label = 'deployment'
```
currentlycodinng: Non-blocking: the deployment label drops the key here — it was `deployment:<key>` before, now just `deployment`. The agent branch still carries the key (`agent:<key>`), so the dashboard/run label is asymmetric. The key is available as `value` in `_resolve_target`; if you want parity you could thread it through the kind hint (e.g. `('orq_deployment', value)`).

src/evaluatorq/simulation/api.py
```diff
@@ -736,28 +728,69 @@ async def _simulate_core(
...
+                "Target string is empty; use 'agent:<key>', 'deployment:<key>', "
+                "or a bare '<key>'."
+            )
+        kind, value = _parse_target(resolved)
+        if kind is TargetKind.AGENT:
+            # Same composite backend the CLI / red-team runner use for
+            # ``agent:<key>``: exec via the Responses router (``agent/<key>``),
+            # NOT the ORQ SDK agents endpoint. Keeps SDK and CLI on one path.
+            from evaluatorq.redteam.contracts import LLMConfig, TargetConfig
+            from evaluatorq.redteam.runner import _make_agent_backend
```
currentlycodinng, line 773: Non-blocking: importing the underscore-private `_make_agent_backend` (and `_parse_target` above) from `redteam.runner` into simulation couples to red-team internals — intentional for parity, but it'll break quietly if red-team refactors those. Worth a short comment pinning the dependency, or promoting both to a shared (non-underscore) module both packages import.

src/evaluatorq/simulation/api.py
```diff
@@ -53,9 +53,8 @@
 async def simulate(
```
currentlycodinng, line 53: Non-blocking: `agent_key` is removed outright rather than deprecated, so this is a hard breaking change to `simulate()` / `generate_and_simulate()`. The PR body owns that intentionally — just make sure it lands in the release notes / changelog so downstream callers aren't surprised at runtime.

currentlycodinng (approved):
Verified live: checked out the branch and ran it. `_parse_target` resolution is correct (bare key + `agent:` → AGENT, `deployment:` → DEPLOYMENT, `agent:` raises, whitespace caught by the strip guard before parsing), 85 resolve/CLI tests + 513 simulation tests pass, and basedpyright is clean on the changed files. The headline backend-parity fix checks out — the `agent:` path now goes through `_make_agent_backend(...).create_target()` (Responses router), the same path the CLI and red-team runner use. Nice catch on `target if target is not None else target_callback` quietly fixing the old falsy-target `or` bug too. LGTM. Three minor non-blocking notes inline.

PR #27, opened 2026-06-26
docs(redteam): fix -v → -V vulnerability flag (RES-1005)

Baukebrenninkmeijer:
## What

`docs/custom-evaluators-and-frameworks.md` showed `eq redteam run -t agent:my-agent -v bias_gender` (and `-v my_custom_vuln`). `-v` is `--verbose` (a count); the vulnerability flag is `-V` (`--vulnerability`, verified in `src/evaluatorq/redteam/cli.py:206-207,321-322`). Those commands silently misparse. Fixed `-v` → `-V` in both spots.

## Scope

Just the wrong-flag bug from RES-1005. The other CLI-accuracy items in that ticket (command-prefix consistency, persistence-flag intro, positional-arg naming, `--sim-model` default check, `eq --version` docs) remain open.

Closes part of [RES-1005](https://linear.app/orqai/issue/RES-1005).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28, opened 2026-06-26
feat(cli): eq --version + CLI-reference accuracy (RES-1005)

Baukebrenninkmeijer:
Remaining RES-1005 items (the wrong-flag fix is in #27).

## Changes
- **`eq --version`** — added a top-level Typer version callback to `src/evaluatorq/cli.py`. Before, `eq --version` errored with "No such option". Short `-V` deliberately **not** aliased (it's the `redteam` vulnerability flag). Test in `tests/test_cli_version.py`.
- **cli-reference.md**
  - New "Top-level options" section: `eq --version` + bare `eq` (prints help).
  - redteam-run "Saving results": `--save {none,final,detail}` + `--output-dir` (required for `detail`). Verified against `redteam/cli.py`.
  - New "Recipes" section: CI smoke run, save artifacts, quick sim.
- **dashboard.md** — launch-table commands prefixed with `eq` (`eq redteam ui` / `eq sim ui`) for consistency; `eq` canonical.

## Items intentionally not changed
- **`--sim-model` default `openai/gpt-5.4-mini`** — verified real (`simulation/types.py`, `llm_jury.py`); sim/jury use 5.4-mini, redteam attack/eval use 5-mini. Docs already correct.
- **validate-dataset positional naming** — `redteam` uses `DATASET` (a dataset *source*: path or `hf:`), `sim` uses `PATH` (a JSONL *file*). Names differ because the semantics differ; docs match source. Not unified.

## Verified
- `uv run pytest tests/test_cli_version.py` ✅
- `uv run ruff check src` ✅
- `uv run --group docs mkdocs build --strict` ✅
- `eq --version` → `evaluatorq 1.3.2`

Closes [RES-1005](https://linear.app/orqai/issue/RES-1005).

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -12,6 +12,12 @@ Subcommands are registered at startup. `eq redteam` requires the `redteam` extra
 
 ---
 
+## Top-level options
+
+`eq --version` prints the installed version (e.g. `evaluatorq 1.3.2`) and exits. Running `eq` with no arguments prints help and exits.
+
+---
+
 ## `eq ui`
 
 Launch the FastHTML dashboard showing all red team **and** simulation runs.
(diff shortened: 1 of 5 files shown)


currentlycodinng (approved):
Approve, one note.

Verified the version callback is the standard Typer pattern (`is_eager=True` + `raise typer.Exit`, `Annotated` imported), `-V` deliberately not aliased since it's the redteam vulnerability flag, and the test covers both `--version` and bare-`eq`. The `tests/dashboard/test_fixes.py` change (naming the `dashboard` subcommand) is required and correct — adding `@app.callback()` means `app` is no longer a single-command Typer app, and real users already went through `main()` so there's no behavior regression. Doc additions match the actual CLI.

Note (non-blocking): the diff reformats much of `cli.py` from double to single quotes. Correct per `quote-style = "single"`, but it mixes a formatting cleanup into the feature diff and means the file was previously unformatted. Harmless, just larger than it looks.

Scope check for whoever merges: this PR closes RES-1005 — worth confirming the other ticket items (and #27's flag fix) are all landed or intentionally dropped before it auto-closes.

PR #33, opened 2026-06-29
feat(evaluatorq): no-inference mode — replay pre-recorded responses (RES-993)

currentlycodinng:
## What

Adds an `inference=False` mode to `evaluatorq()` so evaluators run against pre-recorded responses (already in the dataset's `messages` column) without triggering a generation step. Lets you fix the responses and iterate on evaluators alone (the RES-994 alignment workflow).

Closes RES-993.

## How

- `EvaluatorParams.inference: bool = True`; `jobs` is now optional, with a validator that still requires it when `inference=True`.
- When `inference=False`, jobs are replaced by a synthetic replay job that extracts the recorded response from `inputs["messages"]` and feeds it through the existing evaluator pipeline unchanged.
- Dataset sources force `include_messages` on so the column actually loads.
- A provided `jobs` list is ignored in this mode (loguru warning).

## The recorded "response"

Defined as the **last `assistant` message's text content** (multipart content flattened via `coerce_content_text`). If the alignment consumer needs a different shape (full response dict, whole transcript), easy to adjust.

## Fail-loud semantics

A row with no usable recorded response gets a clear error on its `JobResult` and no evaluator scores, rather than aborting the whole run. This matches how job failures already surface and keeps the streaming dataset path intact.

## Tests

`tests/unit/test_no_inference.py` — response extraction (last assistant, multipart flatten, empty / no-assistant / blank-content errors) and the end-to-end path (replays to evaluators, does not run provided jobs, missing-response errors clearly, `inference=True` without jobs raises).

Full unit suite green (2580 passed, 2 skipped); basedpyright 0 errors.

## DoD

- [x] `inference=False` runs evaluators on `messages`-column responses with zero generation calls
- [x] Default behavior (`inference=True`) unchanged
- [x] Missing/invalid response with `inference=False` errors clearly
- [x] Tests cover the no-inference path (incl. missing-response case)
- [ ] PR merged

## Out of scope (follow-ups)

- Loading responses from an orq experiment (RES-1016)
- Auto-detection of a response column

### DIFF
--- a/src/evaluatorq/evaluatorq.py
+++ b/src/evaluatorq/evaluatorq.py
@@ -7,6 +7,9 @@ from datetime import datetime, timezone
 from itertools import starmap
 from typing import Any, cast
 
+from loguru import logger
+
+from .common.messages import coerce_content_text
 from .fetch_data import fetch_dataset_batches, setup_orq_client
 from .processings import process_data_point
 from .progress import Phase, ProgressService, with_progress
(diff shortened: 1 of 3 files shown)


src/evaluatorq/evaluatorq.py
```diff
@@ -51,6 +54,43 @@ def check_pass_failures(results: EvaluatorqResult) -> bool:
...
+        text = coerce_content_text(content)
+        if text.strip():
+            return text
+    raise ValueError(
+        "inference=False requires a recorded response in the 'messages' column, "
+        "but this row has no assistant message with text content."
+    )
+
+
+async def _replay_recorded_response(data_point: DataPoint, _row_index: int) -> dict[str, Any]:
```
Baukebrenninkmeijer: **Critical — CI lint will fail here.** `ruff check src` (the project lint per CLAUDE.md) flags this with `RUF029`: async function with no `await`. But `async` is **required** — `Job` is awaited in `process_job`, so a sync function would break the contract. Fix is a `# noqa: RUF029` with a reason comment, not removing `async`.

src/evaluatorq/evaluatorq.py
```diff
@@ -115,15 +159,24 @@ async def evaluatorq(
...
+            "Either 'params' or both 'data' and 'jobs' keyword arguments are required "
+            "(omit 'jobs' only when inference=False)"
         )
 
     # Extract validated values
     data = validated.data
-    jobs = validated.jobs
+    inference = validated.inference
+    if inference:
+        jobs = validated.jobs or []
```
Baukebrenninkmeijer: Nit (non-blocking): the `or []` is dead — `_require_jobs_when_inferring` already guarantees `jobs` is non-empty whenever `inference=True`, which is the only way to reach this branch. Can simplify to `jobs = validated.jobs`.

Baukebrenninkmeijer (approved):
Re-reviewed at 5d52a1f. Both prior findings resolved:
- RUF029 lint fixed via `# noqa` + explanatory comment (async required by Job protocol).
- Dead `or []` replaced with `cast` + validator-guarantee comment.

Also reviewed the new fail-loud addition (`treat_errors_as_failure=not inference`) — correctly scoped to no-inference mode so a row with no usable recorded response surfaces as a run failure instead of exiting 0. Test covers both branches.

Changed files lint + basedpyright clean, 9/9 tests pass. LGTM 🚀

PR #34, opened 2026-06-29
build(docs): mkdocs hygiene — dedupe flags, link-check, social cards (RES-1003)

currentlycodinng:
Closes RES-1003.

Build-config cleanups and CI hardening for the docs site.

## Changes
- **Dedupe feature flags** — removed the duplicate `content.code.copy` and `toc.follow` entries in `mkdocs.yml`.
- **Broken-link checking** — added `mkdocs-htmlproofer-plugin`, which hard-fails the existing `mkdocs build --strict` step on broken inline links and `#anchor` targets (local + CI). External URLs are skipped to keep CI deterministic. Two autogenerated mkdocstrings re-export anchors (`evaluatorq.EvaluationResultCell`, `evaluatorq.simulation.AgentTarget`) are explicitly excluded with a note — everything else fails the build.
- **Social/SEO cards** — enabled Material's `social` plugin (og:/twitter previews). Gated to `CI=true` so local `mkdocs serve`/`build` keeps working without Cairo; the docs CI job installs the Cairo system libs.

## Verification
- `mkdocs build --strict` passes locally (social off) — exit 0.
- htmlproofer confirmed catching real broken anchors before excludes were added.

## Note for reviewer
Social card *rendering* couldn't be verified on the dev machine (x86_64 Homebrew Cairo vs arm64 Python mismatch); it's wired for and exercised by CI. Worth a glance at the first CI run to confirm cards generate.


Baukebrenninkmeijer (approved):
Validated: dedupe is clean (no features dropped), social cards generate in CI with og/twitter absolute URLs, htmlproofer link-check active and passing under strict build. All CI green.

PR #26, opened 2026-06-26
feat(dashboard): align combined dashboard with v1 editorial design (RES-974)

currentlycodinng:
## What

Transfers the research-team **v1 design** (the "Orq.ai Design System" results-app from the [RES-974](https://linear.app/orqai/issue/RES-974/add-evaluatorq-combined-dashboard) zip) into the existing FastHTML dashboard. This is the alignment step the ticket asks for: align the current dashboard with the v1 design, then iterate toward orq.

It builds on the FastHTML dashboard already on `main` (`feat(dashboard): FastHTML web dashboard...`) — it does not rebuild it.

## Changes

- **Editorial skin** (`theme.py`) — vendored the v1 `:root` tokens verbatim: sand canvas, serif headings, flat surfaces, teal-led sanctioned palette.
- **Sidebar shell** (`shell.py`) — replaced the top-nav chrome with the v1 left sidebar (Dashboard / Red Team / Agent Sim / Settings) + topbar. Inline SVG nav icons; vendored the orq **mark** for the brand lockup (mark + `evaluator`**`q`** with the orange q, exactly as `Shell.jsx`).
- **Combined Dashboard landing** (`metrics.py` + `view.py`) — aggregates across **both** run stores (`.evaluatorq/runs` + `.evaluatorq/sim-runs`): stat band, runs-by-type, attack-resistance donut, findings-by-severity, token usage, recent runs.
- **Per-kind run lists** — Red Team / Agent Sim nav open filtered run lists at `?surface=…`; each run drills into its existing report. Added a Settings stub route.
- **Report view** — v1 back link + topbar Export action, wrapping the existing report body.

## Deliberate scope boundaries

- **Spend panels → native metrics.** Neither schema tracks dollar cost (sim has token counts, red team none), so the design's spend panels render resistance / severity / token-usage instead.
- **Report bodies + their charts unchanged.** They're the shared `common/reports` renderer (also used by the static HTML exports), and that palette already *is* the editorial teal/orange set — re-skinning them would diverge from the exports. A further "iterate toward orq" pass on the report bodies can be a follow-up.
- The resistance headline stat is **attack-weighted** so it agrees with the donut (the per-run mean was inflated by runs with zero evaluated attacks).

## Tests

- New `tests/dashboard/test_landing.py` (metrics aggregation, landing render, run lists, unknown-surface empty state, settings, report header).
- Dashboard suite green (276), full unit suite green (2571), typecheck clean.

## Open for review

[...]

PR #38, opened 2026-06-29
fix(tests): isolate run stores so tests don't pollute the real .evaluatorq

Baukebrenninkmeijer:
## Problem

`red_team()` and the sim runner auto-save every run to `Path.cwd()/.evaluatorq/{runs,sim-runs}`. Running the test suite therefore wrote throwaway runs (`description: "Test report"`, `tested_agents: ["agent:test"]`, 0 attacks) into the developer's **real** store — which then showed up as junk entries in the combined dashboard.

## Fix

- `get_runs_dir()` / `get_sim_runs_dir()` now honor an `EVALUATORQ_DIR` env override (fall back to cwd as before).
- Autouse fixture `_isolate_run_stores` in `tests/conftest.py` points `EVALUATORQ_DIR` at a per-test `tmp_path`, so **no test can write into the repo store**.

No dashboard behavior change — empty runs still display. This only stops tests leaking into the store.

## Verification

- Ran full `tests/redteam` suite: repo store count unchanged (0 leak; previously it grew each run).
- `uv run pytest -m 'not integration'` → all pass.

Stacked on `aminaakhmedova/res-974-add-evaluatorq-combined-dashboard`.

### DIFF
--- a/src/evaluatorq/common/run_store_dir.py
+++ b/src/evaluatorq/common/run_store_dir.py
@@ -1,0 +1,21 @@
+"""Resolve the on-disk store directory for persisted runs.
+
+Both red teaming (``runs/``) and agent simulation (``sim-runs/``) persist
+their reports under ``.evaluatorq/``. The base resolves to ``$EVALUATORQ_DIR``
+when set (tests point this at a tmp dir so runs never leak into the repo
+store), otherwise ``.evaluatorq`` relative to the current working directory.
+"""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+STORE_DIR_NAME = ".evaluatorq"
+
+
+def get_store_dir(subdir: str) -> Path:
+    """Return ``<base>/<subdir>`` where base honors ``EVALUATORQ_DIR``."""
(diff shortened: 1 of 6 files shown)


currentlycodinng (approved):
Approve. Clean fix for a real problem — the suite was writing throwaway runs into the dev's real `.evaluatorq/` store, surfacing as junk in the dashboard.

Verified: `get_store_dir()` honors `EVALUATORQ_DIR` with empty-string-as-unset handled and the cwd fallback preserved; `evaluatorq.common` is a real package so the new module imports fine; the autouse fixture covers the whole suite and the two cwd-fallback tests correctly `delenv` first (fixture ordering is right). New module has unit tests for all three branches. CI green 3.10–3.13.

Two coordination notes, non-blocking:
- `EVALUATORQ_DIR` points at the store dir itself, not its parent (`.../x/.evaluatorq`, not `.../x`) — documented in the module docstring, but slightly counterintuitive, and it's a new semi-public env var not listed in CLAUDE.md's env-var section or the docs.
- Stacked on #26 (combined-dashboard) — should land after that, not directly to main.

PR #40, opened 2026-06-29
feat(evaluatorq): no-inference mode loads responses from an orq experiment (RES-1016)

currentlycodinng:
## What

Extends `inference=False` (RES-993) so pre-recorded responses can come from an existing **orq experiment**, not just a dataset's `messages` column. Re-run evaluators against a prior experiment's outputs with **zero generation calls**:

```python
from evaluatorq import evaluatorq, ExperimentInput

await evaluatorq(
    "rescore",
    data=ExperimentInput(experiment_id="01KW9AMT..."),  # run_id optional → latest run
    evaluators=[...],
    inference=False,
)
```

## How

- `ExperimentInput(experiment_id, run_id=None)` — a second response source for the no-inference path. Valid only with `inference=False` (enforced by the `EvaluatorParams` validator).
- `fetch_experiment_datapoints` (raw httpx, since the SDK exposes no experiments resource):
  - `GET /v2/spreadsheets/{id}/manifests` → pick the run (latest by `created` when `run_id` omitted).
  - `POST /v2/spreadsheets/{id}/manifests/{run_id}/export {format: jsonl}` → follow the signed-URL redirect and download the JSONL.
  - Map each row to a `DataPoint`, surfacing `task_output` as the final assistant message so the existing replay path (`_replay_recorded_response` / `extract_recorded_response`) consumes it unchanged. Existing conversation in the row's inputs is preserved.
- Clear `ValueError` when the experiment, run, or rows can't be loaded (DoD: fail loud).

## Testing

- Unit tests (`tests/unit/test_no_inference_experiment.py`, mocked `httpx` transport): row mapping (JSON-string inputs, dict inputs, missing/blank response), latest-run selection, explicit `run_id`, permission/empty-runs/empty-rows errors, validation, and the end-to-end `evaluatorq()` flow.
- **Live-tested** against the research workspace experiment `01KW9AMT6W8H3K05CVYE7RR2KZ`: loaded 2 recorded responses, replayed and scored them via `evaluatorq(inference=False)` with no generation calls; bad-experiment-id returns a clear error.
- Full non-integration suite: 2614 passed.

## Notes

Builds on RES-993 (#33, merged). Related: RES-994.

Closes RES-1016.


Baukebrenninkmeijer (changes requested):
## Review — no-inference experiment replay (RES-1016)

Solid, well-tested implementation. `ruff`, `basedpyright`, and the new test file all pass, and the `ExperimentInput` → row → `extract_recorded_response` contract is correct. Requesting changes for one real bug plus documentation/discoverability gaps that would leave the feature unusable-in-practice for an engineer who doesn't already know it exists.

### Blocking

**1. Wrong-run selection when manifests lack `created`** — `src/evaluatorq/fetch_data.py:256`
```python
latest = max(manifests, key=lambda m: m.get("created") or "")
```
If `created` is missing/None on every manifest, all sort keys collapse to `""` and `max` returns the *first* manifest by iteration order — not the newest. There's no log and `resolved_run_id` stays truthy, so the `if not resolved_run_id` guard doesn't trip. The user asks for "latest" and silently scores an arbitrary run. Fail loudly (or at least warn) when no manifest carries a timestamp:
```python
if all(not m.get("created") for m in manifests):
    raise ValueError(
        f"Experiment '{experiment_id}' manifests have no 'created' timestamps; "
        "cannot determine the latest run. Pass run_id explicitly."
    )
latest = max(manifests, key=lambda m: m["created"])
```

**2. No documentation anywhere** — `docs/`
No mention of no-inference mode *or* `ExperimentInput` (the underlying RES-993 mode is also undocumented). As written, an engineer won't discover the feature exists. Needs a section in `docs/orq-deployment.md` (or a short guide) covering what no-inference mode is, the `ExperimentInput` source, and how it relates to `DatasetIdInput`.

**3. Main `evaluatorq()` docstring is stale and now contradicts itself** — `src/evaluatorq/evaluatorq.py:148-149`
The `data:` Args entry still reads *"Either a DatasetIdInput to fetch from Orq platform, or a list of DataPoint instances/awaitables."* `ExperimentInput` was added to `EvaluatorParams.data`'s docstring but not to the function most users call, so the two docstrings disagree. Add the `ExperimentInput` clause here too.

**4. `experiment_id` / `run_id` aren't discoverable** — `src/evaluatorq/types.py:174-178`, `src/evaluatorq/fetch_data.py`
Docstrings describe these as "experiment (sheet) ID" and "run (manifest) ID". Between experiment/run, sheet/manifest, and the `/v2/spreadsheets/.../manifests/...` URL, three vocabularies describe two IDs, and none of them tells me *where in the Orq UI/URL* to read the values off — which is the first question after running an experiment. State the UI/URL location explicitly in the docstrings and the docs section.

### Non-blocking suggestions

- **Public surface is the right shape.** `ExperimentInput` as a peer in the `data` union (rather than overloading `DatasetIdInput`) keeps each source's fields non-optional and self-describing. No change wanted.
- **Add an `examples/` script** (e.g. `05_replay_experiment.py`) — doubles as a smoke test via the existing `smoke_examples.py` harness.
- **No CLI flag needed.** `cli.py` exposes no eval-run command at all, so the absence of a `--from-experiment` flag is consistent, not a regression. Follow-up ticket if CLI-driven eval runs ever land.
- **Contextualize the final JSONL parse** — `fetch_data.py:301`. A malformed line raises a bare `json.JSONDecodeError` with no experiment/run/line context; wrapping it to name the run would match the error quality of the rest of the function.

### Confirmed safe (investigated, not issues)
- `contextlib.suppress(JSONDecodeError)` on the export body → falls through to a clear `ValueError` at the next guard.
- `{"input": raw}` fallback for unparseable `inputs` → lossless, doesn't touch the replay path.
- Blank `task_output` → empty messages → per-row `ValueError`, counted as a failure via `treat_errors_as_failure=not inference`, exits loudly.
- Signed-URL download dropping the bearer token → correct and intentional; non-success raises.

### Strengths
- 234-line test file covers row parsing, latest-run resolution, explicit `run_id`, every error path, and end-to-end through `evaluatorq()`.
- Thorough, targeted `ValueError`s with the experiment/run id interpolated.

ChielToto:
## Review — RES-1016 no-inference mode loads responses from an orq experiment

**What it does:** Extends `inference=False` so pre-recorded responses can come from a prior experiment (not just a dataset). Adds `ExperimentInput(experiment_id, run_id=None)`, a raw-httpx `fetch_experiment_datapoints` (list manifests → export run as JSONL → map each row to a `DataPoint`, surfacing `task_output` as the final assistant message for the existing replay path), and a validator forbidding `ExperimentInput` with `inference=True`. Well-structured, fails loud on load errors, and reasonably tested — but the row-mapping and run-selection only cover the string/happy path.

### Findings (most severe first)

**1. `fetch_data.py` — "latest run" sorts on the wrong field name, silently selecting an arbitrary run.**
`latest = max(manifests, key=lambda m: m.get("created") or "")`. The manifest timestamp is serialized as `_createdAt` (see `send_results.py:21`, alongside `_id`/`_name`/`_endedAt`), not `created`. So `m.get("created")` is `None` for every manifest → all keys tie at `""` → `max()` returns the **first** element in API order, not the newest. A user re-scoring "the latest run" (the documented default when `run_id` is omitted) can get a stale run with no error. Note the inconsistency: `_id` gets a defensive `latest.get("_id") or latest.get("id")` fallback two lines down, but the timestamp gets none. The unit test only asserts against a self-authored mock that uses `created`.

**2. `fetch_data.py` — structured (dict) `task_output` is replayed as a corrupted Python-repr string.**
`messages.append({"role": "assistant", "content": response})` stores `task_output` raw; the replay path then runs it through `coerce_content_text`, whose fallback is `str(content or "")`. A JSON-mode row with `task_output = {"answer": "yes"}` becomes the string `"{'answer': 'yes'}"` (single quotes, invalid JSON). Any evaluator/judge that parses the response as JSON sees a mangled string, scoring structured-output experiments incorrectly. Only string outputs are tested.

**3. `fetch_data.py` — a list-shaped `task_output` collapses to `""` and raises a misleading "no recorded response" error.**
Same append line: a `task_output` like `["a", "b"]` is passed to `coerce_content_text`, whose list branch keeps only `{"type":"text"}` dict parts → returns `""`. `extract_recorded_response` then finds no assistant text and raises "row has no assistant message with text content" — failing the run loudly even though the row *did* contain a recorded response.

**4. `fetch_data.py` — empty `task_output` with a pre-existing assistant turn replays the stale turn instead of failing.**
The guard `if response is not None and str(response).strip()` skips the append when `task_output` is blank, but the code explicitly preserves any `messages` already in `inputs`. If that existing conversation ends in an assistant message, `extract_recorded_response` returns it. The inline comment ("leaves no assistant message to replay, so the replay path raises a clear per-row error and the run fails loudly") is then false, and the `treat_errors_as_failure` safety net is bypassed because the job doesn't error — a stale response is silently scored.

**5. `fetch_data.py` — export handling has no path for an inline-200 JSONL body.**
The export response is consumed only as a 302 redirect (`location`) or a 200 JSON envelope (`url`/`redirectUrl`/`signedUrl`). If `/export` returns 200 with the JSONL payload directly in the body, `export_resp.json()` raises `JSONDecodeError` (suppressed) → `export_url` stays `None` → "did not return a download URL", failing a successful export. Only the 302 path is tested; worth confirming the export contract can't return inline content.

**6. `fetch_data.py` — `_resolve_orq_base_url` is a third verbatim copy of the base-URL precedence logic (reuse).**
`(base_url or os.getenv("ORQ_BASE_URL", ORQ_DEFAULT_HOST)).rstrip("/")` is character-identical to `send_results.py:96`, and a third variant lives in `common/llm_client.py`. The orq-host precedence now has to be changed in three places to stay consistent; a shared helper in `common/llm_client.py` (where `ORQ_DEFAULT_HOST` already lives) would consolidate this and the duplicated bearer-header construction.

---

Findings 1–3 are the ones to act on before merge (#1 is a silent wrong-results bug; #2/#3 break non-string outputs, which evaluatorq treats as first-class). #4–#5 are real edge gaps; #6 is cleanup. The `inference=False` validator gating and the `ExperimentInput`/`DatasetIdInput` disjointness are sound — no input-routing bug there.

**Rating: 3.5/5** — clean structure, good error messages and test scaffolding, but the row-mapping assumes string outputs and the latest-run selection keys on an unverified field name, both of which can silently mis-score.

currentlycodinng:
Thanks — all four blocking items addressed in f64071f:

1. **Wrong-run selection** — `fetch_experiment_datapoints` now raises when no manifest carries a `created` timestamp, instead of letting `max` return an arbitrary run silently. Pass `run_id` explicitly in that case.
2. **Documentation** — added a no-inference / `ExperimentInput` replay section to `docs/orq-deployment.md`, covering what the mode is, how it relates to `DatasetIdInput`, and where to read `experiment_id` / `run_id` off the Orq UI.
3. **Stale `evaluatorq()` docstring** — the `data:` entry now mentions `ExperimentInput` (requires `inference=False`), matching `EvaluatorParams`.
4. **ID discoverability** — the `ExperimentInput` and `fetch_experiment_datapoints` docstrings now spell out the experiment/sheet + run/manifest vocabulary and the UI/URL location for each ID.

Also picked up the non-blocking suggestion to wrap malformed JSONL export lines with experiment/run/line context.

On the `examples/05_replay_experiment.py` suggestion: skipped for now — there's no `smoke_examples.py` harness in this repo (examples are grouped under `examples/agent_simulation` and `examples/redteam`), so it wouldn't double as a smoke test as described. Happy to add a standalone example in a follow-up if you'd like.

currentlycodinng:
@Baukebrenninkmeijer re-requesting review — all 4 blocking items addressed (see comment above). Ready for another look when you have a moment.

Baukebrenninkmeijer (approved):
## Re-review — all blockers addressed ✅

Verified `f64071f` against the previous review. `ruff`, `basedpyright`, and the 13-case test file all still pass.

- **Undated-run bug (was blocking #1)** — fixed. `fetch_data.py` now raises when no manifest carries a `created` timestamp instead of silently scoring an arbitrary run. The `max(..., key=... or "")` that follows is safe now that at least one timestamp is guaranteed.
- **No docs (#2)** — fixed. `docs/orq-deployment.md` gains a "Replaying an experiment's responses" section with a source-comparison table, ID-location guidance, and usage examples. Clear and accurate.
- **Stale `evaluatorq()` docstring (#3)** — fixed; the `data:` entry now lists `ExperimentInput`.
- **ID discoverability (#4)** — fixed. Both the docstrings and the docs now state exactly where `experiment_id`/`run_id` surface in the Orq UI/URL and reconcile the experiment/spreadsheet and run/manifest vocabulary.
- **Bonus (suggestion)** — malformed JSONL export lines are now wrapped with line number + run/experiment context.

### Non-blocking nit
The new "manifests carry no `created` timestamps" guard (`fetch_data.py`) has no dedicated test — the existing suite doesn't exercise that branch. A one-line case (undated manifests → `ValueError`) would lock it in, but not worth holding the PR.

Approving.

PR #41, opened 2026-06-29
fix(contracts): re-land multi-modal Message content to main (RES-879)

currentlycodinng:
## Why

RES-879 (multi-modal `Message.content`) was approved and merged as **#23**, but it merged into the intermediate `bops-839` branch *after* that branch had already merged to `main` (#17, June 26). #23 merged into `bops-839` on June 29 — so its `contracts.py` changes never reached `main`. On current `main`, `Message.content` is still `str | None`, the `developer` role is missing, and the `_render_chat_content` / `content_to_text` helpers don't exist.

This PR re-lands the exact approved RES-879 work onto `main` by cherry-picking its three commits.

## What

Cherry-pick of `fb54c85`, `22b4b67`, `32aaab0` (the original RES-879 branch commits):

- `Message.content`: `str | list[ContentPart] | None`; adds `developer` role; string shorthand still accepted.
- Helpers `_render_chat_content` (chat-completions blocks) and `content_to_text` (flatten to text, raise on non-text parts for text-only targets).
- `AgentTarget.respond` implementations updated to pass multi-part content through or raise a clear `NotImplementedError`.
- Tests: `tests/contracts/test_message_multimodal.py`, `tests/common/test_messages.py`, plus integration coverage.

One conflict resolved by hand: `redteam/backends/orq.py` import block (main added a tracing import; kept it alongside the new `content_to_text` import).

## Testing

- `tests/contracts/test_message_multimodal.py` + `tests/common/test_messages.py`: 19 passed.
- Full non-integration suite: 2614 passed.
- `basedpyright` clean on the touched modules. No new `ruff` errors vs the `main` baseline.

## Notes

Unblocks **RES-1018** (handle multi-modal `Message.content` at the remaining content sinks), which depends on these helpers being on `main`.

Closes RES-879 (re-land).

### DIFF
--- a/src/evaluatorq/common/messages.py
+++ b/src/evaluatorq/common/messages.py
@@ -12,17 +12,23 @@ def coerce_content_text(content: Any) -> str:
     ``[{"type": "text", "text": "..."}]``) surfaces the joined text rather than a
     Python ``repr`` of the list. ``None`` becomes ``""``; plain strings (and anything
     else) pass through ``str``.
+
+    Unlike :func:`evaluatorq.contracts.content_to_text`, this best-effort helper
+    does not raise on non-text parts (it is used in report/transcript rendering).
+    Image and file parts are surfaced as a ``[image]`` / ``[file]`` placeholder so
+    they are visibly accounted for rather than silently dropped.
     """
     if isinstance(content, list):
         texts: list[str] = []
         for part in content:
-            if isinstance(part, dict):
-                # Both the chat-completions ("text") and Responses ("input_text")
-                # shapes carry their text under a "text" key.
-                if part.get("type") in ("text", "input_text"):
-                    texts.append(part.get("text", ""))
-            elif getattr(part, "type", None) == "input_text":
-                # An InputTextContent pydantic content part.
-                texts.append(getattr(part, "text", ""))
+            part_type = part.get("type") if isinstance(part, dict) else getattr(part, "type", None)
+            # Both the chat-completions ("text") and Responses ("input_text")
+            # shapes carry their text under a "text" key.
+            if part_type in ("text", "input_text"):
+                texts.append(part.get("text", "") if isinstance(part, dict) else getattr(part, "text", ""))
+            elif part_type in ("image_url", "input_image"):
+                texts.append("[image]")
+            elif part_type in ("file", "input_file"):
+                texts.append("[file]")
         return "\n".join(texts)
     return str(content or "")
--- a/src/evaluatorq/common/messages.py
+++ b/src/evaluatorq/common/messages.py
@@ -14,9 +14,15 @@ def coerce_content_text(content: Any) -> str:
     else) pass through ``str``.
     """
     if isinstance(content, list):
-        return "\n".join(
-            part.get("text", "")
-            for part in content
-            if isinstance(part, dict) and part.get("type") == "text"
-        )
+        texts: list[str] = []
+        for part in content:
+            if isinstance(part, dict):
+                # Both the chat-completions ("text") and Responses ("input_text")
+                # shapes carry their text under a "text" key.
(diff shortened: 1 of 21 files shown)


Baukebrenninkmeijer (changes requested):
Reviewed for integration depth, scoping vs RES-879, and modality coverage. Looks good overall — clean discriminated-union seam, broad target wiring (incl. crewai + pydantic_ai beyond the ticket), tests green. Two things before merge:

### 1. ORQAgentTarget: confirm flatten-vs-passthrough is intentional

RES-879 scope lists `ORQAgentTarget` under "pass through", but the implementation flattens via `content_to_text` (raises `NotImplementedError` on image/file parts):

```python
# redteam/backends/orq.py:154
prompt = content_to_text(messages[-1].content)
```

This is defensible — the target sends a string `prompt` to a stateful agent endpoint, not Responses content-parts, so true pass-through would need a different invoke shape. Acceptance criteria don't require it. But it's a stated-scope item being descoped. Please confirm this is intentional and add a one-line note in the PR/code so it's not read as an oversight later.

### 2. Three `contracts.Message` consumers still render a list as raw repr

These were missed by the `content_to_text`/`coerce_content_text` sweep — a multi-part list reaches them and renders as a Python `repr`:

- `integrations/vercel_ai_sdk_integration/target.py:211,213` — the `tool` branch uses `m.content or ''` while the rest of the same function uses `content_to_text`. Inconsistent.
- `integrations/openai_agents_integration/target.py:326` — the `assistant`+`tool_calls` branch passes raw `m.content`; if it's a list you get pydantic objects in the dict. The list-passthrough at :342 only covers the no-tool-calls case.
- `redteam/ui/dashboard.py:1903` — `msg.content or ""` into `st.code()`; a multimodal redteam transcript would show `[InputTextContent(...)]`.

If these are the "remaining content sinks" tracked under RES-1018, that's fine — just confirm here so they're not silently dropped. Otherwise they're one-line swaps to `content_to_text` / `coerce_content_text`.

Everything else (developer role, ContentPart union, OrqResponses + openai_agents passthrough, report rendering) looks correct.

currentlycodinng:
Thanks for the careful review. Both points addressed:

**1. ORQAgentTarget flatten — confirmed intentional.** The agents endpoint takes a plain string `prompt`, not Responses content-parts, so RES-879's "pass through" doesn't apply to this target: a stateful text endpoint can't forward image/file parts anyway. Added a code comment at the flatten site (46e5eab) so it reads as a deliberate descoping rather than a miss.

**2. The three list-repr consumers are RES-1018 scope** (#42, "remaining content sinks"):
- `redteam/ui/dashboard.py:1903` — already swapped to `coerce_content_text` in #42.
- `integrations/vercel_ai_sdk_integration/target.py` (tool branch) and `integrations/openai_agents_integration/target.py` (assistant+tool_calls + function_call_output) — these were the genuine gaps you spotted; the RES-1018 sweep had missed them. Now fixed in #42 (ba26778) with direct unit coverage for all four cases. So nothing is silently dropped.

currentlycodinng:
@Baukebrenninkmeijer re-requesting review — both points addressed (see comment above). Ready when you are.

Baukebrenninkmeijer (approved):
Re-reviewed at 46e5eab. Both points resolved:

1. **ORQAgentTarget** — clear code comment added (`orq.py:154`) documenting the intentional text-flatten and why "pass through" doesn't apply to the string-prompt agents endpoint. Good.

2. **Remaining content sinks** — deferred to RES-1018 (#42), which is stacked on this branch. Merge order #41 → #42 covers all three. Legitimate scoping; keeps this PR a focused re-land.

Contract union, developer role, OrqResponses + openai_agents passthrough, and report rendering all check out. Multimodal + messages tests green (19 passed). LGTM.

PR #42, opened 2026-06-29
fix: handle multi-modal Message.content at remaining content sinks (RES-1018)

currentlycodinng:
## What

Follow-up to RES-879. `Message.content` is now `str | list[ContentPart] | None`, but a few sinks still read it as a plain string — a list would serialize raw Pydantic reprs into a request or display. These are latent today (nothing produces list content for these paths yet), so this closes them out before multi-modal flows spread beyond the user→Responses path.

## Sinks fixed

- **`contracts.py` `to_chat_completion`** — route the `tool` branch through `content_to_text` (text-only result, fails loud on image/file) and the `assistant`-with-`tool_calls` branch through `_render_chat_content`, while **preserving `None`** content on pure tool-call turns (OpenAI accepts null). The general case already routed through `_render_chat_content` (RES-879).
- **`simulation/agents/base.py`** (chat-completions + Responses request dicts) — `content_to_text` (fails loud on the not-yet-supported image/file case, which is correct for a real request).
- **`simulation/runner/simulation.py`** (span recording) — `coerce_content_text`, the lenient flatten, so observability never crashes a run on unexpected content.
- **`redteam/ui/dashboard.py`** (two display paths) — `coerce_content_text`.

## Test gap closed

- `_message_to_responses_input_items` multipart-list branch in the openai-agents target (the `OrqResponsesTarget._messages_to_input` equivalent was already covered).
- Multipart `tool` and `assistant`-with-`tool_calls` branches of `to_chat_completion`.

## Note

The openai-agents target's `_message_to_responses_input_items` already passes multi-part content through correctly (RES-879); the ticket scoped it to a test only, which is what's added here. Its `tool`/assistant-text branches still use `m.content or ''` (a list there is an edge the ticket didn't include) — left as-is to keep the change within the ticket's stated sinks.

## Testing

- New unit tests pass; full non-integration suite: 2618 passed.
- `ruff`/`basedpyright` clean on changed modules.

## ⚠️ Base branch

This PR is **stacked on #41** (`aminaakhmedova/res-879-reland-multimodal-message`), which re-lands RES-879's contracts changes to `main` — they never actually reached `main` (see #41). The helpers this PR routes through (`_render_chat_content`, `content_to_text`) only exist once #41 lands. **Merge #41 first**, then retarget this PR to `main`.

Closes RES-1018.

### DIFF
--- a/src/evaluatorq/contracts.py
+++ b/src/evaluatorq/contracts.py
@@ -252,16 +252,18 @@ class Message(BaseModel):
             param: dict[str, Any] = {
                 'role': 'tool',
                 'tool_call_id': self.tool_call_id or '',
-                'content': self.content or '',
+                'content': content_to_text(self.content),
             }
             if self.name:
                 param['name'] = self.name
             return param
         if self.role == 'assistant' and self.tool_calls:
-            # content may be None on a pure tool-call turn; OpenAI accepts null.
+            # content may be None on a pure tool-call turn; OpenAI accepts null, so
+            # preserve None and only render string/list content (multi-part lists would
+            # otherwise leak raw Pydantic objects into the payload).
             return {
                 'role': 'assistant',
-                'content': self.content,
+                'content': _render_chat_content(self.content) if self.content is not None else None,
                 'tool_calls': [
                     {
                         'id': tc.id,
--- a/src/evaluatorq/integrations/openai_agents_integration/target.py
+++ b/src/evaluatorq/integrations/openai_agents_integration/target.py
@@ -9,6 +9,7 @@ from typing import Any
 from agents import Agent, Runner
 from loguru import logger
 
+from evaluatorq.common.messages import coerce_content_text
 from evaluatorq.contracts import (
     AgentContext,
     AgentResponse,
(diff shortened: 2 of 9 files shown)


Baukebrenninkmeijer:
## Completeness check — 3 parallel content sinks possibly missed

This PR flattens multi-modal `Message.content` at the chat-completions / dashboard / simulation sinks. The same `tool` / assistant-tool_calls content is still raw-flattened in two integration targets — same bug class the PR docstrings warn about (multi-part `list[ContentPart]` leaking raw Pydantic objects into the payload). These files aren't in the diff, so flagging here.

**Question for each: in scope for RES-1018, or deliberately deferred?**

1. **`src/evaluatorq/integrations/vercel_ai_sdk_integration/target.py:211,213`** — tool-result uses `m.content or ''`, while the *same file* (lines 218, 233) already uses `content_to_text(m.content)`. A tool message with multi-part content would leak raw objects here. Should these use `content_to_text` too?

2. **`src/evaluatorq/integrations/openai_agents_integration/target.py:322`** — `function_call_output` output uses `m.content or ''`. Multi-part tool content leaks raw. (File doesn't import `content_to_text` yet.) Flatten here?

3. **`src/evaluatorq/integrations/openai_agents_integration/target.py:326`** — assistant-with-tool_calls passes `'content': m.content` raw; a `list[ContentPart]` is truthy and never dumped, unlike the list-dump branch at line 341. Dump to `[p.model_dump(mode='json') for p in m.content]`?

Severity: these only break on the multi-modal path (tool/assistant-tool_calls content is usually a plain string today), so not blocking — but the PR's stated scope is "remaining content sinks," and the `contracts.py` tool-role change shows tool content is meant to be flattened.

**Verified safe (not sinks):** `redteam/backends/openai.py:120`, `judge.py`, `jobs.py`, `orchestrator.py` (OpenAI SDK response `.content`, always str); `openresponses/target.py:195` (`or ""` only reached for non-list; list handled one branch up).

If in scope, each is a ~1-line fix plus a tool-role multipart test. Happy to push.

currentlycodinng:
@Baukebrenninkmeijer added the two integration-target fixes (vercel + openai_agents) that surfaced in the #41 review, with unit coverage. Ready for review.

Baukebrenninkmeijer (approved):
LGTM. All 3 flagged content sinks fixed with unit coverage; CI green. Minor non-blocking nit: openai_agents function_call_output uses coerce_content_text while contracts/vercel tool paths use content_to_text — align later if desired.

PR #52, opened 2026-07-01
feat(dashboard): Agent Sim + Red Team overviews, real cost, ⌘K search, settings (RES-1021/1022/1038)

currentlycodinng:
## What

Builds out all three dashboard follow-ups from the RES-974 v1-design diff, on one branch since they share the same files. Bundled for combined review per the alignment plan — we sort out any adaptations here.

### RES-1022 — Agent Sim
- Rich overview page (`?surface=sim`): 4 KPI cards + item-level "Recent simulations" table (Scenario · Persona · Model · Turns · Outcome · Score).
- Outcomes donut (achieved / not achieved / errors) on the report Overview tab.

### RES-1021 — Red Team
- Rich overview page (`?surface=redteam`): 4 KPI cards (Attacks run · Break rate · Critical findings · Avg robustness) + item-level "Recent attacks" table (Target · Attack · Model · Severity · Status · Time).
- Resize-on-tab-show fix for wide Vega charts (they rendered tiny inside hidden CSS tabs).

### RES-1038 — Shell
- **⌘K global search** — topbar box + `/search` HTMX fragment; Cmd/Ctrl+K focuses, Escape clears.
- **Landing spend model** — Total-spend stat + "Spend by job type" panel.
- **Settings** — read-only runtime-config view (run stores, default model, API-key presence).

## Two things that turned out differently than the tickets assumed

1. **Cost already exists.** The tickets (and PR #26) assumed no dollar cost is tracked and proposed building a cost model. Not needed — both surfaces already record `cost_usd` upstream (red team `summary.token_usage_total`, sim per-result `token_usage`). So the cost cards and landing spend show **real dollars**, and RES-1038's "build a cost model vs keep native metrics" decision collapses to "wire up the existing field." The sim card upgraded from the token stand-in to real `Avg cost/sim`.

2. **Settings doesn't map to a local viewer.** The v1 design's Settings is workspace management (editable name, API keys, delete project) — that belongs to the ORQ platform UI, not a local read-only run viewer. I built an honest read-only config page reflecting what the dashboard actually reads, and noted the gap in-page. Flagging for your call: leave as config view, or drop Settings from this dashboard entirely.

## Still open (not in this PR)
- Explorer filterable sample table + per-row export (RES-1021 parity) — overlaps the existing filter sidebar + conversation viewer; wanted your read before duplicating.
- Trend pills on KPI cards — genuinely need historical/time-series data that isn't captured yet.

## Tests
- New `test_sim_overview.py`, `test_redteam_overview.py`; search + settings-config cases in `test_landing.py`.
- Full non-integration suite green (2635 passed), basedpyright clean, real-data smoke test across landing/overviews/search/settings.

@bauke-brenninkmeijer — the two "turned out differently" points are the ones worth your eyes.


src/evaluatorq/dashboard/metrics.py
```diff
@@ -218,3 +240,235 @@ def _results(data: dict[str, object]) -> list[dict[str, object]]:
...
+                ),
+            ))
+
+    dated.sort(key=lambda t: t[0], reverse=True)
+    return SimOverview(
+        simulations_run=total,
+        goal_completion=(achieved / total) if total else None,
+        avg_turns=(turns_total / total) if total else None,
+        avg_tokens=(tokens_total / total) if total else None,
+        avg_cost=(cost_total / total) if (total and has_cost) else None,
```
KarinaKKarinaK: Deep-review finding, non-blocking, worth confirming the intent. `avg_cost` divides `cost_total` by `total` (all sims), but `cost_total` only accumulates results that carry `cost_usd` (the `if isinstance(usage, dict) and 'cost_usd' in usage` guard above), and `has_cost` flips True as soon as **one** result has cost. So with partial cost data across the store (e.g. older runs recorded before `cost_usd` existed, plus newer runs that have it) the "Avg cost/sim" KPI (a) understates the real per-costed-sim cost by spreading it over the cost-less sims too, and (b) suppresses the tokens/sim fallback for the whole store the moment a single sim has cost. `avg_tokens` using the same `/total` denominator is fine (a missing token count really is 0), but a missing `cost_usd` is "unknown", not "$0". If cost is genuinely all-or-nothing per store this never bites; if it can be mixed, consider dividing by the count of costed sims, or only taking the cost path when every result carries `cost_usd`.

KarinaKKarinaK (commented):
Deep review of the three dashboard follow-ups. Read the new logic end-to-end (metrics, view, app routes, dashboard.js, report_tabs) and ran the suite locally.

What I verified:
- **Security.** The Cmd-K `/search` fragment never reflects the user query back into the HTML (empty -> '', no-hits -> a static string, hits -> every field `esc()`'d), so no reflected XSS even though the route returns `NotStr`. Settings shows API keys as `set`/`not set`, never the value, and only surfaces local paths/default model.
- **Cost / KPI math.** break_rate, avg_robustness (attack-weighted), goal_completion, the attack-weighted landing resistance, and the outcomes donut all read correctly, and the donut vs the table share the same error-first bucketing so they can't disagree. Dict parsing is defensive throughout.
- **JS.** The Cmd-K handler and the tab-show Vega resize are guarded (offsetParent check, try/catch, requestAnimationFrame), no resize-loop risk.
- **Tests.** 295 dashboard tests pass locally.

One non-blocking finding inline on the `avg_cost` denominator. The two 'turned out differently' points (cost already exists; settings scope) are product calls for @Baukebrenninkmeijer, not code issues.

Clean, well-tested PR.

KarinaKKarinaK (approved):
Approving. Deep-reviewed the logic (per my earlier comment): search is XSS-safe, settings never leaks key values, the cost/KPI math and donut bucketing are correct, the JS is guarded, and 295 dashboard tests pass locally. The one avg_cost-denominator note is non-blocking — fine to address in a follow-up or confirm it's the intended behavior. Nice, well-tested PR.

src/evaluatorq/dashboard/metrics.py
Baukebrenninkmeijer: Good catch — it can be mixed (older sim runs predate `cost_usd`), so this did bite. Fixed in 211b3e1: `avg_cost` now divides `cost_total` by the count of sims that actually carry a `cost_usd`, and returns `None` only when *no* sim has cost. A missing cost is treated as "unknown", not $0, so cost-less sims no longer dilute the mean. Added a mixed-store regression test (`test_avg_cost_ignores_costless_sims`) asserting one costed sim at $0.006 reads $0.006, not $0.003.

PR #54, opened 2026-07-03
style(evaluatorq): lint-clean the tree + make CI ruff gate blocking (RES-967)

currentlycodinng:
Closes RES-967.

Clears the ~285 ruff lint errors and ~105 unformatted files inherited from the orqkit repo split, then flips the CI `lint` job from advisory (`continue-on-error: true`) to a blocking gate.

## Approach (in the order the ticket recommends)
1. `ruff format src` — 122 files, single-quote style. Cleared ~250 whitespace/blank-line violations on its own.
2. `ruff check --fix` then `--fix --unsafe-fixes` — 32 mechanical fixes (TYPE_CHECKING import moves, `__all__` isort, `Self` return type, `RET504`, dummy-var renames, `logging.error`→`logging.exception`).
3. The remaining 10 by hand, where judgment was needed:
   - `string_contains_evaluator` made keyword-only (FBT001/FBT002).
   - `# noqa: RUF029` on the async `scorer` closures — async is the evaluatorq scorer contract, not incidental.
   - `Yields:` sections added to the tracing context managers (DOC402).
   - module-level `assert` → `raise ValueError` in `objective_generator.py` (S101).
   - `# noqa: RUF067` on the lazy-import table backing `simulation.__getattr__`.

The TYPE_CHECKING moves were verified type-only (TypedDicts used solely in annotations under `from __future__ import annotations`).

## Verification (all green locally)
- `ruff check src` → All checks passed
- `ruff format --check src` → 176 files already formatted
- `basedpyright` → 0 errors, 0 warnings
- `pytest -m 'not integration'` → 2627 passed, 2 skipped

Large diff (125 files) is almost entirely formatting — kept as its own PR per the ticket.


KarinaKKarinaK (approved):
Approving. Reviewed as a lint/format cleanup and independently re-ran the checks on the branch:

- `ruff check src` clean; `ruff format --check src` clean (176 files).
- `pytest -m 'not integration'` -> 2627 passed, 2 skipped, matching the PR.
- CI flip is correct: `continue-on-error` removed from both ruff steps and the job renamed. No required status-check rule keys off the old name, so the rename is safe.
- Spot-checked the non-mechanical hand fixes: the `assert` -> `if not X: raise ValueError` in `objective_generator.py` is equivalent (and safer under `-O`), the `# noqa: RUF029` on the async `scorer` / `with_llm_span` closures are correct (async is the contract), and the `Yields:` / docstring edits are non-behavioral.

One conscious-call, non-blocking note: `string_contains_evaluator` is public (exported from the package top level), so making its params keyword-only is technically breaking for anyone calling `string_contains_evaluator(False)` positionally. It is minor (both args are config flags, almost always passed by keyword) and the FBT rule is legit, so I would keep it, but worth a line in the release notes so it is not a silent break. `with_llm_span` got the same change but is internal, so no external impact.

Ready to merge from a code standpoint.

PR #55, opened 2026-07-03
docs(simulation): explain running agent sim from existing traces and data

currentlycodinng:
## What

Adds a **"From existing traces and data"** section to the agent simulation guide ([`docs/guides/agent-simulation.md`](docs/guides/agent-simulation.md)), documenting how to run agent sim seeded from data you already have. Closes RES-977.

Two workflows, matching what the ticket describes:

1. **Replay stored datapoints** — persist cases (`eq sim generate --output`, `eq sim run --save-datapoints`), re-run the identical bank against any target (`eq sim simulate --datapoints`), or pull them from an Orq dataset (`simulate(dataset_id=...)`). This is the reproducible "same cases, new agent version / new evaluators" path.
2. **Ground new cases in real traces** — the ticket's "generate new data based on previous data": distill recurring archetypes and situations out of production traces into seed phrases, feed them to `generate_personas()` / `generate_scenarios()`, then `simulate()`.

A `!!! note` calls out that reading archetypes out of traces is manual today (no built-in trace-to-persona extractor), marking where the "perhaps we also need to add capabilities" from the ticket would attach.

## Verification

- `uv run --group docs mkdocs build --strict` passes clean.
- Every symbol/flag in the examples verified against the code: `simulate` accepts `datapoints`/`dataset_id`/`personas`/`scenarios`; `load_datapoints_from_jsonl` importable from `evaluatorq.simulation.utils`; `--output` (generate) vs `--save-datapoints` (run) vs `--datapoints` (simulate) mapped to the right subcommands; `--target agent:<key>` confirmed.

Docs-only change, no code touched.

### DIFF
--- a/docs/guides/agent-simulation.md
+++ b/docs/guides/agent-simulation.md
@@ -310,6 +310,132 @@ The `target_callback` is the only structural difference from the Orq path —
 personas, scenarios, criteria, and the result shape are identical. Swap the
 callback body for any HTTP/LLM agent.
 
+## From existing traces and data
+
+You do not have to invent every test case from scratch. If you already have
+recorded conversations, real production traces, or a batch of datapoints from an
+earlier run, you can feed that history back into simulation in two ways: replay
+the exact same cases, or mine them for the archetypes that drive fresh ones.
+
+### Replay stored datapoints
+
+A `SimulationDatapoint` bundles one persona, one scenario, and the opening
+message. Every case simulation runs is one of these, and you can persist them for
+reuse. `eq sim generate` writes the cases it builds to a JSONL file with
+`--output PATH` (one datapoint per line); `eq sim run` does the same alongside a
+live run with `--save-datapoints PATH`:
+
+<!-- termynal -->
+
+```bash
+# Generate cases once and keep them
+eq sim generate --agent-description "e-commerce support agent" \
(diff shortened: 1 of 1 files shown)


KarinaKKarinaK:
Reviewed. Docs-only, and I re-checked the symbols and flags against the code on this branch. They all line up:

- `simulate()` params (`datapoints`, `dataset_id`, `personas`/`scenarios`, `max_turns`, `evaluator_names`) match the signature, and the mutual-exclusivity note matches the actual `Pass exactly one of dataset_id, datapoints, or personas+scenarios` guard.
- `load_datapoints_from_jsonl` is exported from `evaluatorq.simulation.utils`.
- `generate_personas` / `generate_scenarios` take `seeds` + `agent_description` and return one per seed, so the 3x3 to 9 note is right.
- `SimulationResult.goal_achieved` exists, and `goal_achieved` / `criteria_met` are the actual default evaluators.
- CLI checks out: `eq` entry point, `sim generate --output/--num-personas/--num-scenarios`, `sim run --save-datapoints`, `sim simulate --datapoints`, `--target agent:<key>`.

Accurate and well scoped. Nothing to change. LGTM.

KarinaKKarinaK (approved):
Verified all documented symbols, flags, and commands against the code. Accurate and well scoped. Approving.

PR #56, opened 2026-07-03
feat: add pairwise (preference) judging to evaluatorq (RES-760)

currentlycodinng:
## Summary

Adds pairwise (preference) judging to evaluatorq per **ADR-24** / RES-760: a panel of judges compares two responses A vs B and picks a winner, position-bias-corrected by running both orderings.

The canonical implementation lives here; research (RES-759) consumes it by import.

## What's included

- **`evaluatorq.pairwise`** — the ordering-independent core:
  - `reconcile_pair` (consistency gate: consistent pick votes, flip abstains + is recorded, consistent tie is a tie vote)
  - `pairwise_consensus` (plurality of reconciled votes → `A`/`B`/`tie`/`inconclusive`, reusing `_plurality_vote`)
  - `run_pairwise` — runs the panel through the shared `run_jury` in each ordering (swap on by default), un-swaps, reconciles per judge
  - `PairwiseVote` / `PairwiseComparison` / `JudgeStats` / `PairwiseReport` + `build_report` (headline A/B win rates, per-judge lean, position bias, tie rate, mean inter-judge agreement)
- **`llm_jury_pairwise()` → `PairwiseComparator`** — the public entry point. A sibling of `llm_jury()` (not a flag), reusing `run_judge`, the verdict-model builder, and `run_pairwise`. Verdict is `value` (`A`/`B`/`tie`) + `explanation`, matching the existing shape (no confidence field, per ADR-24).
- Pairwise API exported from the package top level; 2-judge integration test added.

## Design decisions (ADR-24, confirmed with Bauke)

- **No confidence field** — keeps the verdict format identical to the shared `JuryVote`/`Prediction` contracts.
- **Plain win-rate ranking only** — BT-σ (RES-1024) and cyclic round-robin (RES-1023) stay deferred.
- **Swap on by default** — position bias is a goal; opt out disables the bias metric.

## Reuse-mandate flags (need Bauke's call, non-blocking)

- **`judge_pool` in a "general eval config"** — no general `EvaluatorConfig` exists; the panel is already exposed via the `judges` kwarg (same names as redteam). Read as already-satisfied.
- **`strict_panel` parity** — it's redteam-specific panel-composition validation, not a `run_jury` passthrough. Flagged rather than ported. See the RES-760 comment thread.

## Tests

23 new unit tests (strict TDD), full suite **2652 passing / 0 regressions**, ruff + basedpyright clean.


arianpasquali (changes requested):
Hey @currentlycodinng, really nice addition! The core design here is solid: `reconcile_pair`/`pairwise_consensus` as pure ordering-independent functions is the right factoring, the swap/un-swap logic is correct end-to-end (I traced the sign handling through both orderings. it checks out), and the unit tests for the reconciliation core are thorough and readable. Reusing `run_jury` instead of building a parallel panel runner was the right call. 🙌

I did a deep pass and found a handful of things worth addressing before merge, the top four are confirmed with runnable repros. Grouped by impact:

### Correctness, the orchestration params don't do what callers expect

1. **`min_successful_judges` never gates the winner** (`pairwise.py:186`). It's validated and forwarded to `run_jury`, but `run_pairwise` only reads `jury.votes` and ignores the deliberation's `inconclusive` flag — so `judges=[j1,j2,j3], min_successful_judges=3` with two judges erroring still returns a confident `winner='A'` from the lone survivor (reproduced). The same config through `llm_jury` returns `inconclusive`. Suggest enforcing the quorum on the *reconciled decisive votes* (which is arguably the more meaningful unit here anyway).

2. **`replacement_judges` is effectively inert under `swap=True`** (`pairwise.py:217`). The reconciliation loop iterates `first_votes` only, and replacements are promoted independently per ordering. A stand-in that fires only in the second ordering is silently dropped — LLM call paid, vote discarded (reproduced: votes list contains only the failed primary, winner `inconclusive` despite a healthy stand-in verdict). One that fires only in the first ordering reconciles to an abstain. A replacement can only contribute a vote if the same primary fails in *both* orderings. Worth either pairing the orderings per judge or documenting the limitation loudly.

3. **`propagate_errors` uses the raw panel, `llm_jury` uses the deduped one** (`llm_jury.py:543` vs `:435`). `judges=['m','m']` dedups to a lone judge inside `run_jury`, but `len(self._panel) == 1` is False — so an outage is swallowed into per-comparison `inconclusive` instead of raising loudly. Passing the already-computed `deduped` into the comparator fixes it and removes the divergence.

4. **`criteria=''` slips through** (`llm_jury.py:583`). The `is not None` check means an empty string bypasses both the default and any error — judges get a blank `# Criteria` section (reproduced). `llm_jury` raises on this; a falsy check (or raising) would match.

### Report metrics can mislead when the panel degrades

5. **`PairwiseReport` has no inconclusive count** (`pairwise.py:86`). 99 inconclusive comparisons + 1 A-win reads as `a_win_rate=1.0` — a landslide from a run that was 99% noise (reproduced with 3 flip-outs + 1 decisive → `a_win_rate=1.0`, `tie_rate=0.0`). An `inconclusive_rate` (or `decisive` count) alongside the win rates would make degraded runs visible.

6. **`mean_agreement` inflates on degraded panels** (`pairwise.py:129`). A comparison with exactly one decisive vote scores agreement 1.0 — highest exactly when the panel is most broken. Maybe require ≥2 decisive votes per comparison.

7. **`position_bias` denominator includes comparisons where flipping was impossible** (`pairwise.py:123`) — failed/missing orderings dilute the flip rate. Flips over *completed pairs* would be the honest measure.

### Smaller items

- Both orderings run sequentially (`pairwise.py:186`, `:206`) — they're independent, so `asyncio.gather` would halve per-comparison latency at no extra cost.
- Token usage and judge explanations are discarded (`PairwiseVote`/`PairwiseComparison` carry neither) — breaks parity with `llm_jury`'s cost accounting, and we pay for 2-3 sentence explanations per call then drop them.
- A caller-supplied `judge_fn` returning a non-`'A'/'B'/'tie'` value (e.g. a bool) leaks into `winner` as `'True'` (`pairwise.py:155`, reproduced) — since `run_pairwise` is a public export, a small validation would keep the documented contract.
- The validation block and inner `judge_fn` in `llm_jury_pairwise`/`PairwiseComparator` are near-verbatim copies of `llm_jury`'s (and have already diverged once, see #3) — worth extracting shared helpers. `self._criteria` is also never read after `__init__`.
- The integration test asserts an exact winner from two live LLMs with an even panel — one `tie` verdict makes it fail nondeterministically; asserting `winner in ('A', 'inconclusive')` or using 3 judges would stabilize it.
- No unit coverage yet for `replacement_judges` / `min_successful_judges` / `repetitions>1` through `run_pairwise` — which is exactly where #1 and #2 hide. A couple of tests there would lock in the fixes.
- Tiny: `tests/unit/test_pairwise_jury.py:13` and `test_pairwise_runner.py:88` fail `ruff format --check` (CI only gates `src`, so not blocking).
- New public API (`llm_jury_pairwise`, `run_pairwise`, `build_report`, …) isn't in the docs site yet — a short page on the swap/reconcile semantics would preempt "why is everything inconclusive?" questions.

None of this undermines the approach — the consistency-gated reconciliation is exactly right per ADR-24, and the fixes are all local. Happy to pair on any of them. Great work! 🚀

arianpasquali:
## Feedback on `mean_agreement` in `PairwiseReport`

Nice work overall — the consistency-gated reconciliation and the report shape read well. One metric needs attention before merge, one follow-up to log.

### 1. Single-vote comparisons inflate `mean_agreement` (fix before merge)

`build_report` computes per-comparison agreement over the decisive reconciled votes and averages, but a comparison with **one** decisive vote scores `1.0` — a judge "agreeing" with itself. Redteam's own reliability code already excludes these: `_krippendorff_alpha_binary` skips units with `m < 2` votes (`redteam/reports/converters.py` — the "pairable" rule). `mean_agreement` should follow the same precedent.

This isn't cosmetic: the worst case is exactly the case the metric is supposed to catch. Heavy position bias → most judges flip out → many 1-vote comparisons → `mean_agreement` goes **up** while the panel is at its least reliable. Inverted signal.

Suggested one-liner:

```python
agreements = [
    rate
    for c in comparisons
    if len(decisive := [v.vote for v in c.votes if v.vote is not None]) >= 2
    and (rate := _agreement_rate(decisive)) is not None
]
```

### 2. Document it as *raw* agreement, not reliability

Modal-vote share is not chance-corrected. With three categories, three judges voting uniformly at random land at an expected modal share of ~0.63 — so `mean_agreement = 0.65` reads as "decent agreement" while being statistically indistinguishable from coin-flipping. Suggest the field description says "raw (not chance-corrected) inter-judge agreement" so nobody reads it as a reliability score.

Related: RES-760's DoD asks for "pairwise reliability reported (adapted agreement, or documented why the existing one doesn't fit)", and ADR-24 lists `raw_agreement` **and** `JuryReliability` (Krippendorff α) for inter-judge agreement. This PR ships the raw half only — fine for v1, but the deviation should be stated somewhere (this PR description or the ADR).

### 3. Follow-up ticket: nominal Krippendorff α (not blocking)

The chance-corrected metric ADR-24 names is a small, additive extension of what redteam already has: generalize `_krippendorff_alpha_binary` to nominal multi-category (sum over unordered category pairs instead of the single 0/1 pair), lift it into `common/`, and let both redteam (binary = special case) and pairwise reuse it. Units = comparisons, coders = judges, categories = `{A, B, tie}`, abstentions = missing data — α handles missing ratings and variable panel sizes natively, which is exactly the post-reconciliation situation. Also feeds RES-1024 (BT-σ), which needs per-judge reliability as input.

For the record, RES-978's per-output-type formulas (flip rate / entropy / stdev) were considered and don't apply here — that's intra-judge stability across repetitions, a different axis than inter-judge agreement; pairwise's analog of its flip-rate is already shipped as `position_bias`.

KarinaKKarinaK:
Reviewed deeply: checked out the branch, ran it, and verified the logic against the code rather than the description.

Solid work. What I confirmed:

- Position-bias correction is correct. In the swapped ordering the judge picks by slot, `_unswap` maps it back to the canonical A/B frame, and `reconcile_pair` abstains and records a flip when the two orderings disagree. A slot-following judge is caught, a genuinely consistent one votes. Traced it by hand and the tests (`test_position_biased_judge_flips_and_abstains`, `test_consistent_judge_wins_for_the_better_response`) pin it.
- Consensus, replacement promotion (run in both orderings, capped at the failure count), `min_successful_judges` gating, and the report degradation guards (single decisive vote excluded from mean agreement, position bias only over flippable pairs) all read correctly.
- `_decisive_value` rejects out-of-contract verdicts, which matters since `run_pairwise` is a public export.
- Exports, mkdocs nav, and the docs page are all in place. Imports resolve and the 33 pairwise unit + integration tests pass locally.

One non-blocking note: the `llm_jury.py` diff folds a file-wide double-to-single quote reformat into the feature, which buries the real changes. Splitting pure formatting into its own commit would make future feature diffs easier to review. Not a fix for this PR.

LGTM.

KarinaKKarinaK (approved):
Reviewed the logic deeply against the code (position-bias correction, reconciliation, consensus, report guards), exports and docs in place, tests pass locally. Approving.

arianpasquali (approved):
It looks good now. The main issues were solved. good to go.

PR #58, opened 2026-07-06
docs: remove internal-spec leakage from public site (RES-1004)

KarinaKKarinaK:
## What

Removes internal review-notes leakage from the public docs (RES-1004).

The `configuration.md` "## Notes" section read like an author's audit and shipped to the public site:

- Dropped the `EVALUATORQ_OWASP_DATASET_ID` note — a `CLAUDE.md`-vs-source audit line ("listed in `CLAUDE.md` but not present in the current source tree…") that is meaningless to public readers, who have no `CLAUDE.md`. Confirmed the var exists nowhere in `src/`.
- Folded the `ROUTER_BASE_URL` deprecation into the env-var table as a **deprecated row** rather than a free-text note. Verified against the code (`src/evaluatorq/redteam/backends/registry.py:43-44`): it is still handled and logs a "no longer supported" warning, so it belongs in the table as deprecated (not removed).

## The other checkbox — already resolved

`docs/redteam-dashboard-only-views.md` (the internal FastHTML spec) was already removed from `main` via the dashboard consolidation (RES-1006). It no longer exists, nothing references it, it is not in nav, and `mkdocs.yml` already carries `exclude_docs: superpowers/`, so the dead `docs/superpowers/specs/...` references are moot. No change needed there.

## Scope

One file: `docs/configuration.md` (+1 / -5).

## Out of scope

`docs/stylesheets/extra.css` has two developer comments referencing `/DESIGN.md`. Those are source comments in a stylesheet, not readable/indexed doc prose, and outside the ticket's file list — left as-is.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -8,6 +8,7 @@ All configuration is via environment variables. No config file is required.
 |---|---|---|---|
 | `ORQ_API_KEY` | Required for Orq features | — | Authenticates against the Orq platform. Required to fetch datasets, upload results, and invoke deployments. Also auto-enables OpenTelemetry tracing (spans are sent to `https://my.orq.ai/v2/otel`). |
 | `ORQ_BASE_URL` | No | `https://my.orq.ai` | Overrides the Orq API base URL. Affects Orq SDK calls (dataset fetch, deployment invocation) and the derived OTLP tracing endpoint (`<ORQ_BASE_URL>/v2/otel`). Does **not** redirect OpenAI-compatible LLM calls — use `OPENAI_BASE_URL` for that. |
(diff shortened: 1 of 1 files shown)


currentlycodinng (approved):
Verified both changes against source:

- **`ROUTER_BASE_URL`** — `src/evaluatorq/redteam/backends/registry.py:43-44` matches the new deprecated table row precisely: it warns when the var is set without `ORQ_BASE_URL`, and the value is never honoured. Correct to fold the free-text note into the env table as a deprecated row rather than drop it.
- **`EVALUATORQ_OWASP_DATASET_ID`** — grep confirms it's absent from all of `src/` (and not in `CLAUDE.md`), so the audit-style note was noise for public readers. Correct to remove.

Clean `+1/-5`, no leftover internal-spec references in the touched file. LGTM.

Baukebrenninkmeijer (approved):

PR #62, opened 2026-07-08
docs(mkdocs): make docs agent-browsable via llms.txt (RES-1037)

KarinaKKarinaK:
## What

Makes the evaluatorq docs agent-browsable per the `llms.txt` convention (RES-1037), so coding agents can read the docs as markdown instead of scraping HTML.

Adds the `mkdocs-llmstxt` plugin, which emits at build time:

- **`/llms.txt`** — a curated markdown index (H1, summary, sections mirroring the nav: Getting started / Guides / Setup, each with linked pages + one-line descriptions).
- **`/llms-full.txt`** — all docs concatenated as one markdown file.
- **per-page `<page>/index.md` endpoints** — clean markdown for each page.

Combined with the `sitemap.xml` that mkdocs-material already auto-generates, this covers the agent-browsable checklist.

## Verification

`mkdocs build --strict` (the CI gate) exits 0 and produces `llms.txt`, `llms-full.txt`, and 12 per-page `.md` files. The `uv.lock` update is exactly the plugin's dependency tree (the `--frozen` docs CI needs it).

## Deliberate calls

- **No `robots.txt`**: the docs are a GitHub Pages project site under `/evaluatorq/`, and crawlers only read `robots.txt` from the domain root (a different repo), so one here would land at the subpath and be ignored. The sitemap + llms.txt are the effective pieces.
- **`docs:` commit type** so no PyPI release is triggered.
- Fits in one ticket; no subtickets needed.

## To confirm after merge

Once the docs redeploy, point `isitagentready.com` at the full path `orq-ai.github.io/evaluatorq/` (not the org root, which 404s per the ticket).


Baukebrenninkmeijer:
## Review: llms.txt generation (RES-1037)

Build passes strict and the output is correct — spec-compliant `llms.txt`, links resolve to real served `.md` siblings, `llms-full.txt` is usable markdown. Below are validated findings (each checked against the plugin source and generated output), roughly in priority order.

### 1. `sections` is a hand-maintained parallel copy of `nav:` with no drift guard
`llmstxt.sections` (mkdocs.yml:85-105) restates the `nav:` tree (mkdocs.yml:174-196) with independently-authored labels — they already disagree (`nav` "Reference" vs llms "Setup"; custom-evaluators grouped differently). A page present in `nav` but missing from `sections` is silently absent from `llms.txt`, and `mkdocs build --strict` stays green (the plugin logs through its own logger, not MkDocs' plumbing). Every doc change is now a two-file edit with nothing enforcing parity.

**Already happened:** the `About` group (`contributing.md`, `changelog.md`, `roadmap.md`, nav lines 192-195) is absent from `llms.txt` in the initial commit.

Fix: generate `sections` from the same data `gen_pages.py`/`nav` already hold, **or** add a build-time parity check (nav pages ⊆ sections) so `--strict` fails on a miss. If About is excluded on purpose, add a comment saying so — right now it's indistinguishable from an oversight.

### 2. `full_output` (llms-full.txt) is a partial dump, not "all docs"
`mkdocs-llmstxt` concatenates only pages listed in `sections` (plugin.py:198-202 iterates `self._sections`). Anything not in `sections` — the whole About group — is dropped from `llms-full.txt` too (verified: contributing/changelog/roadmap page content is not in the generated file). Either drive it off nav so it's honestly full, drop it if there's no consumer, or rename the intent to "curated primary docs."

### 3. Discovery gap — nothing advertises `/llms.txt`
Goal is "agent-browsable," but the built site does not point agents at the file: no `<link rel>` in `<head>` (0 hits in `site/index.html`), no `robots.txt`, no README mention. Only agents that already blind-probe the convention benefit. One `<link>` in the head (via `extra`/overrides), a `robots.txt` line, or a README pointer closes it.

### 4. API-reference pages are lossy HTML→markdown round-trips
The plugin markdownifies rendered HTML (`on_page_content`). For prose that's fine; for `reference/**` it degrades the highest-value content — `site/reference/evaluatorq/index.md` has a params table with a **blank Type column** (types survived only in mkdocstrings' HTML structure, which markdownify flattened). Consider excluding `reference/**` from llms.txt, or sourcing the API surface from docstrings/signatures directly.

### 5. Reuse + glob nits
- `gen_pages.py` already holds `_PACKAGE_DESC` (API blurbs, line 123), `_CATEGORY_DESC` (line 214), and `_docstring_summary()` (line 234). The llms `API reference`/`Examples` sections ignore these and use description-less globs, while static pages get hand-written descriptions — inconsistent, and those hand-written blurbs paraphrase each page's own lead paragraph (a third copy that can rot).
- `examples/**/*.md` re-includes demo scaffolding that `gen_pages.py` deliberately excludes via `_EXCLUDE_FILES` — two competing inclusion policies. Flat titles like "Run"/"Compare" are meaningless out of tree.
- `**` in `examples/**/*.md` is dead syntax: the plugin uses `fnmatch.filter` (plugin.py:70) where `*` already spans `/`. `examples/*.md` matches identically. Reads as POSIX recursive-glob but isn't.
- The `mkdocs.yml:78` comment ("Placed after mkdocstrings so the rendered API pages are available") misdescribes the mechanism — placement matters for `on_page_content` declaration order, not a build-phase dependency. Correct placement, wrong rationale.
- `markdown_description` restacks `site_description` — two near-identical capability blurbs in the header.

### Bottom line
Ships working, correct files. Main concern is #1/#2: a hand-maintained parallel catalog with no lint keeping it honest, and a "full" dump that isn't full. #3 leaves the discovery half of "agent-browsable" unbuilt.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

Baukebrenninkmeijer:
## Re-review after review-response commits (520551a…98e6d98)

Pulled latest and re-verified against a strict build. The response is thorough — nearly everything from the first pass is addressed, and one of my findings was simply wrong. Status per item:

**#1 nav ↔ sections drift — RESOLVED.** `docs/hooks.py` adds an `on_nav` parity guard that raises `PluginError` (fails `--strict`) when a nav page isn't covered by a `sections` entry, with `_PARITY_EXEMPT` for the SUMMARY/index scaffolding. The two lists can't silently diverge again. nav was also reorganized so its group labels (Guides/Setup) match the sections taxonomy. Verified: strict build green with the guard active.

**#2 `full_output` mislabel — RESOLVED.** `About` (contributing/changelog/roadmap) is now in `sections`, so "all docs" is actually true, and the parity guard keeps it honest. Verified: About pages now present in `llms.txt` (3 hits) and `llms-full.txt`.

**#3 discovery gap — RESOLVED.** `overrides/main.html` injects `<link rel="alternate" type="text/markdown" href=".../llms.txt">` into every page head (verified: 1 hit in `site/index.html`), plus a README pointer to `/llms.txt` and `/llms-full.txt`. Agents no longer have to blind-probe.

**#4 lossy API round-trip — ACKNOWLEDGED / deferred.** A NOTE comment in `mkdocs.yml` documents that markdownified mkdocstrings HTML flattens the param-type column, points agents at the served HTML for authoritative signatures, and defers a docstring-sourced dump to a follow-up. Reasonable call — the info loss is real but the pages are still useful, and the honest comment is the right interim.

**#5 nits:**
- Ordering-comment rationale — FIXED (now correctly says declaration-order, not a build-phase dependency).
- `markdown_description` duplicate blurb — FIXED (removed entirely).
- **My "`**` is dead fnmatch syntax" claim was WRONG — retracting it.** The added comment is correct: `examples/**/*.md` forces a subdirectory, so it matches example content at any depth while excluding top-level `examples/index.md` / `SUMMARY.md`; `examples/*.md` would wrongly pull those in (fnmatch `*` spans `/`). Verified directly:
  ```
  examples/**/*.md -> ['examples/redteam/foo.md', 'examples/a/b/c.md']
  examples/*.md    -> ['examples/index.md', 'examples/SUMMARY.md', 'examples/redteam/foo.md', ...]
  ```
  The `**/` is load-bearing, not dead. Good catch.
- Reuse — `api-reference.md` now reuses the generated `index_rows`/`_PACKAGE_DESC`, and `cli-reference.md` was split into overview/redteam/simulation pages, each listed explicitly.

Net: all blocking findings resolved, #4 sensibly deferred with a documented rationale, and one of my findings corrected. Strict build passes. LGTM from my side.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

Baukebrenninkmeijer (approved):
All first-pass findings resolved (parity guard, About in sections, discovery link), #4 sensibly deferred with documented rationale. Strict build + full CI green. LGTM.

PR #64, opened 2026-07-08
docs(mkdocs): orq.ai-branded social preview card (RES-1088)

KarinaKKarinaK:
## What

The docs link-unfurl (Slack / Twitter / OpenGraph) rendered as the generic default Material card. This replaces it with a hand-designed, orq.ai-branded preview.

og-image

## Approach

- A single hand-designed 1200x630 card (`docs/assets/og-image.png`), matching the orq.ai brand: the signature teal/cyan aurora on a warm off-white ground, the monochrome Orq mark + `orq.ai` lockup, near-black Kurrent display type, capability pills, and a `pip install` line.
- Uses the **real self-hosted brand fonts** (Kurrent display, Avio Sans body, Kurrent Mono) — the mkdocs `social` plugin can't, since it only sources faces from Google Fonts. So the `social` plugin is dropped and the card is wired via `overrides/main.html` (`og:`/`twitter:` meta on every page).
- Regenerable and reproducible via `scripts/gen_og_card.py` (committed).

## Files

- `scripts/gen_og_card.py` — card generator (run: `uv run --with "fonttools[woff]" --with cairosvg --with pillow python scripts/gen_og_card.py`)
- `docs/assets/og-image.png` — the committed card
- `overrides/main.html` — sets the og/twitter meta site-wide
- `mkdocs.yml` — drop `social` plugin, add `custom_dir: overrides`

## Test

- Local: `open docs/assets/og-image.png`, or `mkdocs serve` and view-source for the `og:image` meta.
- Live unfurl resolves once merged to `main` and the Pages deploy runs (deploy is skipped on PR branches). Then paste the docs URL in Slack, or check opengraph.xyz.

## Notes

- Every page shares this one card (standard for docs sites).
- The docs workflow still installs Cairo (for the old `social` plugin); now unused — can be stripped in a follow-up to speed CI.


Baukebrenninkmeijer:
Nice, looking great. Couple of points:

* I like the orange highlighted 'q' in 'evaluatorq', that we have for example here: 

  Could you add that?

  
* and perhaps my preference, but some of our orange tint in the other accents? 
* The pills look a bit like labels/tags, not necessarily like product features. 

  Do you think there's a way we can make this more indicative of: "This is what this library contains"?

Baukebrenninkmeijer (changes requested):
see above feedback

KarinaKKarinaK:
Updated preview with feedback applied:
Screenshot 2026-07-10 at 8 23 25 AM

Baukebrenninkmeijer (commented):
nice, looking good. Can you make the 3 buttons the same colour. Other than that, approved.

KarinaKKarinaK:
Final version:
image

Baukebrenninkmeijer (approved):

PR #68, opened 2026-07-13
feat(dashboard): side-by-side sim run comparison (RES-1085)

KarinaKKarinaK:
Implements side-by-side comparison of two agent-sim runs (RES-1085). Dashboard sim views were single-run only; this adds a compare surface on top of the same `sim` adapter/library. No changes to the sim runner or data model.

## What
- **Entry point:** a "Compare runs" bar (Run A / Run B dropdowns + button) on the Agent Sim overview → links to `/compare/sim?a=&b=`.
- **New page `GET /compare/sim`:** editorial header, KPI delta cards (goal-achieved rate, mean score, mean turns, conversations), an all-metrics table (aggregates + per-scorer + terminated-by, each with A / B / delta), an outcome-diff table, and a transcript-compare slot.
- **Charts (server-rendered inline SVG):** Outcomes (A vs B), Scorer averages (A vs B), How conversations ended (A vs B), and a Per-conversation score-change diverging chart. SVG on purpose — no JS and no `/static` dependency, so they render in the page, in tests, and in static HTML exports.
- **Outcome diffs:** conversations matched by `(persona, scenario)`; goal flips highlighted, score/turn deltas shown, and unmatched conversations listed as "Only in A / Only in B" so a mismatched dataset is visible rather than silently dropped.
- **Transcript compare:** `GET /compare/sim/transcript` renders both runs' transcripts side by side (reuses the single-run transcript renderer).

## Scope
Additive only. New `dashboard/sim_compare.py` + `tests/dashboard/test_sim_compare.py`; a compare route registered in `app.py`; an optional `compare_choices` arg on `sim_overview_body` (defaults to hidden). Nothing else in the dashboard is touched — no changes to Dashboard, Red Team, or single-run Agent Sim.

## Tests
- 7 new tests in `test_sim_compare.py`: KPI delta math, `(persona, scenario)` matching incl. unmatched handling, the compare page (KPIs + SVG charts + diffs + flips + unmatched), the transcript route, 400/404 error paths, and the compare bar on the overview.
- Full dashboard suite: 312 passed. `ruff check src` clean, `basedpyright` 0 errors.

## Known limitation (pre-existing, not from this PR)
The transcript-compare click uses HTMX. In this preview dashboard, `head_assets()` points assets at `/static/…` but `build_app` serves them at root, so HTMX (and any Vega chart) does not load when served via `serve()`. The charts here avoid this entirely by being inline SVG; the transcript drill-down still needs it. This affects the whole dashboard, not just this page — worth a small follow-up to align the static route with the asset URLs.

### DIFF
--- a/src/evaluatorq/common/reports/vega.py
+++ b/src/evaluatorq/common/reports/vega.py
@@ -382,8 +382,7 @@ def vl_grouped_bar(
         for i in range(len(categories))
         if i < len(vals)
     ]
-    return {
-        'data': {'values': rows},
+    bar_layer: dict[str, Any] = {
         'mark': {'type': 'bar', 'tooltip': True},
         'encoding': {
             'y': {'field': 'cat', 'type': 'nominal', 'sort': None, 'title': None},
@@ -391,6 +390,22 @@ def vl_grouped_bar(
             'yOffset': {'field': 'series', 'type': 'nominal'},
             'color': {'field': 'series', 'type': 'nominal', 'legend': {'title': None}},
         },
+    }
+    # A zero-value bar has zero width and renders as nothing, which is
+    # indistinguishable from "not measured" — label those bars with a literal 0.
+    zero_label_layer: dict[str, Any] = {
+        'transform': [{'filter': 'datum.value === 0'}],
+        'mark': {'type': 'text', 'align': 'left', 'dx': 3, 'fontSize': 9, 'color': COLORS['ink_700']},
+        'encoding': {
+            'y': {'field': 'cat', 'type': 'nominal', 'sort': None, 'title': None},
+            'x': {'field': 'value', 'type': 'quantitative', 'title': x_title},
+            'yOffset': {'field': 'series', 'type': 'nominal'},
+            'text': {'value': '0'},
+        },
+    }
+    return {
+        'data': {'values': rows},
+        'layer': [bar_layer, zero_label_layer],
         'width': 420,
         'height': {'step': 16},
     }
--- a/src/evaluatorq/common/reports/vega.py
+++ b/src/evaluatorq/common/reports/vega.py
@@ -28,16 +28,22 @@ from evaluatorq.common.reports.palette import COLORS, QUALITATIVE
 
 ORQ_VL_CONFIG: dict[str, Any] = {
     'background': 'transparent',
-    'font': 'Inter, system-ui, -apple-system, sans-serif',
+    # vl-convert renders server-side and has no access to the brand webfonts; it
+    # also can't resolve 'system-ui'/'-apple-system' and would fall back to a
+    # serif. List concrete sans families so it lands on a clean sans everywhere.
+    'font': 'Inter, Helvetica Neue, Helvetica, Arial, sans-serif',
     'axis': {
         'labelColor': COLORS['ink_700'],
         'titleColor': COLORS['ink_700'],
         'gridColor': COLORS['sand_400'],
+        'gridOpacity': 0.6,
         'domainColor': COLORS['sand_400'],
         'tickColor': COLORS['sand_400'],
     },
     'legend': {'labelColor': COLORS['ink_700'], 'titleColor': COLORS['ink_700']},
     'view': {'stroke': 'transparent'},
+    # Softly rounded bar ends read cleaner and match the dashboard's card radii.
+    'bar': {'cornerRadiusEnd': 3},
     'range': {'category': QUALITATIVE},
 }
 
--- a/src/evaluatorq/dashboard/app.py
+++ b/src/evaluatorq/dashboard/app.py
@@ -254,7 +254,12 @@ def build_app(roots: list[Path] | None = None) -> FastHTML:
         # Tabbed body for the known surfaces (Streamlit-aligned); the interactive
         # panels live inside their tabs, so they are no longer appended separately.
         if surface == 'sim':
-            body_html = report_tabs.sim_report_tabs(rid, report_obj)
+            from evaluatorq.dashboard.view import sim_run_compare_control
+
+            # Same choice list as the overview picker (sim only, no error runs,
+            # capped); the control itself drops the current run from the options.
+            choices = [(c.id, c.name) for c in library.scan(roots) if c.surface == 'sim' and not c.error][:100]
+            body_html = report_tabs.sim_report_tabs(rid, report_obj, compare_html=sim_run_compare_control(rid, choices))
         elif surface == 'redteam':
             body_html = report_tabs.redteam_report_tabs(rid, report_obj)
         else:
--- a/src/evaluatorq/dashboard/app.py
+++ b/src/evaluatorq/dashboard/app.py
@@ -50,6 +50,7 @@ from evaluatorq.dashboard.filter_request import parse_selections
 from evaluatorq.dashboard.filters import FILTERS, apply_or_all
 from evaluatorq.dashboard.redteam_views import register_redteam_view_routes
 from evaluatorq.dashboard.shell import page
+from evaluatorq.dashboard.sim_compare import register_sim_compare_routes
 from evaluatorq.dashboard.sim_views import register_sim_view_routes
 from evaluatorq.dashboard.surfaces import ADAPTERS
 from evaluatorq.dashboard.view import (
@@ -171,7 +172,8 @@ def build_app(roots: list[Path] | None = None) -> FastHTML:
         label = SURFACE_LABELS.get(surface, 'Reports')
         pg, per_pg = _paging(req)
         if surface == 'sim':
-            body = sim_overview_body(metrics.sim_overview(roots, page=pg, per_page=per_pg))
+            sim_choices = [(c.id, c.name) for c in library.scan(roots) if c.surface == 'sim']
+            body = sim_overview_body(metrics.sim_overview(roots, page=pg, per_page=per_pg), compare_choices=sim_choices)
         elif surface == 'redteam':
             body = redteam_overview_body(metrics.redteam_overview(roots, page=pg, per_page=per_pg))
         else:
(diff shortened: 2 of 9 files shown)


Baukebrenninkmeijer (changes requested):
We updated a lot of the dashboard yesterday for the demo, so let's rebase this on that, and validate this still holds for there.

Baukebrenninkmeijer:
## Review findings (side-by-side sim compare)

High-confidence issues, each independently flagged by 3+ review passes. Verified against the codebase.

### 1. Absent scorer rendered as a real `0.0` (data integrity)
`compare_kpis` and `_scorer_chart` take the **union** of scorer keys and fill missing ones with `0.0`. Comparing two runs scored by *different* evaluators (a normal reason to compare) shows e.g. `Scorer: safety = 0.00` with a full red downward delta for the run that never measured it — a fabricated regression. Missing scorers should render as `n/a` and be excluded from the delta arrows.

### 2. Self-compare is the default click
Both `<select>`s in `_sim_compare_bar` (`view.py`) default to the first option, and there's no `a == b` guard in the route (`sim_compare.py` `compare_sim`). Clicking **Compare** without touching the dropdowns yields run-vs-itself: all-zero deltas presented as a valid report. Default select B to the second run + reject/warn on `a == b`.

### 3. Hand-rolled SVG reinvents existing charting
`_svg_grouped_bars` / `_svg_diverging` (~150 lines) duplicate `common/reports/vega.py:68 render_svg()` + `vega.py:360 vl_grouped_bar()` — which already render **server-side SVG with no JS / no /static dependency** (the exact property the code comment claims as its reason to hand-roll). `redteam_charts.py` already uses this path; `html_helpers.svg_bar` even shares `_svg_diverging`'s exact signature as a no-vega fallback. Colors are hardcoded hex (`#df5325` = `COLORS['orange_400']`; `#0f9d6b` isn't in the palette) despite the "orq brand palette" comment. Routing through the existing helpers also fixes the `vmax=1.0` scorer-overflow bug (bars/`%` labels break when a scorer average exceeds 1.0).

### 4. Error/status contract is inconsistent and misleading
- Transcript error branches (invalid index, no conversation at index) return **HTTP 200** while the sibling run-missing branch correctly uses 404.
- `_load_run` collapses "absent", "wrong surface", and "load raised" into `None`, so a valid-but-corrupt/unparseable sim file is reported as *"not found or not a simulation run."* Distinguish load-failure from not-found and log it with a stack.

### 5. Duplicate `_entries` helper
`sim_compare._entries` is byte-identical to `sim_views._entries_from_run` — and `sim_compare` already imports two other symbols from that module. Import it and delete the clone. (Same story for `_panel`, which duplicates `view.py:164 _panel`, and the `cmp-*` CSS dumped inline instead of into `styles.py`.)

### 6. Picker + matching robustness at scale / edge cases
- Every sim overview render does **two** full directory scans (`library.scan` in `app.py` on top of the scan inside `metrics.sim_overview`); the dropdown is unbounded and includes `error`-flagged runs that route to a 404.
- Duplicate `(persona, scenario)` pairs are matched **positionally** by load order — arbitrary pairs shown as "the same conversation" with a confident flip/±score badge. Low-overlap datasets show a valid-looking hero + KPI band while the per-conversation diff silently collapses; add a match-rate summary/warning.

_Generated from a multi-agent adversarial review; happy path (two distinct runs, same dataset + scorers) works — these are the trust/robustness gaps._

KarinaKKarinaK:
Rebased on the updated dashboard (main) and addressed all six findings. Full non-integration suite: 2973 passed; `ruff check src` clean; `basedpyright` 0 errors.

**1. Absent scorer rendered as real 0.0** — `KpiRow.a/.b` are now `float | None`; a scorer only one run measured stays `None`, renders `n/a`, and carries no delta arrow. `_scorer_chart` now compares only scorers **both** runs measured and lists the one-run-only scorers as a note instead of charting a fake 0.0.

**2. Self-compare is the default** — both selects now default to *different* runs (A=first, B=second), and the route rejects `a == b` with a 400.

**3. Hand-rolled SVG** — deleted ~150 lines of custom SVG; charts now build Vega-Lite specs via the shared `vl_grouped_bar` / `vl_bar_h` and render server-side through `render_svg()` (no JS, no `/static`). Colors come from `palette.COLORS`. Vega autoscaling removes the `vmax=1.0` >1.0 overflow bug.

**4. Error/status contract** — added `_resolve_run()` that distinguishes not-found (`404`) from a corrupt/unparseable sim file (`422`, logged with a stack via `logger.opt(exception=True)`), instead of collapsing both into "not found". Transcript route: invalid index → `400`, missing conversation → `404`.

**5. Duplication** — now imports `sim_views._entries_from_run` (deleted the clone) and `view._panel`; the `cmp-*` CSS moved into `styles.py` (`_SIM_COMPARE_CSS`, appended to `DASHBOARD_CSS`) instead of an inline per-request `<style>`.

**6. Picker + matching robustness** — the dropdown now excludes `error`-flagged runs and is capped at 100 newest; added a code note that `library.scan` reuses the mtime-keyed cache metrics already warmed. Matching now surfaces a `N matched · X only in A · Y only in B` summary plus a **low-overlap warning** (<50% shared keys), and the docstring flags positional matching of duplicate `(persona, scenario)` pairs.

New tests cover: n/a for absent scorers, self-compare 400, corrupt-report 422, invalid-index 400, default-to-different-runs, and shared-scorers-only charting.

Baukebrenninkmeijer:
## Re-review — fix commit `622ed4a`

Checked out the PR head, re-ran the code against all six findings. **All addressed.** Suite: 12 passed (was 7), `ruff` clean, `basedpyright` 0 errors.

| # | Finding | Status |
|---|---------|--------|
| 1 | Absent scorer = `0.0` | ✅ `KpiRow.a/.b` now `float | None`; `sa.get(key)` no default; `_fmt` → `n/a`, no delta arrow. `_scorer_chart` charts `set(sa_a) & set(sa_b)` only, one-run scorers listed in a note. |
| 2 | Self-compare default | ✅ Route rejects `a == b` (400); picker defaults A=first, B=second (`_opts(0)`/`_opts(1)`). |
| 3 | Hand-rolled SVG | ✅ ~150 lines deleted; now `vl_grouped_bar`/`vl_bar_h` + `render_svg`. Colors from `palette.COLORS`. Vega autoscale kills the `vmax=1.0` overflow. |
| 4 | Error/status contract | ✅ mostly — `_resolve_run` returns `ok`/`missing`/`error` → 404 vs **422** (logged with stack); transcript 400/404. See note below. |
| 5 | Duplication | ✅ imports `_entries_from_run` + `view._panel`; `cmp-*` CSS moved to `styles.py` (`_SIM_COMPARE_CSS`, appended to `DASHBOARD_CSS`). |
| 6 | Picker/matching robustness | ✅ dropdown excludes `error` runs, capped `[:100]`; hero shows `N matched · X only A · Y only B` + `<50%` low-overlap warning; positional-dup matching noted in docstring. |

New tests cover every fix: `test_compare_self_compare_is_400`, `test_compare_absent_scorer_renders_na_not_zero`, `test_compare_corrupt_report_is_422`, `test_compare_transcript_invalid_index_is_400`, `test_compare_bar_defaults_to_two_different_runs`.

**One residual (minor, non-blocking) on #4:** the 422 path fires when a file parses as JSON but fails schema validation (that's what the test exercises). A *syntactically* corrupt file still routes to 404, because `library.load_surface` swallows `JSONDecodeError` and returns `(None, {})` before `_resolve_run` reaches its `try`. Rarer case, still no crash — just labeled "not found" instead of "corrupt". Fix belongs in `load_surface`, not this PR. Fine to defer.

Good to merge. 👍

KarinaKKarinaK:
### How the compare surface looks now (RES-1085)

Ran the branch locally against two sample sim runs (`baseline · gpt-4o-mini` vs `candidate · gpt-4o`) and captured the new views.

**1. Entry point — "Compare runs" bar on the Agent Sim overview** (`/?surface=sim`)
Run A / Run B dropdowns + Compare button, sitting above the recent-runs list.



**2. New compare page** (`/compare/sim?a=…&b=…`)
- KPI delta cards: goal-achieved rate, mean score, mean turns, conversations (each with delta vs A)
- Inline-SVG charts: Outcomes (A vs B), Scorer averages (shared scorers), How conversations ended, and a per-conversation score-change diverging chart
- All-metrics table (aggregates + per-scorer + terminated-by, each A / B / delta)
- Outcome diffs matched by `(persona, scenario)` with goal flips highlighted, plus "Only in A / Only in B" for unmatched conversations



All charts are server-rendered SVG, so they show up with no JS and no `/static` dependency.

KarinaKKarinaK:
Preview (with temporary example data just for illustration purposes):
Screenshot 2026-07-16 at 5 20 49 PM
Screenshot 2026-07-16 at 5 21 12 PM
Screenshot 2026-07-16 at 5 21 25 PM
Screenshot 2026-07-16 at 5 21 35 PM

Baukebrenninkmeijer:
Did you rebase this onto main? Cause these visuals look like the old version.

Baukebrenninkmeijer:
In general, i do like the data and the visuals. Just have to align them with our current style. And how does the ux/ui look like to setup a comparison? Do you select two runs in the run overview? Or on a run, you have a button that allows you to compare with another run?

KarinaKKarinaK:
Screenshot 2026-07-31 at 9 59 03 AM
Screenshot 2026-07-31 at 9 59 13 AM
Screenshot 2026-07-31 at 9 59 23 AM
Screenshot 2026-07-31 at 9 59 34 AM
Screenshot 2026-07-31 at 9 59 42 AM

currentlycodinng (approved):
Approved, looks clean.

One thing I thought about is when a metric is zero for both runs, like goal-achieved rate here, the bar just renders empty, so at a glance you can't tell "both scored zero" from "not measured" without going into the data. Should we add a small empty-state or a literal 0 on the bar? Not blocking, just something I noticed that confused me at first.

Screenshot 2026-08-03 at 08 52 57

KarinaKKarinaK:
Screenshot 2026-08-03 at 9 51 26 AM
added now :))

Baukebrenninkmeijer (approved):
All earlier review points are addressed — rebased on the updated dashboard, vega helpers reused instead of the hand-rolled SVG, error/status contract (404 vs 422) fixed, strict-load corrupt-file case closed (3eaf197), and zero-value bars now labeled (e66547f). CI green, one prior approval. Approving to clear my stale change-request.

PR #69, opened 2026-07-13
feat(redteam)!: unify AttackerResponse onto AgentResponse (RES-883)

KarinaKKarinaK:
Unifies the redteam attacker output onto the shared `AgentResponse` type (RES-883). Follow-up from RES-877 (RES-808 PR4), which added `AgentResponse` + the per-response `AgentResponseError`. The attacker was the odd one out with its own `AttackerResponse` type; this makes one response shape across attacker, target, and simulation agents.

## What changed
- `Turn.attacker` is now `AgentResponse` (was `AttackerResponse`).
  - `generated_prompt` → `AgentResponse.text`.
  - `truncated` is dropped — it was always `finish_reason == 'length'` and is derivable (it was never read anywhere; `truncated_turns` is computed independently in the orchestrator).
  - The attacker gains the shared per-response `error` field for free.
- `AttackerResponse` is kept as a deprecated alias (`AttackerResponse = AgentResponse`), matching the existing `SendResult = AgentResponse` precedent, so imports keep working.
- Construction sites updated: `orchestrator.py` (turn record + marker-stripped rebuild), `pipeline.py`. Read sites updated: `turns_to_messages`, `attacker_input_at`, `openresponses/dataset.py`, `reports/converters.py`.

## Back-compat (breaking schema, mitigated)
The serialized report shape changes (attacker uses `output` items instead of `generated_prompt`/`truncated`). A `@model_validator(mode='before')` on `Turn` migrates pre-RES-883 report JSON (`generated_prompt` → `text`, drops `truncated`), and `converters.py` falls back to `generated_prompt` when reading old serialized turns — so **old reports still load**. This mirrors the existing `_migrate_scope_to_vulnerability_domain` pattern in the same module.

Marked `feat!:` / BREAKING CHANGE for the tag-driven semver.

## Tests
- New `TestAttackerResponseUnifiedOntoAgentResponse` in `test_contracts.py`: attacker is an `AgentResponse` with the `error` field, `truncated` is derivable from `finish_reason`, and pre-RES-883 report JSON migrates.
- All prior `AttackerResponse(generated_prompt=...)` test constructions migrated to `AgentResponse(text=...)`.
- Full non-integration suite: 2683 passed. `ruff check src` clean, `basedpyright` 0 errors.

## Design decision
Chose **full unification with a deprecated alias** over subclassing/wrapping: `Turn.attacker` is genuinely an `AgentResponse` (the ticket's goal — one response shape), and the alias keeps import back-compat without a second model. `.text` is the right mapping for `generated_prompt` — all three read sites already turned it into a user message.

### DIFF
--- a/src/evaluatorq/openresponses/dataset.py
+++ b/src/evaluatorq/openresponses/dataset.py
@@ -141,7 +141,7 @@ def turns_to_openresponses_input(
     """Convert redteam turns to an OpenResponses input array."""
     out: list[dict[str, Any]] = []
     for idx, turn in enumerate(turns):
-        out.append(user_input_item(turn.attacker.generated_prompt))
+        out.append(user_input_item(turn.attacker.text))
         if idx == len(turns) - 1 and not include_final_assistant:
             continue
         append_assistant_turn(out, turn.target)
--- a/src/evaluatorq/redteam/adaptive/orchestrator.py
+++ b/src/evaluatorq/redteam/adaptive/orchestrator.py
@@ -869,12 +869,12 @@ class MultiTurnOrchestrator:
                         # The rationale is kept OUT of the transcript on purpose: it is a
                         # signal, not evidence, and must never reach the scorer/judge.
                         attack_prompt = marker_prompt
-                        # Rebuild attacker record with the stripped prompt as its text,
-                        # preserving usage/finish_reason (RES-883: attacker is an AgentResponse).
-                        current_attacker = AgentResponse(
-                            text=attack_prompt,
-                            usage=current_attacker.usage,
-                            finish_reason=current_attacker.finish_reason,
+                        # Rebuild the attacker record with the stripped prompt as its
+                        # only text item, preserving every other field (usage,
+                        # finish_reason, model, response_id, error) via model_copy
+                        # (RES-883: attacker is an AgentResponse).
+                        current_attacker = current_attacker.model_copy(
+                            update={'output': [TextOutputItem(text=attack_prompt, annotations=[])]}
                         )
                         if not attack_prompt:
                             # Nothing left to send. Distinguish a genuine accepted
--- a/src/evaluatorq/redteam/adaptive/orchestrator.py
+++ b/src/evaluatorq/redteam/adaptive/orchestrator.py
@@ -36,7 +36,6 @@ from evaluatorq.redteam.contracts import (
     DEFAULT_PIPELINE_MODEL,
     PIPELINE_CONFIG,
     AgentContext,
-    AttackerResponse,
     AttackStrategy,
     LLMConfig,
     OrchestratorResult,
@@ -786,12 +785,9 @@ class MultiTurnOrchestrator:
                     # Build the canonical attacker record for this turn (used whether
                     # the target then succeeds or fails).
                     fr = finish_reason if isinstance(finish_reason, str) else None
-                    current_attacker = AttackerResponse(
-                        generated_prompt=attack_prompt,
-                        usage=usage,
-                        truncated=fr == 'length',
-                        finish_reason=fr,
-                    )
+                    # Attacker output is an AgentResponse (RES-883): the prompt is the
+                    # response text; truncation is derivable from finish_reason == 'length'.
+                    current_attacker = AgentResponse(text=attack_prompt, usage=usage, finish_reason=fr)
 
                     # Abort if adversarial LLM returned empty content
                     if not attack_prompt.strip():
@@ -873,8 +869,13 @@ class MultiTurnOrchestrator:
                         # The rationale is kept OUT of the transcript on purpose: it is a
                         # signal, not evidence, and must never reach the scorer/judge.
                         attack_prompt = marker_prompt
-                        # Rebuild attacker record with stripped prompt
-                        current_attacker = current_attacker.model_copy(update={'generated_prompt': attack_prompt})
+                        # Rebuild attacker record with the stripped prompt as its text,
+                        # preserving usage/finish_reason (RES-883: attacker is an AgentResponse).
+                        current_attacker = AgentResponse(
+                            text=attack_prompt,
+                            usage=current_attacker.usage,
+                            finish_reason=current_attacker.finish_reason,
+                        )
                         if not attack_prompt:
                             # Nothing left to send. Distinguish a genuine accepted
                             # success from a rejected turn-0 marker (objective_achieved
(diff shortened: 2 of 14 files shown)


Baukebrenninkmeijer (commented):
## Review — RES-883 unify `AttackerResponse` onto `AgentResponse`

Automated multi-agent review (7 reviewers). Core refactor is correct and tested — `src/` `.generated_prompt`→`.text` swaps are consistent and `ruff check src` is green; truncation info isn't lost (still aggregated into `OrchestratorResult.truncated_turns` from `finish_reason`). The items below are the ones worth addressing before merge.

### Important

**1. `_attacker_text` diverges from `AgentResponse.text` (latent correctness + duplication)** — `src/evaluatorq/redteam/reports/converters.py:111-117`
The helper returns the **first** `output_text` item and stops, whereas the canonical `AgentResponse.text` property (`src/evaluatorq/contracts.py:721`) **concatenates all** `TextOutputItem`s. For a multi-segment attacker output the report transcript silently disagrees with what `turns_to_messages`/the judge actually saw. Latent today (orchestrator emits a single text item), but it's a divergence waiting to bite. Reuse `.text` semantics (join all items) rather than first-hit.

**2. Migration is implemented twice and can drift** — `contracts.py:875` (`Turn._migrate_attacker_generated_prompt`) and `converters.py:111` (`_attacker_text`)
The pre-RES-883 wire format is migrated in two independent places: the `Turn` validator, and a raw-dict path in `_coerce_job_output_payload` that bypasses `Turn` validation entirely. Their semantics already differ (see #1). Collapse to one shared `_dict_output_text(resp) -> str` helper (which would also dedupe the near-identical target-text loop three lines below at `converters.py:124-128`).

**3. The `AttackerResponse` alias is a footgun, not a compat bridge** — `contracts.py:859`
`AttackerResponse = AgentResponse` is a plain rebind: no `DeprecationWarning`, and because `AgentResponse` ignores extra kwargs, old-style `AttackerResponse(generated_prompt='...')` **silently returns an empty-output response** (prompt dropped) instead of raising. It also makes `isinstance(x, AttackerResponse)` identical to `isinstance(x, AgentResponse)`, so any external attacker/target discrimination silently breaks. This is a `feat!` breaking release — either delete the name (clean break) or make it a real `module __getattr__` that emits `DeprecationWarning`. As written it's untested dead weight (nothing in-repo imports it).

**4. Duplicate imports from the mechanical rename** — `tests/unit/test_tool_call_interception.py:13-14` and `tests/redteam/test_dynamic_job.py:132-133`
Both now import `AgentResponse` twice (in `test_dynamic_job` from two different modules — the first line is dead/shadowed). Ruff flags these but `tests/` is outside the `ruff check src` gate, so CI is green by scope, not correctness. Trivial cleanup.

**5. Test gaps for the load-bearing new code**
- `_attacker_text` has no direct test — neither the `output_text` path nor the `generated_prompt` fallback is exercised (the `_coerce_job_output_payload` tests pass `turns` as an int, so the loop never runs).
- No legacy-report round-trip: nothing feeds a full pre-RES-883 `job_output` through `_coerce_job_output_payload` to prove old reports still render. That's the exact scenario the fallback exists for.

### Minor / non-blocking

- **`orchestrator.py:872-877`** — the rationale-stripping rebuild reconstructs `AgentResponse(text=..., usage=..., finish_reason=...)`, silently dropping any other fields (`model`, `response_id`, `error`). Harmless today; `model_copy(update={'output': [...]})` would preserve everything and match the original intent.
- **Design note:** the attacker now carries `tool_calls`, `response_id`, `from_openresponses`, multi-item `output`, and `error` — none of which the attacker path populates (the test even asserts `attacker.error is None`). Full collapse onto the target type captures the real overlap (`text`/`usage`/`finish_reason`/`error`) but over-generalizes. A narrow shared base was an option; unifying is a reasonable call, just don't sell `error` as a gained feature — it's structurally always `None` here.

KarinaKKarinaK:
Rebased on latest main and addressed the review; **CI is now green** (the Typecheck+test jobs were failing on `basedpyright`'s whole-project `reportDuplicateImport` — see finding 4).

**1. `_attacker_text` first-hit vs canonical `.text`** — removed the helper. `_flatten_turns` now validates each turn through `Turn.model_validate` and reads `AgentResponse.text`, which concatenates **all** text items (matches `turns_to_messages`/the judge). Added a test with a multi-segment attacker output asserting concatenation.

**2. Migration implemented twice** — collapsed to one authority: the `Turn` validator. Report coercion no longer has its own raw-dict migration path; routing through `Turn.model_validate` also dedupes the near-identical target-text loop.

**3. `AttackerResponse` alias footgun** — removed outright (clean `feat!` break) instead of a silent rebind. `from ... import AttackerResponse` now fails loudly rather than `AttackerResponse(generated_prompt=...)` quietly dropping the prompt. Nothing in-repo imported it.

**4. Duplicate imports from the rename** — fixed all five test files (the two you named plus `test_multi_turn_state`, `test_contracts`, `test_turns_to_messages`). `uv run basedpyright` (whole project) is now `0 errors, 0 warnings` — this is what was breaking CI.

**5. Test gaps** — added to `test_converters.py`: a multi-segment concatenation test and a full pre-RES-883 report round-trip through `_coerce_job_output_payload` (the exact fallback scenario).

**Minor** — the orchestrator rationale-strip rebuild now uses `model_copy(update={'output': [...]})`, preserving `model`/`response_id`/`error` instead of dropping them. And you're right that `error` is structurally always `None` on the attacker path today — I've dropped the "gained feature" framing from the code comment; it's just part of the unified shape.

Verified: full non-integration suite 2966 passed, `ruff check src` clean, whole-project `basedpyright` clean, all CI checks green.

Baukebrenninkmeijer (commented):
## Re-review — commit 279ea19

All 5 important items from the prior review resolved. Verified locally (affected suite passes, `ruff check src` + basedpyright clean).

| # | Item | Resolution |
|---|------|-----------|
| 1 | `_attacker_text` first-item vs `.text` all-items divergence | Fixed — coercion routes each turn through `Turn.model_validate` and reads canonical `AgentResponse.text` (joins all items). |
| 2 | Two divergent migration paths | Fixed — single `Turn` migrator is now the only authority; target-text loop deduped into the same path. |
| 3 | `AttackerResponse` alias footgun | Fixed — removed outright (clean `feat!` break), no silent rebind. |
| 4 | Duplicate imports | Fixed in both test files. |
| 5 | Test gaps | Fixed — `test_new_shape_turns_flatten_to_conversation` (multi-segment concat) + `test_pre_res883_report_round_trips`. |

Bonus: orchestrator rationale-strip rebuild now uses `model_copy(update={'output': ...})`, preserving `model`/`response_id`/`error` instead of dropping them.

One intentional behavior change to flag (not a blocker): converter target text now joins all `output_text` items via `.text`, where the old loop took last-item-wins. This is more consistent with what `turns_to_messages`/the judge actually see. Fine.

LGTM.

Baukebrenninkmeijer (approved):
Re-reviewed 279ea19: all prior items resolved, CI green. Approving.

PR #70, opened 2026-07-13
feat(dashboard): enrich Orq agent simulation reports

Baukebrenninkmeijer:
## Summary
- add a compact Orq agent card to simulation reports, with safe description previews, tool metadata, and fallbacks for deleted or unavailable agents
- defer live agent-detail enrichment to an HTMX fragment so report rendering stays non-blocking
- render simulation target kinds directly from saved JSON, including an external-entity icon for callback targets
- build valid Studio deep links from `ORQ_WORKSPACE` rather than UUID workspace IDs, repairing legacy agent links at render time
- refine simulation persona and scenario details into compact rows with accessible modal drill-downs

## Validation
- `uv run pytest tests/dashboard/test_orq_links.py tests/simulation/test_agent_info.py tests/dashboard/test_report_tabs.py tests/dashboard/test_sim_overview.py -q`
- `uv run ruff check src/evaluatorq/dashboard/orq_links.py src/evaluatorq/simulation/utils/run_store.py src/evaluatorq/dashboard/report_tabs.py`
- `git diff --check`

## Notes
- `ORQ_WORKSPACE` must contain the Studio workspace key used in URLs, for example `research`; entity API UUIDs are intentionally never used as Studio route slugs.


Baukebrenninkmeijer:
## Review summary — automated pass (dashboard/sim/redteam alignment)

**Verdict: no merge blockers.** No critical or important bugs. The filter rewrite, async agent-card enrichment, deep-link builder, and reasoning-effort caching all hold up under inspection. Findings below are hygiene/robustness, each verified against source.

### Prioritized

**1. Live agent-info fetch failure logged at DEBUG → invisible** — `dashboard/report_tabs.py:672`
`_orq_agent_info_cached` catches all exceptions at `logger.debug`; at default INFO/WARNING a persistently failing enrichment path is silent to operators. The outer catch is also largely dead — `fetch_agent_info` already swallows+warns internally, so it only ever catches the local `import`. Suggest bumping to `warning` and collapsing the double-swallow.

**2. `reasoning_effort` memo keyed on model name only** — `common/llm_call.py:38,88`
The 400 only occurs when function tools are present, but once a model is in `_REASONING_EFFORT_REJECTORS`, `reasoning_effort` is stripped for *every* later call to that model — including tool-free calls that would accept it. Silent reasoning-quality downgrade (deliberate per the `ponytail:` comment). Consider keying on `(model, has_tools)` if it matters.

**3. Test isolation gap** — `tests/unit/test_execute_chat_parse.py`
Missing the `_reset_reasoning_rejectors` autouse fixture that `test_llm_call.py:16` has; both mutate the process-global `_REASONING_EFFORT_REJECTORS` via model `"m"`. Both files run green together today (15 passed), so it's order-fragile, not an active failure. The parse path's memoization (add-to-set + strip-upfront) is also untested, unlike the completion path. Add the reset fixture + one memoization assertion.

**4. `agent_info: dict[str, Any]` is an untyped ad-hoc struct** — `simulation/types.py`, `run_store.py:80-94`
Fixed 10-key shape accessed via `.get()` at 4+ sites with 3 defensive `isinstance(..., dict)` guards; a key typo silently returns `None` and basedpyright can't catch it. Acceptable for a best-effort snapshot, but a typed `AgentInfoSnapshot(BaseModel)` would remove the guards. Related: `AgentContext.target_kind` is `str | None` with a docstring claiming a closed set — could tighten to `Literal[...]` to match `SimulationRun.target_kind`.

### Minor / nits
- agent-card route (`app.py:263`) returns generic `"Error loading report"` + status **422** for a server-side load failure; sibling routes include `exc`/`path.name` and this is a 500-class condition.
- Provenance mislabel (`report_tabs.py:_resolve_agent_info`): an incomplete captured snapshot whose live enrichment fails renders as source `captured` (authoritative), suppressing any "enrichment unavailable" note.
- Filter-rail UX (`view.py:762` vs `769`): Outcome chips render *unchecked* for "empty = all" while Severity chips render *checked* for the same semantic.
- Coverage gaps (moderate): `fetch_agent_info` defensive mapping branches (str-vs-object KBs, dict sub_agents, None-drops), workspace-key/URL composition, and `_simulate_core` agent-info wiring are untested.
- Comment disambiguation (trivial): `spec §Shared infrastructure #1` (styles.py) doesn't name which spec; `parity: dashboard.py:385` (sim_views.py) is ambiguous now that two `dashboard.py` files exist.

### Strengths
Comments are unusually accurate for a PR this size (~40 spec/parity citations spot-checked, none fabricated). `test_report_tabs.py` is a model of behavioral testing (five provenance states, per-field merge invariant, deferred-HTMX contract). Security negatives are asserted (`system_prompt`/`instructions` excluded from snapshots; unrelated-400 not masked). All HTML builders route user content through `esc()`.

<sub>🤖 Automated review (5 specialized passes). Findings verified against source before posting.</sub>

PR #91, opened 2026-07-31
feat(simulation): accept Orq experiments as simulation input (RES-1083)

KarinaKKarinaK:
## Summary

Adds Orq experiments as a simulation input source, filling the Experiment row of the data-sources matrix in `simulation/README.md`. Two modes:

- **Direct**: `simulate(experiment_id=..., experiment_run_id=...)` replays a prior experiment run's rows as datapoints. Reuses the core `fetch_experiment_datapoints` fetcher from `fetch_data.py` and the same shape-tolerant row extractor as the `dataset_id` path, so experiments uploaded by a previous simulation run round-trip as-is. `experiment_run_id` picks a specific run; latest when omitted. Also on the CLI: `eq sim simulate --experiment-id / --experiment-run-id`.
- **Extension**: `extend_from_experiment()` fetches the run's rows, feeds their (deduplicated) personas and scenarios to the standard `DatapointGenerator` as seed context, and returns new similar-but-not-duplicate datapoints.

`experiment_id` joins the existing mutual-exclusivity check (`dataset_id` / `datapoints` / `personas`+`scenarios`).

## Testing

- 12 new tests in `tests/simulation/test_experiment_input.py`: direct extraction (both row shapes), per-row error indexing, zero-row and missing-key errors, resolver wiring + exclusivity, extension seeding/dedup, CLI flag validation.
- Full non-integration suite: 3246 passed. `ruff check src` and `basedpyright` clean.

Linear: RES-1083


currentlycodinng (approved):
**Approve.** Clean addition that's built on already-working pieces — it reuses `fetch_experiment_datapoints` (the `evaluate()` fetcher) and the dataset path's `_extract_single_datapoint`, so the direct mode is mostly wiring plus the row-index-in-error trick (which is correct: `extend` appends each yielded item before pulling the next, so `len(out)` is the failing index).

Verified the round-trip live, not just in mocks: ran a real simulation, let it upload an experiment, then replayed that experiment through `datapoints_from_experiment` — the persona and scenario come back intact, so "experiments uploaded by a previous simulation run round-trip as-is" actually holds against the platform. 12/12 unit tests pass and the exclusivity/CLI wiring is covered.

One non-blocking observation from the live run: the platform's experiment export isn't ready the instant a run finishes — the `.../export` call only returns its 302 signed URL a few seconds later, and `fetch_experiment_datapoints` has no retry, so calling `datapoints_from_experiment` immediately after a run raises "did not return a download URL". Replaying a prior run (the actual use case) is unaffected. Might be worth a short poll/retry in the fetcher eventually, but it's pre-existing and out of scope here.

arianpasquali (approved):
Two things, neither blocks merge.
confirm this. 

1. Replaying an experiment can break if the API returns a string instead of an object

In _extract_single_datapoint (_datapoint_io.py), the persona and scenario fields get converted from string to object when needed, because the Orq API sometimes returns nested objects as JSON strings. The datapoint field doesn’t get this conversion. So if the API ever returns datapoint as a string, replaying that experiment fails with an error.

Fix is small: apply the same conversion to datapoint that persona/scenario already have. Or add one integration test with a real export to prove it never happens.

2. The new retry loop also affects evaluate(), and may create duplicate exports

The retry in fetch_data.py looks correct. But two things to double-check:

When an export never becomes ready, we now wait about 8 seconds and call the API 5 times before giving up. Before, we failed immediately. This also applies to the existing evaluate() flow, not just simulations.
Each retry POSTs to /export again. Does each POST start a new export job on the server? If yes, one failed fetch could queue up 5 jobs.

Minor, for a future cleanup ticket (pattern is older than this PR): the API-key check exists in 3 places, and the “exactly one data source” check in 2. Would be nice to have each in one place.

KarinaKKarinaK:
Addressed both points and resolved the merge conflicts with main (`previous_run` replay + `experiment_id` now coexist as input sources; the exclusivity check and CLI flag validation cover all five sources).

1. **Stringified `datapoint` fields**: `_extract_single_datapoint` now applies the same `_as_obj` coercion to `datapoint` / `datapoints` (container and element) that persona/scenario already had. Covered by three new tests in `test_datapoint_io.py`.

2. **Retry loop / duplicate exports — confirmed against orquesta-web**: the `/manifests/{run_id}/export` POST does **not** enqueue a job per call. `SheetsManager.exportManifest` either 302-redirects to a cached presigned URL, or builds the export inline and returns it directly as the response body (`Content-Disposition: attachment`), caching it in the background. (The job-dispatching route is the separate private `/:id/export` used by the UI.) That also explains the "URL not ready" window: the first POST after a run returns the raw JSONL body, which the fetcher was ignoring while looking for a URL. The fetcher now consumes that inline body directly, so the first call succeeds with no retry and no extra wait — the retry loop remains only as a rare fallback. This applies to `evaluate()` too, which now gets faster on fresh runs rather than slower.

The API-key / exactly-one-source dedup is noted for a follow-up ticket.

Full suite: 3437 passed, ruff and basedpyright clean.

ChielToto:
What worked

All the correctness critical paths check out. The keyword only * in _resolve_or_generate_datapoints makes the mixed defaults valid. DatapointGenerator.__init__, generate_from_description, and close all match the extension call. fetch_experiment_datapoints returns DataPoint objects via _experiment_row_to_datapoint, so the .inputs access holds. The list.extend(generator) row index trick is correct (len(out) equals the failing index). Wiring through _config, api, and cli is consistent and covered by the 12 new tests. Nothing severe.

What didn't

1. fetch_data.py export retry loop: the inline JSONL body is recognized only via the exact Content-Disposition: attachment header. If the server returns a success response carrying the body without that header (or with a differently cased value), the code falls into the .json() branch, fails to parse JSONL, suppresses the error, treats the attempt as not ready, retries 5 times, then raises "did not return a download URL", discarding data that was already in the body. orquesta web is controlled and emits lowercase attachment, so this is defensive, but a content type fallback (x ndjson) would make it robust.
2. experiments.py (~line 57): datapoints_from_experiment reuses _extract_single_datapoint, which is built for the single datapoint wrap_simulation_agent contract. A row carrying a multi element datapoints/personas/scenarios list yields experiment 'ex_1' row 3: wrap_simulation_agent DataPoint must encode exactly one datapoint..., referencing a function unrelated to what the user did. Low impact since simulation uploaded rows use singular datapoint, but confusing if it fires.
3. experiments.py _describe_agent: if every seed scenario has an empty goal, it returns the bare "An agent whose users pursue goals such as: ". Seeds are guaranteed non empty but individual goals are not, so the generator can get a truncated description. A generic fallback would avoid that.

None block merge.

KarinaKKarinaK:
Applied all three in 08da81cd:

1. **Export body detection**: the inline-body branch now also accepts an ndjson/jsonl content type, and both header values are checked case-insensitively, so a success body without the exact attachment disposition is consumed instead of being retried 5 times and discarded. Regression test added (inline body, no Content-Disposition at all).
2. **Row error message**: `_extract_single_datapoint` gained a `source` label; the experiment and dataset call sites pass `source='row'`, so a multi-element row now fails as `experiment 'ex_1' row 3: row must encode exactly one datapoint, got 4. Use one row per conversation.` with no mention of wrap_simulation_agent. The wrap path keeps its own wording. Test asserts the message and the absence of the wrap reference.
3. **`_describe_agent`**: falls back to a generic description when every seed goal is blank, instead of the truncated `goals such as: ` prompt. Test added.

Full suite 3440 passed, ruff + format + basedpyright clean.

Baukebrenninkmeijer:
Found two remaining non-blocking issues in the final diff:

1. **Extension mode does not enforce the advertised non-duplication.** `extend_from_experiment()` returns `generator.generate_from_description(...)` unchanged (`src/evaluatorq/simulation/experiments.py:110-115`). `_seed_context()` deduplicates the *prompt context*, but the generated personas/scenarios are never compared against the seed set or filtered/retried. A model can therefore return an identical seed while the public docstring/README promise “new” and “not duplicate” datapoints. The current test only checks that the prompt contains the instruction, not that duplicate output is rejected. Either enforce this in code/tests or narrow the claim.

2. **The CLI help text is stale.** `eq sim simulate --help` advertises the input sources as only `--input`, `--dataset-id`, and `--from-run`; it omits the supported `--experiment-id` (and `--experiment-run-id`) even though the options and runtime validation exist. Please update the command docstring so the primary help output documents the new input path.

Baukebrenninkmeijer (approved):
Reviewed and verified. The non-programmatic duplicate prevention is an accepted tradeoff.

PR #92, opened 2026-07-31
feat: support a previous run as an input source for sim + red teaming

currentlycodinng:
Closes [RES-1126](https://linear.app/orqai/issue/RES-1126/support-previous-run-as-an-input-source-for-sim-red-teaming-replay).

## What

A first-class "replay this run" input on both surfaces. Passing it short-circuits every other data path — no dataset load, no persona or strategy generation, no attack generation. The stored cases run again against whatever target and evaluators the new invocation names, so version-to-version regression on an identical case bank is one command.

```bash
eq redteam run --target agent:my-agent-v2 --from-run latest
eq sim simulate --from-run latest --target agent:my-agent-v2
```

```python
await red_team(target='agent:my-agent-v2', previous_run='latest')
await simulate(target='agent:my-agent-v2', previous_run='latest')
```

Run references accept what the runs listings already print: `latest`, a run file name, a run name, a run id (or an unambiguous 8+ character prefix), or a path to a saved run JSON.

## How

Replay needs the cases, and neither surface stored them on the default path: red team only wrote datapoints under `save='detail'`, and sim results keep persona/scenario *names*, not the objects. So the saved records now carry what they ran.

- `common/replay.py` — shared reference resolver (`latest` → path → runs-dir name → manifest run id → run-name prefix), with an error that names the runs dir and lists recent runs.
- `redteam/replay.py` — rebuilds `DataPoint`s from a saved run and restores its pipeline mode; `_auto_save_run` now persists the executed datapoint inputs next to the report (not on the report model, so nothing changes for uploads).
- `simulation/replay.py` — `SimulationRun.datapoints` is populated on save and read back on replay.
- Both runners take the replayed set as pre-built datapoints for *every* target, skipping capability classification, planning, estimation, and the dataset load.

`previous_run` is mutually exclusive with the data-selection arguments it would otherwise silently override, and raises naming the conflicts. Runs saved before this carry no datapoints and are rejected with an explanatory error rather than a partial replay.

Also fixes a latent double-count in hybrid runs that mix string targets with `AgentTarget` objects, where the static leg was appended twice.

## Testing

`uv run pytest -m 'not integration'` — 3272 passed (38 new). `ruff check`, `ruff format --check`, and `basedpyright` clean on `src`; `mkdocs build --strict` and the mermaid validator pass.

New coverage: reference resolution (incl. ambiguous prefixes and unreadable files), replay rebuilding for both surfaces, legacy runs without datapoints, mutual-exclusion errors, CLI wiring for both flags, and an e2e pair asserting a replay re-runs the same attack ids against a new target while making zero generation or classification calls.

Data-source matrices in both surface READMEs move "previous runs" from ⚠️ manual to ✅ built-in, with a replay section each.

### DIFF
--- a/src/evaluatorq/common/replay.py
+++ b/src/evaluatorq/common/replay.py
@@ -38,8 +38,37 @@ MIN_RUN_ID_PREFIX = 8
 _SUGGEST_LIMIT = 5
 
 
+REPLAY_VERSION_KEY = 'replay_version'
+"""Key under which a saved run stamps the format its replay payload is written in."""
+
+REPLAY_VERSION = 1
+"""Format version this build writes and is able to read.
+
+Without a marker, recognising an unreplayable run means sniffing for an absent
+``datapoints`` key — a trick that only works for the one transition from
+pre-replay runs. Stamping the version means the *next* format change can say
+"saved by a newer version of evaluatorq" instead of failing structurally
+somewhere downstream. Runs with no marker are pre-versioning and are read as
+version 1, which is what they are.
+"""
+
+
 class ReplayError(RuntimeError):
     """A previous-run reference could not be resolved or replayed."""
+
+
+def check_replay_version(payload: dict[str, Any], path: Path, *, surface: str) -> None:
+    """Reject a run whose replay payload is newer than this build understands.
+
+    Raises:
+        ReplayError: The run declares a format version above :data:`REPLAY_VERSION`.
+    """
+    stored = payload.get(REPLAY_VERSION_KEY)
+    if isinstance(stored, int) and stored > REPLAY_VERSION:
+        raise ReplayError(
+            f'{path.name} was saved by a newer version of evaluatorq (replay format v{stored}; '
+            f'this build reads v{REPLAY_VERSION}). Upgrade evaluatorq to replay this {surface} run.'
+        )
 
 
 def _reports_newest_first(runs_dir: Path) -> list[Path]:
@@ -151,4 +180,5 @@ def load_run_payload(reference: str, runs_dir: Path, *, surface: str) -> tuple[d
         raise ReplayError(f'Previous {surface} run {path} is not valid JSON: {exc}') from exc
     if not isinstance(payload, dict):
         raise ReplayError(f'Previous {surface} run {path} does not contain a run object.')
+    check_replay_version(payload, path, surface=surface)
     return payload, path
--- a/src/evaluatorq/common/replay.py
+++ b/src/evaluatorq/common/replay.py
@@ -1,0 +1,138 @@
+"""Resolve a "previous run" reference to the JSON payload of a persisted run.
+
+Both surfaces persist their runs under ``.evaluatorq/`` (``runs/`` for red
+teaming, ``sim-runs/`` for agent simulation) and both want the same thing: take
+a handle a user can actually type — the filename shown by ``eq redteam runs`` /
+``eq sim runs``, a run id from a manifest, or just ``latest`` — and turn it into
+the stored record so its cases can be replayed.
+
+Resolution order for a reference, first match wins:
+
+1. ``latest`` — the most recently modified report in the runs dir.
+2. An existing file path (absolute or relative to the cwd).
+3. ``<runs_dir>/<reference>`` and ``<runs_dir>/<reference>.json``.
+4. A manifest ``run_id`` (exact, or an unambiguous prefix of at least 8 chars)
+   whose manifest records a ``report_path``.
+5. A report whose filename stem equals the reference, or whose stem starts with
+   ``<reference>_`` (the ``<run-name>_<timestamp>`` convention) — newest wins.
+
+Anything else raises :class:`ReplayError` naming the runs dir and the most
+recent runs, so a typo is one line away from being fixed.
+"""
+
+from __future__ import annotations
+
+import json
+import operator
+from pathlib import Path
+from typing import Any
+
+MIN_RUN_ID_PREFIX = 8
+"""Shortest run-id prefix accepted for resolution (uuid4 hex is 32 chars)."""
+
+_SUGGEST_LIMIT = 5
+
+
+class ReplayError(RuntimeError):
+    """A previous-run reference could not be resolved or replayed."""
+
+
+def _reports_newest_first(runs_dir: Path) -> list[Path]:
+    if not runs_dir.is_dir():
+        return []
+    reports: list[tuple[float, Path]] = []
+    for p in runs_dir.glob('*.json'):
+        try:
+            reports.append((p.stat().st_mtime, p))
+        except OSError:  # noqa: PERF203 — best-effort listing, a vanished file is skipped
+            continue
+    reports.sort(key=operator.itemgetter(0), reverse=True)
+    return [p for _, p in reports]
+
+
+def _resolve_via_manifest(reference: str, runs_dir: Path) -> Path | None:
+    """Match *reference* against manifest run ids (exact, then unique prefix)."""
+    from evaluatorq.common.run_manifest import list_manifests
+
+    manifests = [(m.run_id, m.report_path) for m in list_manifests(runs_dir) if m.report_path]
+    for run_id, report_path in manifests:
+        if run_id == reference:
+            return Path(report_path)
+    if len(reference) < MIN_RUN_ID_PREFIX:
+        return None
+    matches = {report_path for run_id, report_path in manifests if run_id.startswith(reference)}
+    if len(matches) == 1:
+        return Path(next(iter(matches)))
+    if len(matches) > 1:
+        raise ReplayError(
+            f'Run reference {reference!r} is ambiguous — it matches {len(matches)} runs. Use a longer id.'
+        )
+    return None
+
+
+def _resolve_via_name(reference: str, runs_dir: Path) -> Path | None:
+    """Match *reference* against report filename stems, newest first."""
+    for p in _reports_newest_first(runs_dir):
+        if p.stem == reference or p.stem.startswith(f'{reference}_'):
+            return p
+    return None
+
+
+def _not_found(reference: str, runs_dir: Path, surface: str) -> ReplayError:
+    recent = [p.name for p in _reports_newest_first(runs_dir)[:_SUGGEST_LIMIT]]
+    hint = f' Recent runs: {", ".join(recent)}.' if recent else f' No runs found in {runs_dir}.'
+    return ReplayError(
+        f'Could not resolve previous {surface} run {reference!r} in {runs_dir}. '
+        f'Pass a run file name, a run id, a path to a saved run, or "latest".{hint}'
+    )
+
+
+def resolve_run_path(reference: str, runs_dir: Path, *, surface: str) -> Path:
+    """Return the report file a previous-run *reference* points at.
+
+    Args:
+        reference: What the user typed — ``latest``, a path, a file name, a run
+            name, or a manifest run id (full or an unambiguous 8+ char prefix).
+        runs_dir: The surface's runs directory (e.g. ``.evaluatorq/runs``).
+        surface: Human-readable surface name used in error messages.
+
+    Raises:
+        ReplayError: The reference is empty, ambiguous, or matches nothing.
+    """
+    ref = reference.strip()
+    if not ref:
+        raise ReplayError(f'Empty previous {surface} run reference.')
+
+    if ref.lower() == 'latest':
+        reports = _reports_newest_first(runs_dir)
+        if not reports:
+            raise ReplayError(f'No saved {surface} runs found in {runs_dir} — nothing to replay.')
+        return reports[0]
+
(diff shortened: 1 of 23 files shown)


currentlycodinng:
Independent review pass (two reviewers, adversarial) found real defects. Fixed in the follow-up commit:

| # | Finding | Fix |
|---|---------|-----|
| 1 | `basedpyright` (bare, as CI runs it) failed on a new test — my earlier check was scoped to `src` | Typed enum in the fixture; bare run is clean |
| 2 | **Replay wasn't verbatim**: `max_turns` / `attacker_instructions` live on the job, not the datapoints, so replaying a `--max-turns 10` run silently ran 5 turns | Saved runs record a `run_config`; replay restores it. Sim already persisted `max_turns` and now honours it. `max_turns` defaults to `None` on both surfaces so an explicit value still wins |
| 3 | A red-team replay accepted a **simulation** run file (both surfaces write `datapoints`, paths are valid references) — it failed per-row inside the orchestrator | Rejected by name with a pointer to `eq sim simulate --from-run`; rows lacking a strategy *and* messages are rejected up front |
| 4 | Nothing showed a replay was happening — the log line sits below both CLIs' default level | `Replay Of` row in the Run Plan; removed a docstring claiming this already worked |
| 5 | Errors cited "evaluatorq 1.4.0+", a version that never existed | Version-free wording |
| 6 | Same error blamed age for a `--save detail` summary report, which never carries datapoints | Names the real cause |
| 7 | Stale manifest masked a filename match in the resolver | Existence-checked before falling back |
| 8 | READMEs overclaimed (run ids aren't what the runs tables print; reproducibility scope) and dropped still-accurate notes | Corrected and restored |

**On the test I added for the hybrid double-count fix:** it passed with the fix reverted — tautological, exactly what the review warned about. That path's inflated list never reaches execution, so the bug is latent there. Replaced with a hybrid replay against a bare `AgentTarget`, where the AgentTarget *is* the first prepared target and every static row was routed to the dynamic job before the fix. Verified failing on revert.

Also added e2e coverage for static, hybrid, and multi-target replay, plus turn-budget restore and override.

**Held up under review, no change needed:** first-target `shared_datapoints` wiring, hybrid split across all four shapes, mutation aliasing, the frozen-dataclass field, JSON round-trip fidelity, `resolved_vulns=None`, backwards compat both directions, storage size (~1.3 KB/datapoint, dwarfed by transcripts), and privacy (no new class of data on disk).

**Deliberately not done:** `previous_run` still hard-rejects `categories` / `max_*_datapoints` rather than filtering the replayed set. "Re-run just the ASI01 cases" is a reasonable follow-up but it's a new feature, not part of this ticket.

3282 tests pass; `ruff`, bare `basedpyright`, strict mkdocs build and the mermaid validator all clean.

src/evaluatorq/common/replay.py
```diff
@@ -0,0 +1,141 @@
...
+        )
+    return None
+
+
+def _resolve_via_name(reference: str, runs_dir: Path) -> Path | None:
+    """Match *reference* against report filename stems, newest first."""
+    for p in _reports_newest_first(runs_dir):
+        if p.stem == reference or p.stem.startswith(f'{reference}_'):
+            return p
+    return None
```
KarinaKKarinaK: Name-prefix resolution silently returns the newest match here, but `_resolve_via_manifest` raises `ambiguous` for an id prefix that matches more than one run. Both are defensible, but a user who types a run name that prefixes several runs gets the newest with no signal, unlike the id path. It is documented as newest-wins so this is a consistency note, not a bug. If you want them to match, an ambiguity check here (or a one-line log of which run was chosen) would remove the surprise.

src/evaluatorq/redteam/runner.py
```diff
@@ -2047,8 +2173,15 @@ def _ctx_end_meta(target_label: str, ctx: AgentContext, caps: AgentCapabilities)
 
             # Build the appropriate job based on mode (hybrid vs dynamic-only)
             at_static_dps: list[Any] = []
-            if mode == Pipeline.HYBRID and static_data is not None:
-                at_static_dps = list(static_data)
+            if mode == Pipeline.HYBRID:
+                # Shared/replayed datapoints already carry both legs tagged, so
+                # split them back out rather than appending the static set twice.
+                tagged_static = [dp for dp in at_dps if dp.inputs.get('hybrid_source') == 'static']
```
KarinaKKarinaK, line 2181: The double-count fix looks correct. Since it changes hybrid behavior for mixed string + AgentTarget targets, consider a regression test that runs a non-replay hybrid with that exact target mix and asserts the static datapoints are executed once, so this cannot silently regress.

KarinaKKarinaK (approved):
Approving. This is a large change and it is carefully done. Correctness, error handling, and test coverage are all solid, and the write-up is honest about what it stores and why.

What I checked and it holds up:
- The shared resolver (`common/replay.py`) has a clear resolution order, checks the report still exists before trusting a manifest (so a stale manifest cannot mask a good name match), and every failure path names the runs dir and lists recent runs.
- Both replay rebuilders reject legacy runs with no stored cases, and each detects a run of the other kind (sim file handed to red team and vice versa) and points at the right command. Nice touch.
- Mutual exclusion is real on both surfaces, not just red team: `_resolve_datapoints` in sim raises `Pass exactly one of previous_run, dataset_id, datapoints, ...`. The docstring says precedence but the code raises, so no silent override.
- The `max_turns` default flip from 5 to None is the right call and it is safe: `resolved_max_turns` is always an int downstream, so the nullable never reaches the `int`-typed inner functions.
- The hybrid double-count fix is correct: it splits already-tagged static datapoints out of `at_dps` instead of appending `static_data` on top of them.

Suggestions, none blocking:
1. The double-count fix is a real behavior change bundled into a feature PR, but I do not see a dedicated regression test for the bug it targets: a non-replay hybrid run mixing a string target with an AgentTarget object, asserting the static leg is counted once. The replay tests exercise the new `tagged_static` branch, but not the original mixed-target scenario. Worth one test so a future refactor cannot silently bring the double-count back.
2. Minor consistency: name-based resolution silently returns the newest match on an ambiguous prefix, while id-based resolution raises `ambiguous`. It is documented as newest-wins, so intentional, but the two paths behaving differently on ambiguity could surprise someone. Inline note below.
3. Repo-norm nit: the PR body carries a `Generated with Claude Code` footer. This workspace prefers no AI attribution on artifacts.

currentlycodinng:
Thanks @KarinaKKarinaK — all three addressed in 19d7edf.

**1. Regression test for the double-count (non-replay, mixed string + AgentTarget).** You were right that it was missing, and it turned out to be the more interesting of the two. My first attempt at exactly this test passed *with the fix reverted* — the AgentTarget's inflated `all_datapoints` never reaches execution, because the run uses the first prepared target's list, and for a mixed run that's the string target's. So there was nothing observable at the report level.

It is observable one level down: `_datapoint_breakdown` is called per target with that target's own list. `test_mixed_hybrid_run_counts_the_static_leg_once` spies on it and asserts no target's list carries duplicate datapoint ids, and that all targets agree on the count. Confirmed failing on revert.

There's also `test_hybrid_replay_against_a_bare_agent_target_keeps_the_static_leg`, which covers the branch where the same bug *is* user-visible: with no string target the AgentTarget is first, so before the fix every static row was relabelled dynamic and sent to the wrong inner job. Also confirmed failing on revert.

**2. Ambiguity consistency.** Kept the asymmetry, deliberately, and documented why in the module docstring: run ids are opaque, so an ambiguous prefix has no sensible pick and raises; run *names* are expected to repeat, and `--from-run nightly` resolving to the latest nightly is the point of typing one. Took your second suggestion for the surprise — it now logs which run it chose when more than one matched. Between that and the `Replay Of` row in the Run Plan, the pick is never silent.

**3. Footer.** Removed from the PR body.

CI green.

currentlycodinng:
Code review found three more, all real, all fixed in 1a6fff3. Each was verified failing against the unfixed code before the fix went in.

**1. The simulation-side surface guard never fired.** It sat inside the empty-`datapoints` branch — but a red team run saved by this same PR carries a *non-empty* top-level `datapoints`, so it sailed straight past into `SimulationDatapoint.model_validate` and surfaced a raw pydantic dump:

```
ReplayError: ... has unreadable datapoints: 4 validation errors for SimulationDatapoint
persona: Field required ...
```

The check now runs unconditionally on the `pipeline` discriminator, mirroring the red-team side which was already correct. Both guards had no test coverage; both do now.

**2. `orq.simulation.max_turns` silently vanished from traces.** Flipping the default to `None` meant the pipeline span was built pre-resolution, and `set_span_attrs` skips `None` — so every run that didn't pass `--max-turns` lost the attribute that was previously always there. The resolved value is now stamped in `_simulate_core`, which also means a cap restored from a replayed run is reported truthfully rather than as the pre-resolution guess. Pinned by two tests (default, and replay-restored).

**3. Pipeline inference trusted the stored label over the datapoints' tags.** Confirmed the mechanism: `merge_reports` derives the label from whichever sub-reports produced rows, so a hybrid run whose static leg returns nothing is saved as `pipeline='dynamic'` while its datapoints still carry `hybrid_source='static'` — and replaying it routed `messages`-only rows into the attack-generation job. Tags now win where they contradict the label, and *only* there: an untagged run (pure dynamic, or static mode — neither tags anything) still takes the stored label, so static-mode replays are unaffected.

3291 tests pass; `ruff`, bare `basedpyright`, CI all green.

Separately, the feature is now verified live against real Orq agents. Same two ASI01 attacks, byte-identical case bank (`b['datapoints'] == r['datapoints']` → `True`), restored turn budget, replayed onto a second agent:

| attack | pizza-faq-agent | orq-skills-test-echo |
|---|---|---|
| `data_source_injection` | resisted | resisted |
| `tool_output_hijack` | resisted | **VULNERABLE** |

0% → 50% ASR on an identical case bank, which is the version-to-version regression the ticket asked for. The sim side replayed its persona, scenario, and first message with no generation step and the turn cap carried over.

src/evaluatorq/common/replay.py
```diff
@@ -0,0 +1,154 @@
...
+
+from loguru import logger
+
+MIN_RUN_ID_PREFIX = 8
+"""Shortest run-id prefix accepted for resolution (uuid4 hex is 32 chars)."""
+
+_SUGGEST_LIMIT = 5
+
+
+class ReplayError(RuntimeError):
```
KarinaKKarinaK, line 56: Non-blocking: ReplayError is not exported from evaluatorq, evaluatorq.redteam, or evaluatorq.simulation, so an SDK caller doing red_team(previous_run=...) has to import it from evaluatorq.common.replay, an internal path, to catch it. Worth adding to the public __init__ of both surfaces alongside the other run errors.

src/evaluatorq/redteam/runner.py
```diff
@@ -139,6 +160,9 @@ def _auto_save_run(report: RedTeamReport, name: str | None = None) -> Path | Non
         data = report.model_dump(mode='json')
         data['run_name'] = resolved_name
         data['saved_at'] = datetime.now(tz=timezone.utc).isoformat()
+        if datapoints:
+            data[REPLAY_DATAPOINTS_KEY] = datapoints
```
KarinaKKarinaK, line 165: Non-blocking: the saved payload has no version marker, so this transition works by absence sniffing (no datapoints key means pre-replay run). That trick only works once. A replay_version: 1 written here and in SimulationRun would let the next format change produce a clear saved-by-an-older-version error instead of a structural one. Cheap now, expensive to retrofit once run files are in the wild.

KarinaKKarinaK (approved):
Deep re-review after the two post-approval commits. Still an approve, and a more confident one now.

Both of my earlier comments are properly closed, and I verified the code rather than the commit messages:

- Ambiguous run-name matches are now logged with the chosen file (common/replay.py:90), and the docstring spells out why names resolve newest-first while ids raise. That is exactly the split I was after.
- The hybrid double-count is pinned by test_mixed_hybrid_run_counts_the_static_leg_once, which asserts no duplicate datapoint ids per target and that all targets agree on the count, at the per-target breakdown where the bug actually surfaced. The commit notes it was verified failing on the reverted fix.

The Aug 3 repair commit holds up too: the red-team-run misdirection guard now checks the pipeline discriminator unconditionally, the resolved max_turns is stamped on the pipeline span after resolution (default and replayed cases both tested), and pipeline inference prefers datapoint tags over an under-reported stored label, with three targeted unit tests covering the contradiction, the more-specific-label and the untagged cases.

I also ran all six new test files locally: 57 pass. Test isolation checks out, the root conftest autouse fixture points EVALUATORQ_DIR at a tmp dir, so replay tests never touch a real run store.

Two non-blocking suggestions inline (public export for ReplayError, a version marker on the saved payload). One observation: running the six new test files as a single batch ends with a live OTLP span export attempt at interpreter shutdown (401 Unauthorized with my key). Each file alone is clean, and the new span fixture is properly in-memory, so something in the combination still installs a real exporter. Possibly pre-existing, but worth a look for test hygiene.

PR #93, opened 2026-07-31
feat(dashboard): add pairwise judging surface (RES-1039)

currentlycodinng:
Closes RES-1039.

RES-760 added pairwise judging, panel reconciliation and the rollup metrics, but returned all of it as in-memory data. Nothing persisted and nothing rendered. This adds the persisted run artifact and the dashboard surface that reads it.

## What was missing

Two gaps had to close before anything could be drawn:

1. **Nothing persisted.** The dashboard is a file scanner: `sniff_kind` identifies a surface from one unique field and dispatches through the `ADAPTERS` registry. Pairwise had no run artifact, no run store, and no adapter.
2. **A comparison had no identity.** `PairwiseComparison` holds a winner, the votes and token usage, but `PairwiseComparator.compare()` takes the question and the two responses and throws them away. So a per-comparison view had nothing to label rows with and no text to put beside a judge's explanation.

## The run artifact

`PairwiseRun` pairs each comparison with the inputs that produced it, records the panel configuration, and stores the rollup computed once at `save()` time. `PairwiseComparison` is untouched and stays a pure result of the judging engine.

Two fields are worth calling out:

- **`label_a` / `label_b`.** The slot letters are meaningless to a reader, since nothing in the judging data says what was in each slot. The whole view uses these instead, so the consensus bar reads `prompt-v2 67% / prompt-v3 33%`.
- **`swap`.** `JudgeStats.position_bias` falls back to `0.0` when no pair was flippable. Without knowing whether swapping ran, a single-ordering run is indistinguishable from a panel with no position bias, which is the flattering reading and the wrong one. Such a run now renders that column as unavailable.

No CLI. Producing a run means a Python loop over your own A/B pairs, which is how pairwise judging is used today. An `eq pairwise` command needs a dataset format for pairs and a decision about where the two responses come from, neither of which this ticket requires; both can land later against this file format.

## The surface

Follows the existing split exactly: `sections.py` emits `ReportSection`s, `export_html.py` dispatches by kind through `evaluatorq.common.reports`. Three sections:

- **Consensus** — one labelled bar per side, captioned with how many of the total comparisons the panel actually decided (both win rates share a denominator that excludes ties and inconclusive results, so a noisy run can otherwise post a flattering rate).
- **Judges** — per-judge win rates, tie rate and position bias, with a warning marker at or above 0.15.
- **Comparisons** — consensus winner plus one vote chip per judge in fixed column order, so a split panel reads as a ragged column. Rows expand to the two responses side by side and every judge's vote and rationale.

No filter dimensions: a pairwise run has no natural facet beyond the consensus winner. A dimensionless `FilterDef` is registered because the report route requires one, `FilterDef` gains a `results` accessor so the shared no-selection path does not need to know pairwise keeps its items under `entries`, and the filter rail is omitted rather than rendered as an empty "Filters" panel.

## Screenshots

The three sections: consensus win rates per side, the per-judge table, and the comparison list.

Pairwise run overview

A comparison row expands to the two responses side by side with every judge's vote and rationale. Row 3 is a split panel; row 4 is one the panel could not decide, where all three judges flipped and abstained.

Two expanded comparisons

Both are committed under `docs/assets/` and used by the docs site, so they stay in sync with the surface.

## Definition of done

- [x] Pairwise runs render a comparison view (consensus winner + per-judge picks)
- [x] Per-judge win rates, agreement rate, and position bias shown
- [x] Non-pairwise runs unaffected — a file with no `judging` field sniffs exactly as before, and a scanner test pins that

## Verification

- `uv run pytest -m 'not integration'` — 3263 passed, 2 skipped. 29 of those are new.
- `uv run ruff check src` — clean. `uv run basedpyright` — 0 errors.
- `uv run --group docs mkdocs build --strict` — builds.
- Rendered end to end against a synthetic run through `eq dashboard`: index, run list, report view and all three export routes.

New tests cover the JSON round trip, `save()` computing the rollup, the sniff plus the two existing surfaces staying unchanged, adapter name/timestamp/row export, label propagation, fixed judge column order, split and inconclusive marking, the `swap=False` path rendering bias as unavailable rather than zero, and the empty and fully-inconclusive runs rendering without error.

## Note for reviewers

The demo run used for the screenshots is synthetic. Generating a real one means running a three-judge panel over both orderings of every pair, which is a few hundred paid LLM calls; it would exercise the judging engine, which this PR does not touch, rather than the view, which it does. Happy to do a small live run if you want one before merge.


src/evaluatorq/pairwise_reports/sections.py
```diff
@@ -0,0 +1,156 @@
...
+    return len(decisive) > 1
+
+
+def _build_consensus_section(run: PairwiseRun) -> ReportSection:
+    report = run.rollup()
+    total = report.comparisons
+    # Win rates are computed over decided comparisons only, so a mostly
+    # inconclusive run can still post a strong-looking rate. Carry the decided
+    # count through so the renderer can say so out loud.
+    decided = round(total * (1.0 - report.tie_rate - report.inconclusive_rate))
```
KarinaKKarinaK: Minor: `decided` is reconstructed from the rounded rates rather than counted directly. `round()` makes it effectively exact here, so this is not a bug, but if `PairwiseReport` exposes a decisive/decided count (or the ties+inconclusive counts), reading it would be exact by construction and would not depend on the rate arithmetic round-tripping. Only worth it if that field already exists.

KarinaKKarinaK (approved):
Approving. Clean, well-structured surface that follows the existing dashboard split, and the write-up is honest (synthetic demo run, no CLI yet). I focused on the two things that matter most for a render surface: HTML escaping of model-generated text, and the rollup math.

Escaping is thorough. Every model- or user-authored string reaches the DOM through `_esc`: the two responses, judge rationales, model names, side labels, the question, criteria, run name, and the winner label (via `status_badge`, which escapes and clamps its status). Colors come from a fixed palette and bar widths are numeric and clamped, so there is no style-injection path either. For an exported HTML file that someone opens in a browser, this is the right bar and it is met.

Correctness holds up:
- The two anti-flattering fixes are both real and correct. `position_bias` renders as `None` (n/a) when `swap` is off rather than a flattering 0.0, and `biased` is gated on `swap`, so a single-ordering run cannot masquerade as an unbiased panel. And win rates carry the decided count so a mostly-inconclusive run cannot post a strong-looking rate without saying so.
- `sniff_kind` checks sim and red team before the `judging` discriminator, so a non-pairwise file is unaffected, and a test pins it.
- `save()` computes the rollup once and stores it; `rollup()` prefers the stored report. Empty and fully-inconclusive runs fall back to `-` via `_rate(None)` rather than dividing by zero.

Nits only, none blocking:
1. The votes comprehension in `_build_comparisons_section` mixes access styles: `vote` uses a walrus on `votes.get(model)`, while `label`/`flipped`/`explanation` use `model in votes` plus `votes[model]`. Hoisting `v = votes.get(model)` once and reading every field off `v` would be a touch clearer and avoids the repeated dict lookups. Cosmetic.
2. Repo-norm nit: the PR body carries a `Generated with Claude Code` footer. This workspace prefers no AI attribution on artifacts.

One inline note on the decided-count derivation below.

src/evaluatorq/dashboard/surfaces.py
```diff
@@ -164,14 +164,74 @@ def _sim_load(p: _Path) -> SimulationRun:
...
+        """
+        from evaluatorq.pairwise_reports.sections import judge_order, vote_label, winner_label
+
+        order = judge_order(run)
+        rows: list[dict[str, Any]] = []
+        for index, entry in enumerate(filtered, start=1):
+            votes = {v.model: v for v in entry.comparison.votes}
+            row: dict[str, Any] = {
+                '#': index,
+                'Question': entry.question,
```
KarinaKKarinaK, line 191: Non-blocking: this is the first CSV export in the repo that writes free text into cells, and the question can be model-generated. A value starting with = + - @ or a tab is executed as a formula when the CSV lands in Excel or Sheets, and given this tool renders red-team output, adversarial content in that field is not far-fetched. A tiny guard that prefixes those leading characters with a quote (applied to Question, and to the label-derived cells for completeness) closes it. The existing red team CSV only exports enum-ish values, so nothing else needs the treatment.

src/evaluatorq/pairwise_reports/export_html.py
```diff
@@ -0,0 +1,283 @@
...
+    return (
+        f'<section id="{section.kind}"><h2>{_esc(section.title)}</h2>'
+        f'<div class="pw-sides">{bars}</div>{caption}{stats}</section>'
+    )
+
+
+def _bias_cell(row: dict[str, Any]) -> str:
+    bias = row.get('position_bias')
+    if bias is None:
+        return '<span class="pw-muted" title="Run used a single ordering, so no flip was possible">n/a</span>'
```
KarinaKKarinaK: Non-blocking: after the bias-gating fix, position_bias is None in two distinct cases, swap off, or swap on but this judge never completed both orderings. This tooltip hardcodes the first explanation, so a judge with systematic ordering failures in a swap-enabled run reads as if the run never swapped. The section data already carries swap, so a per-row measured flag (or picking the title off d.get('swap')) would let the two cases explain themselves correctly.

KarinaKKarinaK (approved):
Deep re-review after the two post-approval commits. Approving again.

Everything from the earlier review is properly closed, verified in the code rather than the commit messages: the vote-cell refactor goes further than the nit asked (a single _vote_cell helper reading every field off one resolved vote), and decided is now counted directly from the entries, with a thirds-based test pinning exactly the case where the rate arithmetic would have been fragile. PairwiseReport indeed has no decided field, so counting was the right call.

The Aug 3 fix commit holds up across all five repairs, each with a regression test: the CSV ragged-keys 500 (judge_order shared between rows and sections), the bias gate now requiring both swap and a completed pair (so an unmeasurable judge reads n/a rather than a flattering 0.00), the exclusive-create save retry (genuinely atomic, matching the sim store), the back-arrow routing, and the landing panel dropping zero-count kinds.

Verified on the head: 532 dashboard tests pass with the dashboard extra, ruff clean, basedpyright 0 errors.

Two new non-blocking findings inline: the CSV export is this repo's first to write free text into cells, so it picks up the classic spreadsheet formula-injection gap, and the n/a bias tooltip now has two causes but one hardcoded explanation. Both are small.

One repo-norm nit: both follow-up commits carry a Co-Authored-By Claude trailer. Same note as the PR body footer last time, this workspace prefers no AI attribution on artifacts; if this merges by squash, please make sure the squash message drops it.

ChielToto (commented):
Findings

1. No producer. There's no eq pairwise. Red team and sim each have a CLI that auto-saves into their run store; pairwise gets a peer nav column and a peer store, but the only way to fill it is a hand-written loop ending in PairwiseRun.save(). The nav promises a first-class citizen. Is there a follow-up ticket, or does this surface sit empty?

2. swap is self-reported when the data can prove it. pairwise.py:308 sets completed=False on every vote when swap is off, so any(v.completed …), which is exactly the measured map you already build, is "did this run swap." The tooltip fix in 818d71a7 is right, but the caption (export_html.py:132) and header (:258) still trust the stored flag. A run executed single-ordering but saved with the default swap=True renders "Swap: on", no caption, and per-judge tooltips blaming the judges for a run-level fact. Deriving it closes this and lets swap go back to being recorded intent.

3. measured is dead. Added in 818d71a7 with a comment saying the renderer uses it to tell the two unmeasurable cases apart. _bias_cell branches on swap alone and measured is never read. Either wire it up (see 2) or drop it.

4. Landing.pairwise_runs is dead too. Set in metrics.py:308, read nowhere; its only consumer is a test assertion on a value the UI never shows.

5. by_kind zero-dropping is unscoped. It silently changes the landing page for every existing red-team/sim workspace, and _bars assigns colors positionally (view.py:183), so kind→color is no longer stable across workspaces. Consider a fixed map, or call the change out in the PR body.

6. The score column carries three unrelated metrics: resistance rate, mean scorer average, mean inter-judge agreement, against one >= 0.8 threshold, unlabelled, in a list that mixes surfaces. Two fixes worth splitting: add a per-surface title on the score cell (cheap, in scope), and reconsider whether mean_agreement is the right pick. It measures the panel rather than the systems being compared, it's quantized by panel size so 0.8 means different things at 2 vs 3 vs 5 judges, and it's undefined for a single-judge run. 1 - inconclusive_rate is what your own docs call the health signal, isn't panel-size-dependent, and makes 0.8 defensible. Per-surface thresholds are a separate ticket — changing _score_cls would alter red-team and sim rendering too.

Smaller: CSV header escaping is deliberate and commented but untested (a judge named =evil). build_report() runs three times per filtered render,  _with_entries nulls the cached report and three call sites each re-roll it. run.add() requires re-passing question/response_a/response_b that compare() already saw, with nothing checking they match. A mismatched pairing renders the wrong response text beside a rationale, silently and plausibly. _pairwise_adapter has a stray blank line and no docstring while its siblings have one. _pairwise_row's score path is uncovered.

Rating: 4/5. Correct, well-tested, fits the architecture rather than fighting it, and the security catch was unprompted and sound. Held back by three dead/self-reported fields where the data already knows the answer, and by an unscoped landing-page change riding along on a pairwise ticket.

currentlycodinng:
Correction to my last comment: there is a ticket now — [RES-1178](https://linear.app/orqai/issue/RES-1178/pairwise-runs-have-no-automatic-producer), backlog.

I also walked back the "wire it into orq-arena" framing. That prejudged one of three possible fixes, and the ticket now lists all of them: `pairwise-winrate` (research#232), orq-arena, or an `eq pairwise` command here. The last one is the only self-contained option, since the other two depend on a released `PairwiseRun`.

Not proposing to build any of them now — nobody has hit the gap yet, and the library path is documented. The ticket is the record so it gets picked up on real demand rather than on spec.

ChielToto (approved):

PR #94, opened 2026-07-31
refactor(redteam): registry-backed delivery methods, pass objects through as-is (RES-966)

KarinaKKarinaK:
Closes RES-966. Follow-up to RES-295.

## What

Delivery methods now mirror the vulnerability-registry pattern, and `DeliveryMethod | str` objects flow through the planner, loader filter, and reporting **unchanged** — no `enum -> string -> enum` conversion hops. This is safe because `DeliveryMethod` is a `StrEnum`: a set of members matches the dataset's string `delivery_method` directly (same hash and equality), so the round-trip the old seam performed was never needed.

## Changes

- **New `delivery_method_registry.py`** mirroring `vulnerability_registry.py`: the canonical set (the enum, grouped by technique family in `DELIVERY_METHOD_CATEGORY`) plus a mutable custom registry, with `register_delivery_method` / `is_known_delivery_method` / `resolve_delivery_method` / `resolve_delivery_methods` / `list_available_delivery_methods`.
- **Removed the conversion seam** (`runner.py`): deleted `_delivery_method_values`; `resolved_delivery_methods` is built through the registry and passed as the object set straight to the loader (3 sites) and `_check_filter_results` (display strings derived inline via `str()`, which is the value for a StrEnum).
- **Normalized once at the contract boundary** (item 4): `RedTeamInput._coerce_delivery_method` and `is_known_delivery_method` both route through the registry, so it is the single source of truth and a known value is coerced exactly once.
- **Loader** (`evaluatorq_bridge.py`): `delivery_methods` params widened to `Collection[DeliveryMethod | str]`; filter log lines render values, not enum reprs.
- **CLI** (`cli.py`): `--delivery-method` validates against registry + enum via `is_known_delivery_method`.

## Strictness policy (the decision item 3 asks for)

Delivery methods stay **coerce-known + passthrough-unknown** (the open set RES-295 deliberately chose, now registry-extensible). Vulnerabilities keep **reject-unknown**. The *registry pattern* is consistent (both: enum + registered, resolve/register); the *strictness* differs on purpose because the semantics do — an unknown delivery method is a harmless filter label (it matches a dataset row spelled the same or it does not), whereas an unknown vulnerability has no strategies and no evaluator, so passing it through would silently produce an unscoreable run. Same registry shape, different failure mode, each chosen for its own contract. This is documented in the registry module docstring; happy to make vulnerabilities open-set too if reviewers prefer literal symmetry.

**Out of scope** (per ticket): no fuzzy normalization — a value equals a known method exactly or it is unknown.

## Testing

- 14 new registry tests (`tests/redteam/test_delivery_method_registry.py`): coerce-known / passthrough-unknown / enum-identity, register + is_known + list, the StrEnum set-membership property that makes the seam removal safe, and the `RedTeamInput` boundary coercion.
- Full redteam suite: 1315 passed. `ruff` and `basedpyright` clean on the 5 changed src files.
- One existing CLI test's warning-text assertion updated (message reworded; behavior identical).

An adversarial verification swarm is running against the ticket spec (StrEnum-equality correctness, the mutable-global registry lifecycle, and whether the strictness decision holds); any confirmed findings will land as follow-up commits.

Ticket: RES-966

### DIFF
--- a/docs/custom-evaluators-and-frameworks.md
+++ b/docs/custom-evaluators-and-frameworks.md
@@ -230,6 +230,31 @@ Strategies can declare capability requirements to skip attacks that don't apply 
 
 Available capability tags: `code_execution`, `shell_access`, `file_system`, `web_request`, `database`, `email`, `messaging`, `memory_read`, `memory_write`, `knowledge_retrieval`, `user_data`.
 
+### Custom delivery methods
+
+`delivery_method` is an **open set**. The canonical methods live in the `DeliveryMethod`
+enum (each mapped to a technique family in `DELIVERY_METHOD_CATEGORY`), and
+`delivery_method_registry.py` mirrors the vulnerability registry so you can add your own
+without touching the enum:
+
+```python
+from evaluatorq.redteam.delivery_method_registry import (
+    register_delivery_method,
+    is_known_delivery_method,
+)
+
+# Register a custom method so it validates and filters as known.
+register_delivery_method('emoji-smuggling', category='obfuscation')
+
+is_known_delivery_method('emoji-smuggling')  # True
+```
+
+Unlike vulnerabilities (reject-unknown, since an unknown vuln has no strategies or
+evaluator), delivery methods are **coerce-known + passthrough-unknown**: an unregistered
+value is a harmless filter label that either matches a dataset row spelled the same or
+does not. Registering is only needed when you want `is_known_delivery_method` /
+`--delivery-method` to treat your value as canonical rather than passthrough.
+
 ## Adding a new framework
 
 Frameworks are a reporting/compliance layer on top of vulnerabilities. Adding a framework means:
--- a/docs/custom-evaluators-and-frameworks.md
+++ b/docs/custom-evaluators-and-frameworks.md
@@ -253,8 +253,10 @@ Unlike vulnerabilities (reject-unknown, since an unknown vuln has no strategies 
 evaluator), delivery methods are **coerce-known + passthrough-unknown**: an unregistered
 value is a harmless filter label that either matches a dataset row spelled the same or
 does not. Filtering therefore works without registering anything — registering only
-suppresses the "unknown delivery method" warning that `--delivery-method` and the
-`RedTeamInput` boundary emit. A registered value stays a plain string; only enum members
+suppresses the "unknown delivery method" warnings: the `--delivery-method` CLI flag warns up
+front via `typer.echo`, and a programmatic `red_team()` run surfaces an unmatched method through
+the pipeline's post-filter check as a `loguru` warning (the `RedTeamInput` validator itself resolves
+silently). A registered value stays a plain string; only enum members
 resolve to a `DeliveryMethod` object.
 
 The registry is **in-memory and process-local** — it is not persisted and there is no
--- a/docs/custom-evaluators-and-frameworks.md
+++ b/docs/custom-evaluators-and-frameworks.md
@@ -243,7 +243,7 @@ from evaluatorq.redteam.delivery_method_registry import (
     is_known_delivery_method,
 )
 
-# Register a custom method so it validates and filters as known.
+# Register a custom method so it validates as known.
 register_delivery_method('emoji-smuggling', category='obfuscation')
 
 is_known_delivery_method('emoji-smuggling')  # True
@@ -252,8 +252,16 @@ is_known_delivery_method('emoji-smuggling')  # True
 Unlike vulnerabilities (reject-unknown, since an unknown vuln has no strategies or
 evaluator), delivery methods are **coerce-known + passthrough-unknown**: an unregistered
 value is a harmless filter label that either matches a dataset row spelled the same or
-does not. Registering is only needed when you want `is_known_delivery_method` /
-`--delivery-method` to treat your value as canonical rather than passthrough.
+does not. Filtering therefore works without registering anything — registering only
+suppresses the "unknown delivery method" warning that `--delivery-method` and the
+`RedTeamInput` boundary emit. A registered value stays a plain string; only enum members
+resolve to a `DeliveryMethod` object.
+
+The registry is **in-memory and process-local** — it is not persisted and there is no
+plugin/entry-point loading. Registering in a standalone script does not make the value
+known to a separate `eq redteam run` process; to get the CLI benefit, register in the
+same process that invokes the CLI (or accept the warning, since filtering works either
+way).
 
 ## Adding a new framework
 
--- a/src/evaluatorq/redteam/delivery_method_registry.py
+++ b/src/evaluatorq/redteam/delivery_method_registry.py
@@ -44,6 +44,8 @@ from __future__ import annotations
 import types
 from typing import TYPE_CHECKING
 
+from loguru import logger
+
 from evaluatorq.redteam.contracts import DeliveryMethod
 
 if TYPE_CHECKING:
@@ -77,8 +79,10 @@ def delivery_method_str(value: DeliveryMethod | str) -> str:
 # The one piece of real structure delivery methods carry; kept so ``register``
 # has somewhere to place a custom method and the registry mirrors the
 # vulnerability registry's structured defs rather than being a bare set.
-# Category is the only metadata a consumer needs today; add richer per-method
-# defs here if one ever does.
+# The category value itself has no runtime consumer today (the import-time
+# completeness check below inspects keys only) — it is structural/placeholder
+# metadata kept so the shape matches the vulnerability registry and a future
+# consumer has somewhere to read from. Add richer per-method defs here if one does.
 DELIVERY_METHOD_CATEGORY: Mapping[DeliveryMethod, str] = {
     DeliveryMethod.DAN: 'persona',
     DeliveryMethod.ROLE_PLAY: 'persona',
@@ -122,12 +126,12 @@ def register_delivery_method(value: str, *, category: str = 'custom') -> str:
     CLI's unknown-method warning; filtering already worked without it.
 
     A value equal to an existing :class:`DeliveryMethod` member is a no-op (the
-    enum is already canonical); a member *name* that is not also a value is
-    rejected, since registering ``'ROLE_PLAY'`` would shadow ``'role-play'``
-    with a spelling no dataset row carries. Registration is process-local and
-    not persisted, so it must happen in the same process that later runs the
-    CLI. Returns the registered value, or raises ``ValueError`` on an empty
-    value or a member name.
+    enum is already canonical) — it warns and discards the passed ``category``
+    rather than storing it. A member *name* that is not also a value is rejected,
+    since registering ``'ROLE_PLAY'`` would shadow ``'role-play'`` with a spelling
+    no dataset row carries. Registration is process-local and not persisted, so it
+    must happen in the same process that later runs the CLI. Returns the registered
+    value, or raises ``ValueError`` on an empty value or a member name.
     """
     if not value:
         msg = 'Cannot register an empty delivery method.'
@@ -140,7 +144,13 @@ def register_delivery_method(value: str, *, category: str = 'custom') -> str:
     except ValueError:
         pass
     else:
-        return value  # already canonical — no-op
+        # Already canonical: no-op. Warn (not silent) and drop the category —
+        # there is no custom entry to attach it to.
+        logger.warning(
+            f'{value!r} is already a canonical delivery method; register_delivery_method is a no-op '
+            f'(category {category!r} discarded).'
+        )
+        return value
     member_by_name = DeliveryMethod.__members__.get(value)
     if member_by_name is not None:
         msg = (
--- a/src/evaluatorq/redteam/delivery_method_registry.py
+++ b/src/evaluatorq/redteam/delivery_method_registry.py
@@ -25,6 +25,12 @@ mode, each chosen for its own contract.
 
 No fuzzy normalization: a value equals a known method exactly, or it is unknown
 (RES-295 non-goal, unchanged).
+
+Registry lifecycle: registrations are **process-global and idempotent** — a
+custom method registered by one call is known to every later call in the same
+process, by design (that is what a registry is for). It is never persisted to
+disk, so a fresh process starts with only the enum. Tests that register a custom
+method should reset ``_CUSTOM_DELIVERY_METHODS`` between cases to stay isolated.
 """
 
 from __future__ import annotations
@@ -33,12 +39,25 @@ from evaluatorq.redteam.contracts import DeliveryMethod
 
 __all__ = [
     'DELIVERY_METHOD_CATEGORY',
+    'delivery_method_str',
     'is_known_delivery_method',
     'list_available_delivery_methods',
     'register_delivery_method',
     'resolve_delivery_method',
     'resolve_delivery_methods',
 ]
+
+
+def delivery_method_str(value: DeliveryMethod | str) -> str:
+    """Canonical string for a delivery method — the value, version-independently.
+
+    ``str(DeliveryMethod.DAN)`` returns ``'DAN'`` on native 3.11+ ``StrEnum`` but
+    the ``'DeliveryMethod.DAN'`` repr on the 3.10 ``StrEnum`` polyfill (which has
+    no ``__str__`` override). So never call ``str()`` on a member for display or
+    string-keying — use this. Filtering is unaffected (a member compares/hashes
+    equal to its value on every version); only string rendering diverges.
+    """
+    return value.value if isinstance(value, DeliveryMethod) else value
 
 
 # Technique family for each canonical method (the groupings the enum documents).
@@ -121,7 +140,7 @@ def resolve_delivery_methods(values: list[DeliveryMethod | str]) -> list[Deliver
     result: list[DeliveryMethod | str] = []
     for value in values:
         resolved = resolve_delivery_method(value)
-        key = str(resolved)
+        key = delivery_method_str(resolved)
         if key not in seen:
             seen.add(key)
             result.append(resolved)
--- a/src/evaluatorq/redteam/delivery_method_registry.py
+++ b/src/evaluatorq/redteam/delivery_method_registry.py
@@ -7,11 +7,13 @@ and API validate against **registry plus enum** rather than a frozen enum.
 
 Strictness policy (RES-966 decision)
 ------------------------------------
-Delivery methods are **coerce-known + passthrough-unknown**: a value that is a
-known method (enum member or a registered custom) resolves to its canonical
-:class:`DeliveryMethod` object; anything else passes through as a raw string.
-This preserves the open set RES-295 deliberately chose (a dataset may carry a
-delivery method the enum does not list), now extensible via
+Delivery methods are **coerce-known + passthrough-unknown**: an enum member
+resolves to its canonical :class:`DeliveryMethod` object; anything else passes
+through as a raw string. A *registered custom* is known (it validates without a
+warning) but stays a string — there is no enum member for it to resolve to, so
+registration's observable effect is :func:`is_known_delivery_method` returning
+``True``. This preserves the open set RES-295 deliberately chose (a dataset may
+carry a delivery method the enum does not list), now extensible via
 :func:`register_delivery_method`.
 
 This differs on purpose from :func:`vulnerability_registry.resolve_vulnerabilities`,
@@ -26,16 +28,24 @@ mode, each chosen for its own contract.
 No fuzzy normalization: a value equals a known method exactly, or it is unknown
 (RES-295 non-goal, unchanged).
 
-Registry lifecycle: registrations are **process-global and idempotent** — a
-custom method registered by one call is known to every later call in the same
-process, by design (that is what a registry is for). It is never persisted to
-disk, so a fresh process starts with only the enum. Tests that register a custom
+Registry lifecycle: registrations are **process-global** — a custom method
+registered by one call is known to every later call in the same process, by
+design (that is what a registry is for). Re-registering the same value is
+last-write-wins on its category. Registrations are never persisted to disk, so a
+fresh process starts with only the enum: a value registered in one script is
+*not* known to a separate ``eq redteam`` invocation. Tests that register a custom
 method should reset ``_CUSTOM_DELIVERY_METHODS`` between cases to stay isolated.
 """
 
 from __future__ import annotations
 
+import types
+from typing import TYPE_CHECKING
+
 from evaluatorq.redteam.contracts import DeliveryMethod
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
 
 __all__ = [
     'DELIVERY_METHOD_CATEGORY',
@@ -66,7 +76,7 @@ def delivery_method_str(value: DeliveryMethod | str) -> str:
 # vulnerability registry's structured defs rather than being a bare set.
 # Category is the only metadata a consumer needs today; add richer per-method
 # defs here if one ever does.
-DELIVERY_METHOD_CATEGORY: dict[DeliveryMethod, str] = {
+DELIVERY_METHOD_CATEGORY: Mapping[DeliveryMethod, str] = {
     DeliveryMethod.DAN: 'persona',
     DeliveryMethod.ROLE_PLAY: 'persona',
     DeliveryMethod.SKELETON_KEY: 'persona',
@@ -85,6 +95,15 @@ DELIVERY_METHOD_CATEGORY: dict[DeliveryMethod, str] = {
     DeliveryMethod.WORD_SUBSTITUTION: 'tool-agent',
 }
 
+# Fail at import, not in CI: a new enum member without a technique family would
+# otherwise ship silently. Frozen afterwards so no caller can mutate the
+# canonical set at runtime (mirrors vulnerability_registry.py).
+_missing = sorted(delivery_method_str(m) for m in set(DeliveryMethod) - set(DELIVERY_METHOD_CATEGORY))
+if _missing:
+    raise RuntimeError(f'Missing DELIVERY_METHOD_CATEGORY entries for: {_missing}')
+
+DELIVERY_METHOD_CATEGORY = types.MappingProxyType(dict(DELIVERY_METHOD_CATEGORY))  # pyright: ignore[reportConstantRedefinition]
+
 # Registered custom methods: value -> category. Seeded empty; the enum members
 # are always known without being listed here. ``register_delivery_method``
 # appends to this so ``is_known``/``resolve`` see the addition immediately.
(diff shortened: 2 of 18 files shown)


KarinaKKarinaK:
Verification swarm fixes landed in 9199866.

The swarm caught a real critical bug: I assumed `str(DeliveryMethod.X)` returns the value, but that only holds on native 3.11+ `StrEnum`. On Python 3.10 (supported, in the CI matrix), the repo's `StrEnum` polyfill has no `__str__` override, so `str(DeliveryMethod.DAN)` returns `'DeliveryMethod.DAN'`. Filtering was always safe (str-subclass equality/hash — the data path never used `str()`), but every display/keying path broke on 3.10:
- `_check_filter_results` set difference reported a method that matched every datapoint as UNMATCHED, and leaked `'DeliveryMethod.DAN'` into the user-facing empty-run `RedTeamError`.
- The CLI unknown-method warning and the bridge filter logs printed `'DeliveryMethod.X'` instead of the token the user must type.

Fix: a version-independent `delivery_method_str()` in the registry (`m.value` for a member, the string otherwise), used at every `str(m)` site, including the `resolve_delivery_methods` dedup key. Added 4 regression tests (the helper contract; `_check_filter_results` not mis-reporting a matched enum method and not leaking reprs in the empty-run error) and pointed the existing `str(m)` test assertions at the helper.

Full redteam suite: 1319 passing; ruff + basedpyright clean on src. The strictness-policy decision and the process-global registry lifecycle were confirmed as defensible/intentional (the latter now documented in the module).

currentlycodinng (approved):
**Approve.** Clean refactor and the core reasoning is sound: `DeliveryMethod` is a `StrEnum`, so a member compares and hashes equal to its string value — I verified the set-intersection filtering (`{DeliveryMethod.DAN} & {"DAN"}`) and dataset-string membership directly, which is the invariant that makes removing the `enum -> str -> enum` seam safe rather than clever. 18/18 registry tests pass and CI is green across 3.10-3.13.

Nice catch on the one place this actually bites: the 3.10 `StrEnum` polyfill has no `__str__`, so `str(member)` renders `DeliveryMethod.DAN` instead of `DAN` — and `runner.py`/`_check_filter_results` correctly route display through `delivery_method_str` (not bare `str()`), with the comment explaining why. That's the subtle failure the refactor could have introduced and didn't. (The PR description says "display via `str()` inline", which undersells the actual code — it uses the version-safe helper.)

The strictness asymmetry you flagged for review — delivery methods coerce-known/passthrough-unknown, vulnerabilities reject-unknown — is the right call, not an inconsistency: an unknown delivery method is a harmless filter label that either matches a dataset row or doesn't, whereas an unknown vulnerability yields no strategies/evaluator and a silently unscoreable run. Same registry shape, different failure mode, each matched to its contract. I'd keep it as-is rather than force literal symmetry.

One thing to keep an eye on: `_CUSTOM_DELIVERY_METHODS` is process-global mutable state — the docstring's "reset between tests" note is the right mitigation, just make sure any future test that registers a custom method actually does the teardown so isolation doesn't rot.

KarinaKKarinaK:
Smoke tested the refactored delivery-method path live against an agent (static mode), since a behavior-preserving refactor is exactly where mocked tests can pass while the real path breaks. All correct:

- --delivery-method base64 pulled 3 attacks, all with delivery_methods ['base64'], no leakage from other methods. The object passthrough through the loader filter matches the dataset strings correctly, so removing the enum to string to enum seam is safe in practice.
- Unfiltered run pulled a diverse mix (crescendo, many-shot, refusal-suppression, authority-impersonation, direct-request), confirming the base64 run actually filtered rather than coincidentally getting base64 only.
- A typo'd method (base644-typo) gives a clean actionable error listing every valid method, which is the documented coerce-known plus passthrough-unknown open set policy, not a silent empty run.
- Display strings render as values (base64), not enum reprs.

No issues found. LGTM from a live run.

arianpasquali (dismissed):
Three small fixes before merge:

-  Docs: add delivery_method_registry.py to the CLAUDE.md package tree and mention register_delivery_method in docs/custom-evaluators-and-frameworks.md.

- Strictness policy: ticket says decide it “consistently for both”. PR keeps them different on purpose. Fine by me, but @bauke wrote the ticket, so he needs to OK it here.

- reports/converters.py ~line 907: dict keys built from DeliveryMethod objects without delivery_method_str. Works only via pydantic key coercion; leaks DeliveryMethod.X on 3.10 otherwise. Use the helper.

Non-blocking, but still

-  register_delivery_method / DELIVERY_METHOD_CATEGORY have zero callers. Drop until needed, or fix the asymmetry (resolve ignores custom registry, is_known uses it).
-  4 dashboard files hand-roll getattr(dm, "value", dm). Point them at delivery_method_str in a follow-up.

- Add a completeness test: every enum member present in DELIVERY_METHOD_CATEGORY.
-  Remove leftover # ponytail: comment.

KarinaKKarinaK:
Thanks Arian, all applied in 9eec05a:

- Docs: added delivery_method_registry.py to the CLAUDE.md package tree and a Custom delivery methods section in docs/custom-evaluators-and-frameworks.md showing register_delivery_method / is_known_delivery_method and the coerce-known plus passthrough-unknown policy.
- converters.py: the by-delivery-method dict is now keyed via delivery_method_str, so a DeliveryMethod member can never leak its 3.10 repr into report keys.
- Dashboard: the 4 hand-rolled getattr(dm, 'value', dm) sites (filters.py x2, redteam_transcripts.py, redteam_charts.py) now use delivery_method_str. Done here rather than a follow-up since it is a one-line swap each.
- Completeness test: test_every_enum_member_has_a_category asserts DELIVERY_METHOD_CATEGORY maps every DeliveryMethod member, so a new method cannot ship without a technique family.
- Removed the leftover ponytail: comment.

On the two I did not change:
- register_delivery_method / DELIVERY_METHOD_CATEGORY: kept rather than dropped. They mirror register_vulnerability in the vulnerability registry as the public extension API, DELIVERY_METHOD_CATEGORY now has the completeness test as a consumer, and the register path is covered by the existing registry tests. On the asymmetry: resolve intentionally returns a registered custom as its string, because a custom method has no enum to coerce to, that passthrough IS the resolution for an open set, whereas is_known just answers the membership question. Happy to drop both if you would rather keep the surface minimal.
- Strictness policy stays coerce-known for delivery methods vs reject-unknown for vulnerabilities. Tagging @Baukebrenninkmeijer since he wrote the ticket: are you OK keeping these different (rationale is in the registry docstring, different failure modes), or do you want literal symmetry?

Also smoke tested the whole path live against an agent earlier (static mode): base64 filter pulled only base64 attacks, a typo gives a clean error listing valid methods. 649 report/dashboard/registry tests pass.

src/evaluatorq/redteam/delivery_method_registry.py
```diff
@@ -0,0 +1,152 @@
...
+Mirrors :mod:`evaluatorq.redteam.vulnerability_registry`: a canonical set (the
+:class:`DeliveryMethod` enum, grouped by technique family) plus a mutable
+registry for custom entries, with ``register``/``resolve`` helpers so the CLI
+and API validate against **registry plus enum** rather than a frozen enum.
+
+Strictness policy (RES-966 decision)
+------------------------------------
+Delivery methods are **coerce-known + passthrough-unknown**: a value that is a
+known method (enum member or a registered custom) resolves to its canonical
+:class:`DeliveryMethod` object; anything else passes through as a raw string.
```
Baukebrenninkmeijer: Docstring contradicts the code below it. It says a known method — "**enum member or a registered custom**" — "resolves to its canonical `DeliveryMethod` object", but `resolve_delivery_method` (line 121) never consults `_CUSTOM_DELIVERY_METHODS`; a registered custom cannot become a member because no member exists. The PR's own `test_contract_boundary_coerces_registered_custom_to_string` (test file line 136) asserts the real behavior.

So registration's entire observable effect is `is_known_delivery_method` returning `True`, which suppresses the CLI warning at `cli.py:427`. Legit feature — just say that.

Suggested: "an enum member resolves to its `DeliveryMethod` object; a registered custom is *known* but stays a raw string." Same wording is mirrored at `contracts.py:486` and needs the same fix.

Minor, same docstring: line 29 calls registration "idempotent", but re-registering with a different `category` silently overwrites — that's last-write-wins.

src/evaluatorq/redteam/delivery_method_registry.py
```diff
@@ -0,0 +1,152 @@
...
+    DeliveryMethod.CRESCENDO: 'multi-turn',
+    DeliveryMethod.MANY_SHOT: 'multi-turn',
+    DeliveryMethod.AUTHORITY_IMPERSONATION: 'social-engineering',
+    DeliveryMethod.REFUSAL_SUPPRESSION: 'social-engineering',
+    DeliveryMethod.DIRECT_REQUEST: 'direct',
+    DeliveryMethod.CODE_ELICITATION: 'direct',
+    DeliveryMethod.CODE_ASSISTANCE: 'direct',
+    DeliveryMethod.TOOL_RESPONSE: 'tool-agent',
+    DeliveryMethod.WORD_SUBSTITUTION: 'tool-agent',
+}
```
Baukebrenninkmeijer, line 103: `DELIVERY_METHOD_CATEGORY` is a public, plain, mutable dict with no import-time completeness check. The prior art this module explicitly mirrors does both (`vulnerability_registry.py:184-188`):

```python
_missing = set(DeliveryMethod) - set(DELIVERY_METHOD_CATEGORY)
if _missing:
    raise RuntimeError(f'Missing DELIVERY_METHOD_CATEGORY entries for: {_missing}')
DELIVERY_METHOD_CATEGORY = types.MappingProxyType(DELIVERY_METHOD_CATEGORY)
```

Without it, adding a `DeliveryMethod` member ships it with no technique family, guarded only by a unit test that fires in CI rather than at import — and any caller can mutate or delete entries at runtime. Four lines, and it's the single highest-value change here since "mirror the prior art" is the PR's stated goal.

Separately: nothing reads a category back. `register_delivery_method(value, category=...)` writes into `_CUSTOM_DELIVERY_METHODS` and no production code ever reads the value side. Either freeze this as documented taxonomy or drop the `category` parameter until a consumer needs it.

docs/custom-evaluators-and-frameworks.md
```diff
@@ -230,6 +230,31 @@ Strategies can declare capability requirements to skip attacks that don't apply
...
+# Register a custom method so it validates and filters as known.
+register_delivery_method('emoji-smuggling', category='obfuscation')
+
+is_known_delivery_method('emoji-smuggling')  # True
+```
+
+Unlike vulnerabilities (reject-unknown, since an unknown vuln has no strategies or
+evaluator), delivery methods are **coerce-known + passthrough-unknown**: an unregistered
+value is a harmless filter label that either matches a dataset row spelled the same or
+does not. Registering is only needed when you want `is_known_delivery_method` /
```
Baukebrenninkmeijer: This oversells what registration buys, in the one direction a user will actually try.

`_CUSTOM_DELIVERY_METHODS` is an in-memory dict with no persistence and no entry-point/plugin loading anywhere in the package. So a user who calls `register_delivery_method('emoji-smuggling')` in their own script and then runs `eq redteam run -d emoji-smuggling` in a **separate process** gets a fresh, empty custom registry and the "not known delivery methods" warning regardless. The documented `--delivery-method` benefit is only reachable in-process (library caller, or `CliRunner` in tests) — which is likely why no test caught it: the registry tests and the CLI-warning test never compose.

Either add a sentence that registration must happen in the same process before invoking the CLI, or wire an entry-point loader. Also worth a `CliRunner` test that registers then runs `-d emoji-smuggling` and asserts no warning — it'll pass today, and writing it pins the process boundary.

tests/redteam/test_delivery_method_registry.py
```diff
@@ -0,0 +1,197 @@
...
+    assert dp.delivery_method is DeliveryMethod.CRESCENDO
+
+
+def test_contract_boundary_passes_unknown_through() -> None:
+    dp = _input('bespoke-technique')
+    assert dp.delivery_method == 'bespoke-technique'
+    assert not isinstance(dp.delivery_method, DeliveryMethod)
+
+
+def test_contract_boundary_coerces_registered_custom_to_string() -> None:
```
Baukebrenninkmeijer, line 169: Coverage gap: the unknown/passthrough path is never exercised end to end. The CLI warn-and-continue test (`test_cli.py:366`) mocks `red_team` out, and the only e2e empty-match test (`e2e/test_static_pipeline.py:129`) uses `DeliveryMethod.BASE64` — a *known* enum, so it takes the coerced-object branch. An unknown string takes a different route through `resolve_delivery_methods`, the `set()`, the loader's `in selected`, and `delivery_method_str` in the error message. A regression there (e.g. unknowns getting dropped instead of passed through) leaves every existing test green.

Smallest fix, ~10 lines next to `test_static_delivery_method_empty_match_hard_fails`: run with `delivery_methods=['bespoke-technique']` and assert the `RedTeamError` names both `bespoke-technique` and the dataset's present `direct-request`.

Second gap, same theme: this PR switched four dashboard call sites (`dashboard/filters.py:88,161`, `redteam_charts.py:98`, `redteam_transcripts.py:57`) from `getattr(dm, 'value', dm)` to `delivery_method_str`, and no test asserts any produced string — `test_filter.py:340` only checks the `'delivery_method'` key exists. On 3.10 a revert to `str(dm)` puts `'DeliveryMethod.DIRECT_REQUEST'` in the option list and chart labels with nothing failing. Two assertions on the existing `_rt_report()` fixture would cover it. (`reports/converters.py` *is* covered — `test_converters.py:236` asserts the `'role-play'` key, and name≠value makes that a real test.)

tests/redteam/test_delivery_method_registry.py
```diff
@@ -0,0 +1,197 @@
...
+def test_register_makes_custom_known_and_resolvable_as_string() -> None:
+    assert not reg.is_known_delivery_method('honeypot')
+    reg.register_delivery_method('honeypot', category='experimental')
+    assert reg.is_known_delivery_method('honeypot')
+    # Registered customs stay strings (open set): no enum member exists for them.
+    assert reg.resolve_delivery_method('honeypot') == 'honeypot'
+    assert 'honeypot' in {reg.delivery_method_str(m) for m in reg.list_available_delivery_methods()}
+
+
+def test_register_existing_enum_value_is_noop() -> None:
```
Baukebrenninkmeijer: This pins the *value* spelling (`'crescendo'`) as a no-op, but the member-*name* spelling is the trap: `register_delivery_method('ROLE_PLAY')` fails the `DeliveryMethod('ROLE_PLAY')` lookup, so it lands in `_CUSTOM_DELIVERY_METHODS`, reports as known, suppresses the CLI warning, and then filters against a value no dataset row carries — a run that looks validated and matches nothing. One assertion here would make that behavior deliberate rather than accidental (reject enum names, or at least document that they shadow).

Related, on the fixture at line 16: the reset reaches into the private `reg._CUSTOM_DELIVERY_METHODS`, and the module docstring instructs users to do the same. Isolation is correct *within this file*, but there's no public reset, so any future test that registers outside this module leaks process-wide — the concrete victim is `test_cli.py::TestDeliveryMethodFlag`, which asserts a warning for `'not-a-real-method'` and would flip to an order-dependent failure. A public `clear_custom_delivery_methods()` (or `unregister_delivery_method`) plus moving the autouse fixture to `tests/redteam/conftest.py` turns a docstring instruction into an API.

Baukebrenninkmeijer (commented):
Ran a focused review pass over the diff (correctness / type design / test coverage / silent failures) and then verified each finding against the code myself. **No correctness bugs** — the load-bearing invariant holds: `DeliveryMethod` is a `StrEnum`, so a member hashes and compares by *value*, and every membership site (`evaluatorq_bridge.py:531,570`, `strategy_registry.py:229`, `runner.py:1208`) matches raw dataset strings directly. Dropping `_delivery_method_values` is safe. Routing every render through `delivery_method_str` instead of `str()` is the right call for the 3.10 polyfill and is done consistently.

Five things worth fixing before merge, in impact order. Details inline.

1. **Module docstring contradicts `resolve_delivery_method`** — a registered custom does *not* become a `DeliveryMethod`. Doc-only fix.
2. **`DELIVERY_METHOD_CATEGORY` is unfrozen and has no import-time completeness check**, unlike the `vulnerability_registry.py` prior art it mirrors. Four copied lines.
3. **Docs oversell registration**: it is process-global in-memory with no entry-point loading, so the `--delivery-method` benefit is unreachable from a separate CLI invocation.
4. **The unknown/passthrough path has no end-to-end test** — the only e2e empty-match test uses a *known* enum, so it exercises a different branch.
5. **`register_delivery_method('ROLE_PLAY')`** (name spelling, not value) registers a shadow method that reports known and matches nothing.

Also: `test_delivery_method_set_membership_matches_string` and `test_check_filter_results_matched_enum_method_no_unmatched_warning` both use `DeliveryMethod.DAN`, whose name equals its value — so they cannot distinguish value-based from name-based hashing, the exact property they claim to pin. Swap in `CRESCENDO`/`'crescendo'` and they become real regression tests.

Non-blocking, pre-existing: `red_team(delivery_methods=[])` yields an empty set, not `None`. The static loader's `if delivery_methods:` then skips filtering (full sweep, exit 0) while `strategy_registry.py:222` checks `is not None` and filters everything out — empty means "everything" on one path and "nothing" on the other. Same on `main`, not introduced here; the registry hop just adds a second place a value could be dropped into that state.

Baukebrenninkmeijer:
Pushed `9dd5e75` addressing items 1–3 from my review (docs/wording only + a 4-line import guard — no behavior change):

1. **Docstrings corrected** (`delivery_method_registry.py` module + `register_delivery_method` + `resolve_delivery_method`, and the mirrored text at `contracts.py:486`): only enum members resolve to a `DeliveryMethod` object; a registered custom is *known* but stays a string. Registration's observable effect is `is_known_delivery_method` → `True`, i.e. suppressing the CLI warning — filtering already works without it. Also dropped "idempotent" (re-registering is last-write-wins on `category`).
2. **`DELIVERY_METHOD_CATEGORY` now mirrors `vulnerability_registry.py:184-188`**: import-time `_missing` check (a new enum member with no technique family now fails at import instead of only in CI) then `MappingProxyType`. Annotation changed `dict` → `Mapping` to match the prior art.
3. **Docs**: added that the registry is in-memory and process-local — registering in a standalone script does *not* make the value known to a separate `eq redteam run` process — and that filtering works regardless.

`ruff check src` clean, `basedpyright` clean on both files, `tests/redteam` 1289 passed (the 2 failures are `huggingface_hub` missing in my worktree, unrelated).

Items 4–5 (unknown-path e2e test, dashboard string assertions, `DAN`→`CRESCENDO` in the two name==value tests, public `clear_custom_delivery_methods()`, `ui/dashboard.py:1973`) left as follow-up — happy to open a ticket. @arianpasquali your four items were already fixed in `9eec05a`; this branch needs your CHANGES_REQUESTED cleared to merge.

Baukebrenninkmeijer:
Pushed `3fcb6ec` — follow-up items 4–5 done here rather than deferred.

- **Unknown-delivery e2e** (`test_static_pipeline.py`): the existing empty-match test uses a known enum, so it takes the coerced-object branch. New test drives `delivery_methods=['bespoke-technique']` down the raw-string branch (resolve → loader filter → error) and asserts the error names both the requested value and the dataset's `direct-request`. Guards against a passthrough regression silently disabling the filter and running the whole dataset.
- **`register_delivery_method` now rejects a member *name***: `'ROLE_PLAY'` used to land in the custom set (reported known, warning suppressed, filtering on a spelling no row carries). Raises with a pointer to `'role-play'`.
- **Public `clear_custom_delivery_methods()`** + isolation fixture moved to `tests/redteam/conftest.py`. Registration is process-global, so a registering test outside the registry module could flip `test_cli`'s unknown-warning assertion by order. Tests no longer touch `_CUSTOM_DELIVERY_METHODS`.
- **`DAN` → `CRESCENDO`** in the two membership/warning tests. `DAN`'s name equals its value, so they couldn't distinguish value- from name-based matching — the property they exist to pin. Added an explicit `'CRESCENDO' not in selected`.
- **Dashboard assertions**: `options()['delivery_method'] == ['direct-request']` and an apply round-trip, covering the `filters.py` switch to `delivery_method_str`. `ui/dashboard.py:1973` now uses the helper instead of `getattr(dm, 'value', dm)`.

Full suite green in a fully-synced worktree: **3257 passed, 2 skipped** (random order), `ruff check src` and `basedpyright src` clean.

From my earlier review that's everything except the pre-existing `delivery_methods=[]` asymmetry (empty means "everything" in the static loader, "nothing" in `strategy_registry`) — same on `main`, left alone deliberately.

Baukebrenninkmeijer:
Pushed `ed089e2` — fixes a regression I introduced in `3fcb6ec`, found by an adversarial review pass.

**The bug:** `register_delivery_method('DAN')` raised `ValueError`. `DAN = 'DAN'` is the one member spelled identically as name and value, so my name check — which ran *before* the value lookup — treated the canonical value as a member name and told the caller to "use 'DAN'" instead of 'DAN'. Every other member has name != value, so the documented no-op path worked for all 15 of them.

**Fix:** value lookup runs first and returns the no-op early; the name rejection now only fires for a name that is not also a value (`'ROLE_PLAY'`), which is the case it was written for.

**Why the tests missed it:** they exercise `'crescendo'` and `'ROLE_PLAY'` — both name != value. That is the exact blind spot that motivated the `DAN` → `CRESCENDO` swap two commits ago, and I walked straight into the mirror image of it. Replaced the single-value assertion with a parametrized sweep over every `DeliveryMethod` member, so this class of bug can't hide in whichever member happens to be special.

Also finishes the `delivery_method_str` cleanup: `ui/dashboard.py` had two more `getattr(dm, 'value', dm)` sites at lines 369 and 478 (the filter rail) that I missed when I did line 1974. All three now use the helper; no `getattr(dm, 'value', dm)` left in `src/`.

Suite: **3338 passed, 2 skipped**; `ruff check src`, `ruff format --check src`, `basedpyright src tests` all clean.

src/evaluatorq/redteam/delivery_method_registry.py
Baukebrenninkmeijer: Fixed in 9dd5e75. The module docstring now says an enum member resolves to its `DeliveryMethod` object while a registered custom stays a raw string, and states that registration's observable effect is `is_known_delivery_method` returning `True` (suppressing the `cli.py` warning). Same wording applied to the `_coerce_delivery_method` docstring in `contracts.py`. The false "idempotent" claim is gone — it is last-write-wins on `category`. Resolving.

src/evaluatorq/redteam/delivery_method_registry.py
Baukebrenninkmeijer: Fixed in 9dd5e75, mirroring `vulnerability_registry.py:184-188`: import-time completeness check raising `RuntimeError`, then `types.MappingProxyType` freeze, with the annotation widened to `Mapping[DeliveryMethod, str]`. A unit test asserting completeness is in `test_delivery_method_registry.py:209` as well.

On the second half — `category` is written but never read back — I have deliberately left the parameter in place rather than dropping it from a public function on someone else's PR. Flagging it for @karinakalicka as a follow-up call: either a consumer reads the taxonomy or the parameter goes. Not merge-blocking. Resolving.

docs/custom-evaluators-and-frameworks.md
Baukebrenninkmeijer: Fixed in 9eec05a. The docs section now states that filtering works without registering at all, that a registered value stays a plain string, and — the point of this comment — that the registry is in-memory and process-local, so registering in a standalone script does **not** carry into a separate `eq redteam run`. No entry-point loader was added; that would be a real feature, not a doc fix. Resolving.

tests/redteam/test_delivery_method_registry.py
Baukebrenninkmeijer: Both gaps closed in 3fcb6ec.

Passthrough e2e: `test_static_unknown_delivery_method_passes_through_and_hard_fails` runs with `delivery_methods=["bespoke-technique"]` and asserts the `RedTeamError` names both `bespoke-technique` (the value survived as-is) and `direct-request` (what the dataset carries) — so a regression that drops unknowns instead of passing them through now fails.

Dashboard strings: `test_filter.py` gained `test_redteam_delivery_method_options_use_values_not_enum_reprs`, asserting `options(report)["delivery_method"] == ["direct-request"]` — a revert to `str(dm)` puts `DeliveryMethod.DIRECT_REQUEST` there on 3.10 and fails. Plus `test_redteam_apply_delivery_method_filter`. Resolving.

tests/redteam/test_delivery_method_registry.py
Baukebrenninkmeijer: Both fixed in 3fcb6ec.

Enum names are now **rejected**, not silently shadowed: `register_delivery_method("ROLE_PLAY")` raises `ValueError` pointing at `'role-play'`. Worth noting the ordering trap this surfaced — the value lookup has to run *before* the name check, because `DeliveryMethod.DAN` is spelled `DAN` on both sides, and a name-first check rejected the canonical value while suggesting the identical string back. A parametrized sweep over every member (`test_register_any_enum_value_is_noop`) pins that, so the next member whose name equals its value cannot reintroduce it.

Isolation: `clear_custom_delivery_methods()` is now public, and the autouse fixture moved to `tests/redteam/conftest.py` and calls it instead of poking `_CUSTOM_DELIVERY_METHODS` — so `test_cli.py::TestDeliveryMethodFlag` can no longer flip on test order. Resolving.

Baukebrenninkmeijer:
@arianpasquali — your three blocking items are done, plus three of the four non-blocking ones. Details below; CI is green across 3.10–3.13.

**Blocking**

1. **Docs** — `delivery_method_registry.py` added to the CLAUDE.md package tree (`CLAUDE.md:79`), and `docs/custom-evaluators-and-frameworks.md` gained a "Custom delivery methods" section covering `register_delivery_method`. That section also states the thing a user will actually trip on: the registry is in-memory and process-local, so registering in a standalone script does **not** carry into a separate `eq redteam run`. (9eec05a)

2. **`reports/converters.py`** — now routes through `delivery_method_str` (`converters.py:911`) instead of relying on pydantic key coercion. (9eec05a)

3. **Strictness policy** — this one is @Baukebrenninkmeijer's call and I am not making it on his behalf, so it stays open. For context, both reviewers so far land on "keep as-is": an unknown *delivery method* is a harmless filter label that either matches a dataset row or does not, whereas an unknown *vulnerability* yields no strategies and no evaluator — a silently unscoreable run. Same registry shape, different failure mode. But the ticket did say "consistently for both", so it needs an explicit OK here.

**Non-blocking, done anyway**

- **Completeness check** — went one better than a test: import-time `RuntimeError` if a `DeliveryMethod` member is missing from `DELIVERY_METHOD_CATEGORY`, then a `MappingProxyType` freeze so callers cannot mutate it at runtime. Mirrors `vulnerability_registry.py:184-188`, which is the prior art this module set out to follow. The unit test is there too (`test_delivery_method_registry.py:209`). (9dd5e75)
- **`# ponytail:` comment** — removed. (9dd5e75)
- **`resolve` / `is_known` asymmetry** — resolved as a documentation defect rather than a code change: `resolve_delivery_method` genuinely cannot return a member for a registered custom, because no member exists. The docstrings claimed otherwise; they now say an enum member resolves to its object while a registered custom stays a raw string, and that registration's whole observable effect is suppressing the CLI warning. (9dd5e75)

**Non-blocking, deliberately left**

- **4 dashboard files hand-rolling `getattr(dm, "value", dm)`** — the four you named (`dashboard/filters.py`, `redteam_charts.py`, `redteam_transcripts.py`, `converters.py`) are converted and now have assertions on the produced strings, so a revert to `str(dm)` fails on 3.10 instead of silently rendering `DeliveryMethod.DIRECT_REQUEST`. Three sites remain in the legacy Streamlit dashboard (`redteam/ui/dashboard.py:369,478,1374`). They are correct as written — `getattr(dm, "value", dm)` does the right thing on both 3.10 and 3.11+ — just duplicated, so I left them for the follow-up you suggested rather than growing this diff.
- **`register_delivery_method` / `DELIVERY_METHOD_CATEGORY` have zero callers** — `DELIVERY_METHOD_CATEGORY` is now a frozen, import-time-validated taxonomy, so I would keep it. `register_delivery_method` is a public API for library users, but its `category` argument is written and never read back by anything. @karinakalicka — your call whether a consumer reads it or the parameter goes; not merge-blocking either way.

PR #95, opened 2026-08-03
fix(simulation): send memory entity_id for agent targets with memory stores (RES-1164)

KarinaKKarinaK:
## Summary

Simulating against an orq agent with a memory store attached failed every conversation with `400: Memory entity ID is required when using memory tools` — the sim's `agent:<key>` target never sent `memory.entity_id`, and `OrqResponsesTarget` stored the id without including it in the request body.

- `OrqResponsesTarget._call_responses_api` now sends `memory: {entity_id}` when set (same shape `ORQAgentTarget` already sends).
- New `eq sim simulate --memory-entity <id>` flag; applied on the `agent:<key>` path, rejected with a clear error for deployment/vercel/openai targets.
- SDK: `simulate(memory_entity_id=...)` threads through `SimulationConfig` to `_resolve_target` for string agent targets.
- One entity id is shared across the run's conversations (the runner shares a single target).

## Testing

- 4 new tests in `tests/simulation/test_resolve_target.py` (applied on agent path, default None, rejected for deployment/instance targets).
- Full simulation suite: 650 passed. `ruff check src` and `basedpyright` clean.
- Verified live against `agent:fastned-analyst` (memory store attached): without the flag → every conversation 400s; with `--memory-entity` → conversations run to completion (terminated by max_turns, judged normally).

Linear: RES-1164


currentlycodinng (approved):
**Approve.** Clean, minimal fix for the memory-store 400, and it's the same `{'entity_id': …}` shape `ORQAgentTarget` already sends, riding the same `extra_body` channel `thread`/`pipeline` metadata use — consistent, not a new convention.

Verified live: against a real memory-store agent, without the flag every conversation fails with the exact `400 memory_entity_id_required`, and with `--memory-entity` that error is gone (the request clears the memory gate). Also checked in code: `memory_entity_id` defaults to `None` on the base `AgentTarget` so non-memory sims can't AttributeError on the new read; the "one id shared per run" claim holds because the sim path reuses the single resolved target (the fresh-uuid `.new()` is the red-team runner's path, not sim); and the reject-for-non-agent-targets guard is present on both the SDK and CLI. 12/12 target tests pass.

Nice catch on the root cause. Nothing blocking.

KarinaKKarinaK:
Manual testing results:

- Same seed datapoints against agent:fastned-analyst: without --memory-entity every conversation fails with a 400, with it all 10 run to completion and get judged normally. Checked the transcripts in the dashboard as well.
- The second commit fixes retry behavior found during that testing: 400 and 403 target errors were retried 3 times each, now they fail fast with a single log line. 408, 429 and 5xx still retry.
- Reproduction note: the fastned memory store is project-scoped, so testing needs the project key. The workspace key returns 403 Project out of scope.

Dashboard issues found in the same session were split out to https://github.com/orq-ai/evaluatorq/pull/96 (RES-1165).

Baukebrenninkmeijer:
## Luna/Codex PR review (RES-1164)

Reviewed with three parallel codex agents (correctness, tests, comments) against the branch; each ran the suite — **161 passed**. No blocking bugs: `attempt` scoping, the 4xx-break state handling, and 429/500 retry logic all verified correct.

### 🔴 Must fix

- **No regression test for the actual bug this PR fixes** — `src/evaluatorq/openresponses/target.py:143`. The core fix is "target stored `memory_entity_id` but never sent it in the body." Nothing asserts `memory: {entity_id: ...}` now lands in `extra_body` (nor that it coexists with the thread/pipeline metadata). This is the one test that would catch a regression of the exact reported 400. Add an `OrqResponsesTarget` request-payload assertion.
- **Fix doesn't cover `generate_and_simulate()` / `sim run`** — `src/evaluatorq/simulation/api.py:371`, `src/evaluatorq/simulation/cli.py:740`. `memory_entity_id` is threaded through `simulate()` / `sim simulate` only. A memory-backed agent invoked via the generate-then-simulate path still 400s. Given this surfaced while dogfooding RES-1083 (experiments-as-input), that path likely matters. Thread it through, or explicitly scope it out.

### 🟡 Suggestions (follow-up OK)

- **`status_code` extraction is narrow** — `target_call.py:157`. Only reads `exc.status_code` (openai `APIStatusError` shape). Vercel targets raise `httpx.HTTPStatusError` (`exc.response.status_code`); wrapped exceptions may hide it under `__cause__`/`.status`. Those 4xx still burn all retries. A small shared extractor (direct + `.response` + chained) generalizes the fail-fast beyond agent targets.
- **Empty `memory_entity_id` silently dropped** — `openresponses/target.py:145`. `if self.memory_entity_id:` skips `""`, so a blank id reproduces the same 400 instead of erroring. Reject blank at input validation.
- **Doc/comment nits** — bare `<key>` also parses as an agent target, but docstrings/help say only `agent:<key>` is supported (`api.py:198`, `_config.py:60`, `cli.py:477`). The 4xx comment could note "4xx non-retryable by default, except 408/429" since 409/425 are (correctly) also non-retried.

### ✅ Confirmed good
- `attempt = 0` pre-init + `attempts = attempt + 1` correct for both break and exhaust paths.
- 429/500 stay retryable, 400/403 fail fast — tests assert attempt counts correctly.
- CLI/API rejection duplication is warranted (separate error types/boundaries) — not worth abstracting.
- Comments in `target_call.py`, `openresponses/target.py`, and the test stub are accurate.

**Rating: 4/5** — correct, well-tested fix with clear comments; loses a point because the regression test for the actual bug is missing and a sibling entry-point (`generate_and_simulate`) still has the gap.

KarinaKKarinaK:
Addressed the full review in 2e8d22d, after resolving the merge conflicts with main in 37020d5.

**Merge note first, since it changes the review's context:** main meanwhile landed its own copy of the target-side fix (OrqResponsesTarget already sends `memory: {entity_id}`, and the sim backend now mints a fresh `red-team-*` id per target when none is given) plus the `previous_run` replay feature. The conflicts were the two versions of the same change plus the adjacent `previous_run`/`max_turns` params — resolved keeping both features; target.py is now byte-identical to main. What remains unique to this PR: the `--memory-entity` / `memory_entity_id` plumbing for pointing at a *specific* (e.g. seeded) entity, and the 4xx fail-fast retry fix.

**Must fix**

1. **Regression test for the actual bug** — `TestOrqResponsesTargetMemory` in `tests/simulation/test_orq_responses_target.py`: asserts `memory: {entity_id}` lands in `extra_body`, that it coexists with the thread + pipeline metadata params in the same body, and that no `extra_body` is sent when unset.
2. **`generate_and_simulate()` / `sim run` covered** — `memory_entity_id` is now a parameter on `generate_and_simulate()` (threaded into `SimulationConfig` exactly like the simulate path), and `eq sim run` gained `--memory-entity`, passed through the CLI's `_resolve_target`.

**Suggestions — all applied**

- **Status extraction generalized** — new `extract_status_code()` in `target_call.py` checks `.status_code` (openai), `.status` (aiohttp-style), `.response.status_code` (httpx, i.e. Vercel targets), and walks the explicit `__cause__` chain (depth-bounded; `bool` excluded since it subclasses `int`). The fail-fast check uses it; tests cover the httpx shape, a wrapped `__cause__` error, and an aiohttp-style 429 that still retries.
- **Blank `memory_entity_id` rejected** at both seams: `ValueError` in the SDK's `_resolve_target`, `typer.BadParameter` for a blank `--memory-entity` — before the falsy check downstream could silently drop it.
- **Doc nits**: docstrings/help/error messages now say `agent:<key>` *or bare `<key>`* everywhere the restriction is stated, and the 4xx comment spells out that all 4xx are non-retryable by default with only 408/429 as exceptions.

One test needed updating for main's new semantics rather than this review: `test_agent_target_memory_entity_id_defaults_to_none` asserted the pre-merge default; post-merge the default is a minted per-target `red-team-*` id (two resolves get distinct ids), and the docstrings now describe that.

Full suite 3435 passed; ruff + format + basedpyright clean on the touched files.

Baukebrenninkmeijer:
## Follow-up engineering review (post-fix) — RES-1164

Re-reviewed at the current head (`2e8d22d`). All 5 items from the earlier review are genuinely addressed and tested. A few gaps remain that the review didn't surface:

### 🔴 Worth fixing — `ORQAgentTarget.new()` silently drops `memory_entity_id`

`src/evaluatorq/redteam/backends/orq.py:364`

```python
def new(self) -> ORQAgentTarget:
    return ORQAgentTarget(
        agent_key=self.agent_key,
        orq_client=self.orq_client,
        model=self.model,
        timeout_ms=self._timeout_ms,
    )   # memory_entity_id NOT passed -> __init__ mints a fresh red-team-<uuid>
```

Any clone path discards the seeded id. `BareTargetBackend.create_target` (`backends/base.py:184`) does `self._target.new()` — that's the bring-your-own-target / red-team runner path.

The `--memory-entity` flag survives today **only because** `SimulationRunner` reuses one `_effective_target` instance (`runner/simulation.py:278`) and never calls `.new()`. That's coupling by accident: `.new()` exists precisely to give parallel jobs isolated state, so the day the sim runner clones a target for isolation, the seeded id silently reverts to a random one with no error. No test asserts "seeded id survives `.new()`" — because it currently doesn't.

Fix: propagate `memory_entity_id=self.memory_entity_id` in `new()` and add a regression test.

### 🟡 `extract_status_code` walks `__cause__` only, not `__context__`

`src/evaluatorq/common/target_call.py:85`

The new extractor is a real improvement, but it only follows the explicit `__cause__` chain. Implicit chaining — `raise WrappedError(...)` inside an `except` block *without* `from` — attaches the original under `__context__`, not `__cause__`. A target that wraps an httpx 400 that way still burns all 3 retries, which is exactly the fail-fast case the extractor was added for. Adding a `__context__` fallback in the same loop closes it.

### 🟡 Docs not updated

- `docs/cli-reference/simulation.md` has an exhaustive flag table — `--memory-entity` is missing from it, so the reference is now incomplete.
- `docs/guides/agent-simulation.md` covers `agent:<key>` targeting but never mentions that memory-backed agents 400 without an entity id — the exact gotcha this PR fixes. A one-line note there saves the next person the debugging that surfaced this.

(The `simulate()` / `generate_and_simulate()` docstrings *are* updated — API-level docs are fine.)

### ℹ️ Pre-existing (not this PR, noting for context)

Sim reuses one target instance across all conversations in a run (`parallelism` default 5), and `ORQAgentTarget.respond()` mutates `self._task_id` (`orq.py:262,301`). Concurrent conversations on the shared instance can race on `_task_id`/thread state. This already holds on `main` for any `agent:<key>` string target (`api.py:1339` → one `create_target`), so it's out of scope here — flagging only because the shared-memory-entity design rides on top of it.

KarinaKKarinaK:
Addressed in b7daca1e:

1. **Seeded id now survives `new()`, with one refinement over the suggested fix.** Propagating `memory_entity_id` unconditionally would have broken the other contract: an unseeded `ORQAgentTarget` auto-mints in `__init__`, and carrying that minted id through `new()` would make unseeded parallel jobs share one memory scope, exactly what the minting prevents. So the target now tracks whether the id was explicitly seeded (constructor arg, or later attribute assignment, which is how the sim layer's `--memory-entity` path seeds it) via a property setter, and `new()` propagates seeded ids while auto-minted ones still re-mint per clone. Regression tests cover all four cases: constructor-seeded survives, assignment-seeded survives, minted stays per-clone, and the `BareTargetBackend.create_target()` path you cited keeps the seeded id.
2. **`extract_status_code` follows `__context__`** when `__cause__` is unset, same loop and depth bound. Test raises the wrapper inside an except block without `from` and asserts both extraction and the fail-fast (1 attempt, not 3).
3. **Docs**: `--memory-entity` added to both flag tables in `docs/cli-reference/simulation.md` (`eq sim run` and `eq sim simulate`), and the agent-simulation guide now notes the memory-store 400 (`memory_entity_id_required`) and the flag next to the `agent:<key>` targeting section. Strict mkdocs build passes.

On the pre-existing shared-instance `_task_id` race: agreed it is out of scope here, and worth its own ticket since the shared-memory-entity design sits on top of it. Say the word and I will file it.

Full suite 3440 passed, ruff + format + basedpyright clean.

Baukebrenninkmeijer (changes requested):
Requesting changes for the `.new()` seeded-id drop (`orq.py:364`) — latent correctness trap that needs propagation + a regression test. The `__context__` extractor gap and the missing docs (`cli-reference/simulation.md` flag table + `agent-simulation.md` gotcha) should land in the same pass. Details in the follow-up review comment above.

KarinaKKarinaK:
@Baukebrenninkmeijer your changes-requested review and b7daca1e crossed paths (pushed a few minutes before it) — all three items in it are already on the branch head. Full status of every issue raised across this PR's comments:

**First review (3 Aug), all landed in 2e8d22d:**
- Regression test that `memory: {entity_id}` actually lands in `extra_body` (alone and with thread/pipeline metadata): `TestOrqResponsesTargetMemory`.
- `memory_entity_id` threaded through `generate_and_simulate()` and `eq sim run --memory-entity`.
- `extract_status_code` covering `.status_code` / `.status` / `.response.status_code` and the `__cause__` chain.
- Blank `--memory-entity` / `memory_entity_id` rejected at both seams.
- Bare-`<key>` wording and the 408/429 exception spelled out in docs and errors.

**Follow-up review (5 Aug), landed in b7daca1e:**
- Seeded `memory_entity_id` survives `new()` — via a seeded-vs-minted flag rather than unconditional propagation, so unseeded parallel jobs still get isolated minted ids. Four regression tests including the `BareTargetBackend.create_target()` path you cited.
- `extract_status_code` follows `__context__` when `__cause__` is unset, with an implicit-chaining fail-fast test.
- `--memory-entity` in both `cli-reference/simulation.md` flag tables and the memory-store 400 gotcha in `guides/agent-simulation.md`; strict mkdocs build passes.

**Pre-existing shared-target `_task_id` race:** now tracked as [RES-1195](https://linear.app/orqai/issue/RES-1195/shared-sim-target-instance-races-on-task-id-under-parallel), noting that the seeded-id-survives-`new()` fix removes the old blocker to cloning per conversation.

Suite 3440 passed, ruff + format + basedpyright clean, CI green at 20e886b; b7daca1e run in progress. Ready for re-review.

Baukebrenninkmeijer (approved):
Re-reviewed `b7daca1e`. All three follow-up items verified fixed:

- **`.new()` seeded-id** — your seeded-vs-minted flag is better than the unconditional propagation I suggested: seeded ids survive clones, auto-minted ones still re-mint so unseeded parallel jobs stay isolated. All 4 cases tested (`test_orq_target_memory_clone.py`), including the `BareTargetBackend.create_target()` path.
- **`extract_status_code`** — `__cause__ or __context__` (`target_call.py:87`), with extraction + implicit-chaining fail-fast tests.
- **Docs** — `--memory-entity` in both flag tables + the memory-store 400 gotcha in the guide.

CI green, mergeable. Please file the pre-existing `_task_id` race ticket — thanks for confirming RES-1195 already covers it. LGTM.

PR #96, opened 2026-08-03
feat(dashboard): multi-root eq dashboard and honest avg cost per job (RES-1165)

KarinaKKarinaK:
## Summary

Four fixes found while dogfooding the dashboard with real runs spread across repos (fallout from RES-1164 testing):

- **Multi-root `eq dashboard`**: accepts multiple path arguments. Runs live per repo and worktree (`.evaluatorq/runs`, `.evaluatorq/sim-runs`), so viewing red team and sim runs together previously required symlinking report files into one folder. `build_app(roots)` already took a list; only the CLI limited it to one path. Each directory argument expands to the same conventional store subdirs as before, and file arguments still print their direct report URL. `library.scan()` resolves and dedupes overlapping roots (repeated arg, relative + absolute forms, an arg inside another's expansion) so the landing rollups never double-count, and a non-existent path argument errors instead of silently becoming a parent root.
- **Honest "Avg cost / job"**: the landing tile divided total spend by all runs, including runs that record no cost. With 331 runs where one recorded cost it showed $0.0010 per job, which reads as "red teaming is basically free". It now averages over cost-bearing runs only (the same rule `metrics.py` already used for per-sim averages) and shows n/a when nothing records cost.
- **No resistance score for zero-evaluated red team runs** (behavior change): a run with `evaluated_attacks == 0` previously displayed a perfect 1.00 score; its row now shows no score, matching the attack-weighted logic the landing headline already used. Legacy reports without the field keep their rate.
- **Bar-share rounding**: nonzero shares never round to 0% and partial shares never to 100% (`<1%` / `>99%` clamps), so a 328-vs-1 split stays visibly a split.

Also in this branch per review: the redteam LLM-config follow-ups from the merged ruff refactor (`extra_kwargs` merged instead of splatted beside explicit sampling kwargs, so a user-supplied temperature no longer raises `TypeError` and silently degrades every evaluation to inconclusive).

## Testing

- New tests: multi-path CLI roots merging, root dedupe (same dir via two spellings), non-existent path rejection, costed-run counting, avg-over-costed-runs rendering, n/a case, `TestBarsRounding`, `TestZeroAttackScore`, `extra_kwargs` merge.
- Full non-integration suite: 3435 passed. `ruff check src`, `ruff format --check src`, and `basedpyright` clean.
- Verified live: served three repo roots in one dashboard (12 runs, red team + sim together) and confirmed the avg cost tile shows the real $0.3444 instead of $0.0010.

Linear: RES-1165


currentlycodinng (commented):
Read the four commits, the full diff, and the surrounding code in `library.py`, `metrics.py` and `cli.py`.

## One blocker

**Overlapping roots get counted twice, and it lands on the exact tile this PR is fixing.**

`library.scan()` (library.py:217-231) iterates roots with no dedupe, and `_iter_report_files` yields `*.json` per root independently. The `covered` set only suppresses reports already claimed by a manifest, not the same file reached through two roots. Each directory argument expands to five paths, so any of these double counts every report:

```
eq dashboard repo repo                 # same arg twice, easy with shell globs
eq dashboard . .evaluatorq/runs        # second arg is already inside the first expansion
eq dashboard ./repo /abs/path/to/repo  # relative and absolute forms of one repo
```

Result: "Jobs run" doubles, "Total spend" doubles, `costed_runs` doubles. Avg cost per job survives by luck since numerator and denominator both double, but the two tiles next to it are wrong. This is new: the five expansions of a single argument can never collide with each other, so before this PR the roots list was collision-free by construction.

Cheapest fix is a resolve-and-dedupe preserving order, and I'd put it at the top of `library.scan` rather than in `cli.py`, since `build_app(roots)` takes roots from other callers too:

```python
seen: set[Path] = set()
roots = [r for r in roots if r.resolve() not in seen and not seen.add(r.resolve())]
```

## Worth fixing before merge

**The description covers two of the four commits.** "fix(dashboard): never round nonzero bar shares" and "fix(dashboard): no resistance score for red team runs with zero evaluated" aren't in the summary, and the Testing section lists neither `TestBarsRounding` nor `TestZeroAttackScore`. The second one changes a run's displayed score from 1.00 to blank, which is a user-visible behavior change a reviewer reading "two fixes" will skim straight past. Either update the body or split those commits out.

**"Avg cost / job" no longer means what the label says.** The denominator is now cost-bearing runs, so in the motivating case it reads $0.3444 from a single run while 330 others contribute nothing. Much better than $0.0010, but a reader still takes it as typical per job. Consider renaming the tile or showing the denominator, something like `$0.3444 · 1 of 331 runs`. Related: `if run_cost > 0` means a genuinely free run (local model, cached) is excluded rather than counted as $0.00, so the average is structurally biased toward paid runs.

**A non-existent path is silently treated as a file.** `path.is_dir()` is false for a typo, so the parent becomes a root and `report_id` (which only hashes the resolved string, it never touches disk) prints a confident "Direct report URL" for a file that doesn't exist. With several arguments a typo in one is easy to miss, and it quietly widens the scan. An `elif path.is_file()` with an error branch would cover it.

## Minor

`test_eq_dashboard_accepts_multiple_paths` asserts membership only. Adding `assert len(roots) == len(set(roots))` would have caught the blocker above. Nothing mixes a file argument with a directory argument, though the rewritten loop newly supports that combination.

In `_bars`, the clamped dominant row keeps `width = max(pct, 1)` of 100, so the label reads `>99%` next to a bar drawn at full width. `min(pct, 99)` when clamping would make the two agree.

## What's good

The zero-evaluated fix mirrors the attack-weighted logic already documented in `landing()`, so the row score and the headline stat now fail the same way instead of disagreeing, and the comment correctly preserves legacy reports missing the field. `total = sum(...) or 1` keeps the div-by-zero guard. The `<1%` and `>99%` clamps are the right call for a 328-versus-1 split.

Happy to approve once the root dedupe is in; the rest are comments.

ChielToto (commented):
Findings

1. redteam/adaptive/evaluator.py:165 — extra_kwargs collides with the now-explicit temperature/max_completion_tokens. self.llm_kwargs = {**self._cfg.extra_kwargs, ...} is splatted alongside explicit temperature=/max_completion_tokens=. Pre-refactor, llm_kwargs={'temperature': ...} was the only way to set temperature and worked; now the same key in LLMCallConfig(extra_kwargs=...) raises TypeError: got multiple values for keyword argument, caught by the broad except, so every evaluation returns inconclusive (passed=None). Same pattern in evaluatorq_bridge.py:184. Defaults are safe (empty extra_kwargs); this only bites a user who routes those keys through extra_kwargs.
2. redteam/adaptive/evaluator.py:162 — the evaluator now always sends temperature, which it never did before. If a user lowers the evaluator temperature off the default 1.0 while the model is reasoning-class (the default is gpt-5-mini), OpenAI-style providers return 400 and except (APIConnectionError, APIStatusError): raise (line 186) re-raises it, aborting the dynamic run. Out-of-box (temperature=1.0) is fine, and whether the ORQ router actually rejects the value is provider-dependent, so this is a plausible risk rather than a confirmed break. The static path (evaluatorq_bridge.py:181) swallows the same error into an inconclusive result instead.

Notes

- Latent, not active: _create_openai_backend (registry.py:120) never forwards max_tokens/timeout_ms to OpenAITargetFactory, so the propagation params the PR added to the factory are currently unfed. No live path hits it (resolve_backend is only called with 'orq'), but the plumbing is incomplete. Also _create_orq_backend computes cfg (line 133) then never uses it — harmless dead variable.
- Verified correct: the config/llm_config shim (rejects both, warns, forwards), timeout_ms non-nullability (int = 90_000), the client None-fallback chain, and evaluator/attacker model wiring (config.evaluator.model / config.attacker.model).
- Process: labeling a breaking LLMConfig and red_team signature change as chore(ruff cleanup) hides it from changelog/semver tooling; feat!/refactor! would fit better. Already merged, so the two findings are best handled as a fast follow-up.

Rating: 4/5 — clean Ruff mechanics and a sound core refactor, with two edge-case rough spots in the new config surface (an extra_kwargs keyword collision and an unconditional temperature that can reject on reasoning models) plus some incomplete factory plumbing.

ChielToto:
You're right, my bad I got them mixed up. Here is the correct review.

What worked / tested

- CI green across the board: 4 Python versions, 3 SDK versions, lint, basedpyright, build smoke. Author reports 884 passing.
- New tests cover the real edge cases: multi-path root merging, costed-run counting, avg-over-costed-runs, the n/a case, <1%/>99% bar rounding, and the zero-evaluated-attacks score.
- Div-by-zero is safe in both spots: _bars uses sum(...) or 1, and avg cost guards on costed_runs.
- resistance=None on a zero-attack run is handled downstream since sim rows already emit None scores.

Issues

1. Scope creep vs the description. The summary says "Two fixes," but the diff ships a third behavioral change: _redteam_row now returns score=None when evaluated_attacks == 0 (previously 1.00 / "perfect"). It's correct and tested, but unrelated to cost/multi-path and unmentioned in the PR body. Same for the <1%/>99% bar-label rounding in view.py:197. Call these out in the description or split them, so a future bisect doesn't hunt a scoring change inside a "cost display" PR.

Nits

- metrics.py: the legacy-report reasoning (evaluated_attacks is not None) is load-bearing but only explained in a comment. A one-line note in the PR body on why legacy reports keep their rate would save the next reader a round trip.

Rating

Rating: 4/5. Correct, small, and well-tested with CI green; docked one point for bundling an undocumented scoring change under a "two fixes" description.

KarinaKKarinaK:
@ChielToto thanks, the findings hold on this branch even though the cited file names and line numbers have drifted (`evaluator.py`'s judge path and `registry.py` no longer carry the pattern). The same collision exists in eight call sites here: `orchestrator.py` (x2), `capability_classifier.py` (x2), `tool_chaining.py`, `attack_generator.py`, `objective_generator.py`, and `reports/recommendations.py` all pass explicit `temperature=`/`max_completion_tokens=` next to a `**extra_kwargs` splat, so a user routing those keys through `extra_kwargs` gets `TypeError: got multiple values` and, on the swallowed paths, silently inconclusive evaluations.

Fixed in 85cb21c9:

1. **Keyword collision (finding 1)**: new `LLMCallConfig.completion_params()` builds one merged dict with `extra_kwargs` last, so user keys override instead of raising, preserving the pre-refactor llm_kwargs-wins semantics (same rule `execute_chat_completion` already applied). All eight sites converted; `recommendations.py` keeps its historical `llm_kwargs`-on-top precedence.
2. **Reasoning-model temperature (finding 2)**: with the merge order, `extra_kwargs={'temperature': 1}` is now a working escape hatch for reasoning-class models; documented in the helper docstring. Left the field default (1.0, which reasoning models accept) unchanged rather than making it optional, since that is a contract change.
3. **Latent factory plumbing**: `_create_openai_backend` now forwards `pipeline_config.target_agent_timeout_ms` like the orq/openresponses factories. `max_tokens` stays deliberately unfed (no target-level config source exists) with a comment saying so. The `_create_orq_backend` dead `cfg` variable from your review does not exist in this repo's version.

Tests: merge-precedence units, a grep-level guard that fails if the explicit-kwargs-plus-splat pattern ever returns under `src/`, and the factory timeout forwarding. Full suite 3433 passed, ruff and basedpyright clean.

Two of your citations genuinely do not exist here (`evaluator.py`'s judge path goes through `execute_chat_completion`'s dict merge, and `OpenAITargetFactory` was removed in RES-808), which is what initially threw me. The semver note about the `chore(ruff cleanup)` labeling applies to the repo history where that PR merged.

Baukebrenninkmeijer:
Not approving yet — one blocker from @currentlycodinng's review is still open:

- **Root dedupe** (the stated approval gate): overlapping roots still double-count Jobs / Total spend / costed_runs. `cli.py` still builds `roots` with `+=`/`.append` and no resolve-and-dedupe, and `library.scan` wasn't touched. This corrupts the exact cost/jobs tiles this PR is fixing — needs resolve-and-dedupe before merge.

Non-blocking but worth clearing in the same pass:
- **Silent non-existent path**: the new loop still uses `else:` instead of `elif path.is_file()`, so a typo'd path becomes a parent root and still prints a confident Direct-report URL.
- **PR description**: still says "Two fixes" and omits the scoring + bar-rounding changes (TestBarsRounding / TestZeroAttackScore).

The redteam-config feedback from @ChielToto (extra_kwargs collision, reasoning-model temperature, timeout forwarding) is all addressed in 85cb21c9 — good. CI is green. Just the dedupe blocker + the two nits above.

KarinaKKarinaK:
All three items from your last comment addressed:

- **Root dedupe (the blocker)**: `library.scan()` now resolves and dedupes roots order-preserving before scanning — placed there rather than in `cli.py` per @currentlycodinng's note, so every `build_app(roots)` caller is covered. Tested with the same directory reached via a repeated arg and via a relative/absolute alias (one card, not two).
- **Silent non-existent path**: the CLI loop is now `is_dir()` / `elif is_file()` / error — a typo'd path raises `BadParameter` instead of becoming a parent root with a confident Direct-report URL. Tested.
- **PR description**: rewritten to cover all four changes (multi-root + dedupe, honest avg cost, the zero-evaluated score behavior change, bar-share rounding) plus the redteam `extra_kwargs` follow-up commit, with the legacy-report note @ChielToto asked for.

Full suite 3435 passed, lint/format/typecheck clean.

Baukebrenninkmeijer (commented):
## Adversarial review findings

- **P1 — zero-evaluated runs still show a perfect score in the red-team views.** The new guard only applies in `_redteam_row()` (`src/evaluatorq/dashboard/metrics.py:114-122`). `redteam_overview()` still copies `summary.resistance_rate` directly (`metrics.py:683-694`), and the detail view does the same (`dashboard/report_tabs.py:1572-1585`). A report with `evaluated_attacks: 0` and the schema/default `resistance_rate: 1.0` therefore still renders `1.00` outside the landing list. Apply the same no-score rule at the shared overview/detail data boundary and add route-level coverage.

- **P2 — legacy resistance fallback is incomplete at aggregation.** `_redteam_row()` intentionally preserves the rate when `evaluated_attacks` is absent, but `landing()` treats the missing field as zero while aggregating (`metrics.py:262-265`, `310-320`). Legacy reports can consequently lose their aggregate resistance rate even though their individual row retains it.

- **P2 — bar labels are fixed but the bar width is still rounded.** `_bars()` computes `width` from the rounded percentage (`dashboard/view.py:197-207`), so 328/329 gets `width:100%` while its label says `>99%`. The test checks labels and the tiny sliver case, but not that a partial dominant bar remains visually partial. Use the unrounded ratio for width.

- **P2 — `completion_params()` allows `extra_kwargs` to overwrite structural request fields.** The final `**self.extra_kwargs` in `contracts.py:143-159` can replace `model`, `messages`, `response_format`, or `extra_body`, not just sampling/provider options. In particular, recommendation generation could lose its required JSON response format or retry body. Reserve structural keys or reject them explicitly.

The local targeted test run could not collect because this workspace is missing the `fasthtml` dependency.

Baukebrenninkmeijer (changes requested):
Request changes: zero-evaluated red-team runs still render a false perfect score in the red-team overview and detail views. The guard currently only applies to landing rows; please apply the no-score behavior consistently and add route-level regression coverage. The other review findings are non-blocking follow-ups.

KarinaKKarinaK:
Blocker fixed plus two of the three P2s, and the merge conflict with main is resolved.

- **P1 (blocking)**: the no-score rule is now a shared `zero_evaluated_attacks()` predicate applied at all three render sites — landing rows, `redteam_overview()` rows, and the detail KPI band (which shows `n/a` with neutral status instead of a green 100%). Coverage: metrics-level (overview row `score is None`), route-level (`/?surface=redteam` renders the no-score dash, never `1.00`), and the detail render seam (`redteam_report_tabs` with a zero-evaluated report asserts `n/a` in the resistance card).
- **P2 bar width**: `_bars` now draws widths from the unrounded share, so 328/329 renders `width:99.7%` next to its `>99%` label instead of full width (tested, plus a full-share-stays-100% case).
- **P2 `completion_params`**: `extra_kwargs` can no longer replace structural request fields — `model`/`messages`/`response_format`/`extra_body` are reserved and raise `ValueError`; sampling/provider keys still pass through (both tested).
- **P2 legacy aggregation**: deferred to RES-1202 as agreed — it needs a decision between deriving legacy counts from `results` and visibly excluding legacy runs from the aggregate, so I filed it with both options laid out.

Merge conflict (the `Landing` field comments vs the pairwise cost rollup) resolved; full suite 3592 passed on the merged state, lint/format/typecheck clean.

Baukebrenninkmeijer (approved):
Reviewed the new commits. Approved.

The previous blocker is resolved:
- `library.scan()` now resolves and deduplicates overlapping roots.
- Nonexistent CLI paths now fail with `BadParameter`.
- The zero-evaluated no-score rule is applied consistently to landing rows, the red-team overview, and detail KPIs, with route-level regression coverage.
- Partial bar widths now use the unrounded share.
- Structural completion fields are protected from `extra_kwargs` overrides.

Verification: 604 focused tests passed; Ruff check/format and basedpyright passed. The legacy-report aggregate resistance behavior is explicitly deferred to RES-1202.

Baukebrenninkmeijer (approved):
Re-approving after the latest commits and verification. The prior blocker is resolved: multi-root scanning deduplicates overlapping roots, validates paths, and the zero-evaluated score rule is applied across all dashboard views. Tests and static checks are green.

PR #97, opened 2026-08-04
feat(redteam): make an empty filter list mean "match nothing" (RES-1182)

Baukebrenninkmeijer:
Rebased onto `main` now that #94 is merged.

## Problem

`categories=[]`, `vulnerabilities=[]` and `delivery_methods=[]` were read as **"no filter"** on the static path (truthiness checks) while the planner read the same input as `is None` and dropped everything. Passing an empty selection ran the **whole dataset at full cost and exited 0** — the opposite of what the caller asked for, and inconsistent between the two consumers of the same argument.

`strategy_registry.filter_strategies` already documented the correct rule ("Empty set filters out everything"); this aligns the rest of the pipeline to it.

## Semantics after this PR

Uniform across all four filters:

| Value | Meaning |
|---|---|
| `None` (default) | no filtering |
| `[]` | a real selection that matches nothing → run is empty → `RedTeamError` |
| `[...]` | filter to those values |

An empty run now raises `RedTeamError` naming the filters responsible, e.g.
`Filter selection produced zero datapoints — nothing to run. (categories=[], strategies=None, delivery_methods=None)`.

## Changes

- **`evaluatorq_bridge.py`** — 4 truthiness checks → `is not None` in `load_owasp_agentic_dataset` / `_apply_filters`.
- **`runner.py`**
  - `_run_dynamic_or_hybrid` no longer re-widens `categories=[]` back to the full category sweep (`categories or list_available_categories()`).
  - `previous_run` conflict detection and the vulnerability/category resolution chain use `is not None`.
  - `_check_filter_results` now also fires on an empty category/vulnerability selection instead of returning early, and reports the selection in the error.
- **Not changed:** `strategies` in static mode. Static dataset rows carry no strategy name, so a strategy filter is warned-and-ignored there regardless of whether it is empty. Planner-side behaviour is unchanged and already correct.

## Review follow-up (2nd commit)

Both reviewers found the same real defect and I reproduced it: the guard's category selection was re-derived inside the run functions from `resolved_categories`, which is already defaulted to the full sweep on an unfiltered run — so `None` never happened, the early return never fired, and an empty run caused by anything else got blamed on a 36-code filter the caller never set. The selection is now snapshotted once in `red_team()` from the raw arguments and threaded down.

Also: the error reports the argument name the caller actually used (no more merged `categories=` carrying vulnerability IDs); the delivery-method diagnostic reload passes `categories=None` so `categories=[]` no longer suppresses the hint it exists to produce; the dynamic path (the expensive one) now has coverage; and the `red_team()` docstring notes the empty-list semantics are SDK-only, since `_split_csv` maps blank CLI input back to `None`.

## Why `feat:` and not a breaking change

No caller who passes `None` or a non-empty list sees any difference. The only behaviour that changes is the empty-list case, which previously produced a silently wrong (and expensive) run — it had no defensible contract to break.

## Tests

- `_apply_filters` unit level: empty selection → `[]`; `None` → unfiltered.
- Static e2e: `categories=[]`, `vulnerabilities=[]`, `delivery_methods=[]` each raise `RedTeamError`.
- Full suite green; `ruff check src`, `ruff format --check src`, `basedpyright src tests` clean.

### DIFF
--- a/src/evaluatorq/redteam/frameworks/owasp/evaluatorq_bridge.py
+++ b/src/evaluatorq/redteam/frameworks/owasp/evaluatorq_bridge.py
@@ -523,10 +523,10 @@ def _load_from_file(
         raise DatasetError(msg) from e
     samples = dataset.samples
 
-    if categories:
+    if categories is not None:
         samples = _filter_by_categories(samples, categories, category_of=lambda s: s.input.category)
 
-    if delivery_methods:
+    if delivery_methods is not None:
         selected = set(delivery_methods)
         samples = [s for s in samples if s.input.delivery_method in selected]
         logger.info(
@@ -562,10 +562,10 @@ def _apply_filters(
     cap limits the already-filtered set rather than slicing first and filtering the
     remainder (which would yield fewer than ``num_samples`` rows).
     """
-    if categories:
+    if categories is not None:
         datapoints = _filter_by_categories(datapoints, categories, category_of=lambda dp: dp.inputs.get('category', ''))
 
-    if delivery_methods:
+    if delivery_methods is not None:
         selected = set(delivery_methods)
         datapoints = [dp for dp in datapoints if dp.inputs.get('delivery_method') in selected]
         logger.info(
--- a/src/evaluatorq/redteam/runner.py
+++ b/src/evaluatorq/redteam/runner.py
@@ -431,15 +431,21 @@ async def red_team(
             boundary; names that match no datapoint emit a warning. Does not
             apply to static (dataset) datapoints, which carry no strategy name —
             a warning is emitted if combined with ``static`` mode. ``None``
-            disables the filter.
+            disables the filter; an empty list selects nothing (see below).
         delivery_methods: Restrict the run to attacks whose delivery method
             overlaps the supplied selection. Applies to dynamic strategies
             (``AttackStrategy.delivery_methods``) and to static datapoints
             (``inputs['delivery_method']``). Combines with ``strategies`` using
             AND semantics. Methods matching no datapoint emit a warning. ``None``
-            disables the filter. When a supplied filter leaves zero datapoints to
-            run, ``red_team`` raises :class:`RedTeamError` rather than silently
-            running nothing.
+            disables the filter; an empty list selects nothing.
+
+            Filter semantics are uniform across ``categories``,
+            ``vulnerabilities``, ``strategies`` and ``delivery_methods``:
+            ``None`` means "no filter", and an empty list means "match nothing"
+            — never "match everything". Since an empty selection leaves zero
+            datapoints, ``red_team`` raises :class:`RedTeamError` rather than
+            silently running nothing (or, worse, running the full sweep the
+            caller meant to filter down).
         max_turns: Maximum conversation turns for multi-turn attacks. Defaults
             to 5, or — when replaying via ``previous_run`` — to the turn budget
             the replayed run used. An explicit value always wins.
@@ -542,8 +548,8 @@ async def red_team(
             for label, supplied in (
                 ('mode', Pipeline(mode) != Pipeline.DYNAMIC),
                 ('dataset', dataset is not None),
-                ('categories', bool(categories)),
-                ('vulnerabilities', bool(vulnerabilities)),
+                ('categories', categories is not None),
+                ('vulnerabilities', vulnerabilities is not None),
                 ('strategies', strategies is not None),
                 ('delivery_methods', delivery_methods is not None),
                 ('max_per_category', max_per_category is not None),
@@ -674,10 +680,10 @@ async def red_team(
         # the stored datapoints have already made.
         resolved_categories = replay.categories
         resolved_vulns = None
-    elif vulnerabilities:
+    elif vulnerabilities is not None:
         resolved_vulns = resolve_vulnerabilities(vulnerabilities)
         resolved_categories = [get_primary_category(v) for v in resolved_vulns]
-    elif categories:
+    elif categories is not None:
         # Surface unrecognized codes instead of silently swallowing the ValueError and
         # proceeding with resolved_vulns=None — that skips the evaluability gate and runs
         # a meaningless attack with no scoring.
@@ -1261,6 +1267,20 @@ def _create_job_for_target(
 # ---------------------------------------------------------------------------
 
 
+def _category_selection(
+    categories: list[str] | None,
+    vulnerabilities: list[Vulnerability] | None,
+) -> list[str] | None:
+    """Render the explicit category/vulnerability selection, or None if unset.
+
+    Distinguishes "not set" from "set to nothing": an empty selection returns
+    ``[]``, which the empty-run guard treats as a filter that was applied.
+    """
+    if categories is None and vulnerabilities is None:
+        return None
+    return sorted([*(categories or []), *(v.value for v in vulnerabilities or [])])
+
+
 def _check_filter_results(
     datapoints: list[DataPoint],
     strategy_names: set[str] | None,
@@ -1268,6 +1288,7 @@ def _check_filter_results(
     *,
     names_apply: bool = True,
     present_methods: list[str] | None = None,
+    category_selection: list[str] | None = None,
 ) -> None:
     """Surface strategy/delivery-method filter mismatches; fail on an empty run.
 
@@ -1287,8 +1308,12 @@ def _check_filter_results(
     filter, they are surfaced in the error so the user sees a spelling mismatch
     (``you asked for 'crescendo', the dataset uses 'Crescendo attack'``) rather
     than a bare "zero datapoints".
+
+    ``category_selection`` carries the explicit category/vulnerability selection
+    (``None`` when unset, ``[]`` when set to nothing) so a category filter that
+    empties the run hits the same guard as a delivery filter instead of exiting 0.
     """
-    if strategy_names is None and delivery_methods is None:
+    if strategy_names is None and delivery_methods is None and category_selection is None:
         return
 
     matched_names: set[str] = set()
@@ -1324,11 +1349,16 @@ def _check_filter_results(
             logger.warning(msg)
 
     if not datapoints:
-        names_repr = sorted(strategy_names) if strategy_names else None
-        methods_repr = sorted(delivery_method_str(m) for m in delivery_methods) if delivery_methods else None
+        # `is not None`, not truthiness: an empty selection is exactly the case
+        # that needs naming, and rendering it as None would hide the filter that
+        # emptied the run behind "no filter was set".
+        names_repr = sorted(strategy_names) if strategy_names is not None else None
+        methods_repr = (
+            sorted(delivery_method_str(m) for m in delivery_methods) if delivery_methods is not None else None
+        )
         msg = (
             'Filter selection produced zero datapoints — nothing to run. '
-            f'(strategies={names_repr}, delivery_methods={methods_repr})'
+            f'(categories={category_selection}, strategies={names_repr}, delivery_methods={methods_repr})'
         )
         if present_methods:
             msg += f' Delivery methods present in the dataset: {present_methods}.'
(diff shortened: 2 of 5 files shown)


ChielToto:
The core change is sound and well-tested. _apply_filters/_load_from_file correctly return [] for empty selections, the three static parametrize cases genuinely reach the RedTeamError("zero datapoints") guard, and both bridge unit tests assert the right thing. Two issues survived verification, both in the _category_selection wiring rather than the filter logic itself.

1. runner.py:1281 (medium): _category_selection returns a non-None full list on every default run, defeating the guard's early return

_category_selection's docstring promises "None if unset," but its call sites feed it the resolved selection, not the user's raw args. In the default case (categories=None, vulnerabilities=None), red_team sets resolved_vulns = resolve_vulnerabilities(list_available_categories()) (line 720), a non-None full list, and passes it in. So _category_selection(None, <full list>) returns sorted([...all categories...]), non-None. That skips the early return at line 1316 on runs where the user set no filter at all.

Failure: a no-filter run whose datapoint set is legitimately empty (empty user-supplied dataset, or every static row dropped by the no-evaluator filter). Pre-PR this early-returned; now it raises RedTeamError('Filter selection produced zero datapoints ... (categories=[ASI01, ASI02, ...all...], strategies=None, delivery_methods=None)'), blaming a category filter the caller never set. The "None if unset" branch is effectively dead in the main flow.

Fix: pass the user's raw categories/vulnerabilities into _category_selection, or only pass resolved_vulns when the user actually selected something.

2. runner.py:2842 (low): the "reload unfiltered" diagnostic is no longer unfiltered when categories/vulnerabilities=[]

The comment says "reload unfiltered ... to report the delivery_method values the dataset actually carries," but it calls load_owasp_agentic_dataset(dataset=dataset, categories=categories) where categories is resolved_categories, which is [] when the user passed categories=[] or vulnerabilities=[]. Under the new is not None semantics, [] filters to nothing, so present_rows=[] and present_methods=[].

Failure: caller combines categories=[] with a misspelled delivery_methods. The present_methods hint the reload exists to produce is suppressed, and the user gets the bare "zero datapoints" error the reload was added to avoid. The new single-filter tests don't cover this.

Fix: pass categories=None for the diagnostic reload.

Non-blocking notes

- The vulnerabilities=[] static test passes only because red_team collapses it to resolved_categories=[] before dispatch (_run_static has no vulnerabilities param). It validates the collapse, not static-side vuln handling; a one-line comment would prevent future confusion.
- The list[str] typing on the vulnerabilities=[] param is inert at runtime (an empty list constructs no elements).

Rating: 4/5. Correct, well-motivated fix for the headline bug (empty filter silently running the full sweep) with solid test coverage. Docked one point for the _category_selection wiring passing resolved instead of raw selections, which re-introduces a misleading-error edge and quietly regresses the delivery-method diagnostic. Both are small, localized fixes.

currentlycodinng (commented):
Reviewed against a fresh checkout of the branch; ran the affected suites (`tests/redteam/test_method_selection.py`, `tests/redteam/e2e/test_static_pipeline.py`) — 44 passed.

The core change is right: `[]` meaning "match nothing" is the only defensible reading, and it matches what `filter_strategies` already documented. One real problem with how the new guard is wired, plus a test gap.

## Blocking — unfiltered runs now trip the empty-run guard

Every `_category_selection(categories, resolved_vulns)` call site sits inside `_prepare_target` / `_run_dynamic_or_hybrid` / `_run_static`, and `red_team()` passes those functions `resolved_categories`, not the user's raw argument (`runner.py:831`, `runner.py:864`). For an unfiltered run `resolved_categories = list_available_categories()` (`runner.py:718`), so the helper can never return `None` on the normal path — its docstring's "None when unset" contract is unreachable.

Consequence: the early return at `runner.py:1316` (`if strategy_names is None and delivery_methods is None and category_selection is None`) stops firing for callers who set no filters at all. Verified on the branch:

```
unfiltered category_selection is None? False -> 36 entries
RAISED: RedTeamError Filter selection produced zero datapoints — nothing to run.
  (categories=['ASI01', 'ASI02', ... 36 codes ...], strategies=None, delivery_methods=None)
```

So a default run that yields zero datapoints for an unrelated reason (planner returned nothing, empty dataset file) now dies with a 36-code wall of text blaming a filter the caller never set. That also contradicts "No caller who passes `None` or a non-empty list sees any difference" in the description.

Suggested fix: compute the selection once in `red_team()` from the raw `categories` / `vulnerabilities` arguments and thread it down as an explicit parameter, rather than re-deriving it from the already-defaulted `resolved_categories`. If firing the guard on unfiltered empty runs is actually wanted, that is a separate (and defensible) change that needs its own message — "no datapoints were generated", not "filter selection produced zero datapoints".

## Test gap — the dynamic path fix is uncovered

The re-widening fix at `runner.py:1736` lives in `_run_dynamic_or_hybrid`, and it is the one guarding the expensive outcome (full category sweep against a live attacker model). Coverage is `_apply_filters` at unit level plus the static e2e; there is no `categories=[]` case anywhere outside `test_static_pipeline.py`. Extending the existing parametrize to `mode="dynamic"` would close it.

## Minor

- The description attributes the re-widening fix to `_run_static`; it is in `_run_dynamic_or_hybrid`.
- `_category_selection` merges category codes and vulnerability values into one list rendered as `categories=`, so a vulnerability-driven run reports something like `categories=['ASI01', 'prompt_injection']`. Mixed vocabulary in a user-facing error.
- The CLI cannot express an empty selection: `_split_csv` maps empty/blank input back to `None` (`cli.py:49-60`). Good, no default-run breakage, but the new uniform semantics apply to SDK callers only, which the `red_team()` docstring should say so nobody expects `-c ,` to mean "match nothing".

## Good

The `is not None` conversions are consistent and each carries a comment explaining why truthiness was wrong at that spot; reusing `_check_filter_results` instead of adding a parallel guard is the right call; the test docstrings record the actual failure mode rather than restating the assertion.

Baukebrenninkmeijer:
@currentlycodinng @ChielToto — you both landed on the same defect and you were right. Reproduced it before touching anything:

```
default-run selection is None? False | len = 54
```

Fixed in the second commit.

**Blocking — guard fires on unfiltered runs.** The selection was re-derived inside the run functions from `resolved_categories`, which `red_team` has already defaulted to the full sweep when nothing was filtered. So the "None when unset" branch was dead on the main path. It is now snapshotted once in `red_team()` from the raw `categories`/`vulnerabilities` arguments and threaded down as `filter_selection` — the fix you both proposed. `_category_selection` is deleted.

Not doing the other half you flagged, @currentlycodinng: firing the guard on genuinely-unfiltered empty runs is defensible but it's a different feature and needs its own message ("no datapoints were generated", not "filter selection produced zero"). Out of scope for RES-1182.

**Diagnostic reload** (@ChielToto) — now passes `categories=None`. You're right that reusing the resolved selection meant `categories=[]` emptied the very hint the reload exists to produce. I'd noticed this while writing the first commit and wrongly waved it through.

**Dynamic coverage** — added `test_dynamic_empty_filter_hard_fails_instead_of_sweeping_everything`, parametrized over `categories=[]` and `vulnerabilities=[]`.

**Guard-scope regression** — added `test_unfiltered_empty_run_does_not_blame_a_filter`. Worth noting how this one went: my first two attempts at it passed against the *buggy* code. The first asserted on `_check_filter_results` directly, but the bug was in the wiring, not the guard. The second used an empty dataset file, which never reaches the guard — the loader rejects `{"samples": []}` with a `DatasetError`. The committed version patches the loader to return no rows, and I verified it fails on the pre-fix `runner.py` and passes after.

**Minor, all applied**

- Error now reports the caller's own argument name — no more merged `categories=` list carrying vulnerability IDs.
- PR description corrected: the re-widening fix is in `_run_dynamic_or_hybrid` (line 1736), not `_run_static`. Verified.
- `red_team()` docstring notes the empty-list semantics are SDK-only, since `_split_csv` maps blank CLI input back to `None`.
- Comment on the `vulnerabilities=[]` static test recording that it pins `red_team`'s collapse to `resolved_categories=[]`, since `_run_static` has no `vulnerabilities` parameter.

`ruff check src`, `ruff format --check src`, `basedpyright src tests` clean; 3402 passed.

currentlycodinng (approved):
The blocking issue from my last pass is fixed. Approving, with one small regression the fix introduced and one design note.

Verified the fix:

- Reproduced the old defect and the new behaviour side by side. Under the old wiring a default run handed the guard a non-`None` selection (18 categories on this checkout), so the `None` branch was dead and any empty run got blamed on a filter nobody set. With the snapshot taken from the raw arguments, a default run now yields `filter_selection=None` and `_check_filter_results([], None, None, filter_selection=None)` returns without raising.
- `categories=[]` raises `Filter selection produced zero datapoints ... (categories=[], ...)` and `vulnerabilities=[]` raises the same message under `vulnerabilities=`, so the error names the argument the caller actually typed.
- The `vulnerabilities`-before-`categories` precedence in the snapshot mirrors the `elif` precedence in the resolution chain exactly, so a caller who sets both gets the filter that actually took effect reported back. I checked this specifically because reporting only one of two set filters would have been wrong; it is right.
- Threading is complete. All four `_check_filter_results` call sites pass `filter_selection`, both `_prepare_target` calls go through `**common_prepare_kwargs`, and `_category_selection` is gone with no dangling references.
- The SDK-only claim in the docstring holds: `_split_csv` ends in `return tokens or None`, and all four filters including `vulnerabilities` pass through it, so no CLI invocation can express `[]`.
- Full suite: 3402 passed, 2 skipped. `ruff check src` clean, `ruff format --check src` clean.

Minor regression, worth fixing before merge but not blocking:

`_run_static` line 2857 changed the delivery-method diagnostic reload from `categories=categories` to `categories=None`. That fixes the `categories=[]` case, but `categories` here is `resolved_categories` (passed at line 880), so for a genuinely filtered run like `categories=['ASI01']` the hint used to be scoped to the selected categories and now reports every delivery method in the whole dataset. The hint can therefore suggest a method that yields nothing within the categories the caller chose. `categories=categories or None` gets both cases: `[]` and `None` collapse to unfiltered, a real selection stays scoped.

I could not demonstrate the divergence empirically. The only local fixture, `static_e2e_dataset.json`, carries a single delivery method across all three of its categories, so both scopes return `['direct-request']`. The finding is from reading the code, and it depends on the real OWASP dataset having per-category variation in delivery methods.

Design note, take it or leave it: `filter_selection` is declared with a `= None` default on `_prepare_target`, `_run_dynamic_or_hybrid` and `_run_static`. All are internal and keyword-only, so a future call site that forgets to thread it silently disables the guard rather than failing loudly, which is the exact failure mode this commit exists to fix. Dropping the default would turn a miss into a `TypeError`. Everything is wired correctly today; this is about keeping it that way.

ChielToto:
What worked / tested

- CI fully green: typecheck + test on 3.10–3.13, lint, SDK-compat matrix, strict build. ruff and basedpyright src tests clean.
- Strong test coverage: _apply_filters unit level ([]→[], None→unfiltered), static e2e for all three empty filters, and the dynamic (expensive) path added in test_dynamic_pipeline.py.
- The 2nd commit fixes a real defect both reviewers caught: the guard used to re-derive its selection from resolved_categories (already defaulted to the full 36-category sweep on an unfiltered run), so the early return never fired and an unrelated empty run got blamed on a filter the caller never set. Now snapshotted once from raw args in red_team() and threaded down. Pinned by test_unfiltered_empty_run_does_not_blame_a_filter.
- Errors name the caller's actual argument (vulnerabilities=[], not a merged categories=), verified by TestEmptyRunGuardScope.

Issues

None blocking.

Nits

- _run_static diagnostic reload now passes categories=None (runner.py ~2857). Correct for the categories=[] case it was changed for, but it also widens the "delivery methods present in the dataset" hint for a legitimately category-filtered run: the user who filtered to one category now sees methods from every category. Minor diagnostic-quality trade-off, defensible given the comment frames the reload as dataset-wide.
- _check_filter_results no longer early-returns when only a category/vulnerability filter is set, so it now iterates all datapoints on every filtered-but-nonempty run instead of skipping. Cost is negligible (O(n) over already-materialized datapoints) and the warning loop is guarded on strategy_names/delivery_methods, so no spurious output — noting it only because the early-return contract changed.
- Comment density is high, but for truthiness-vs-is None semantics the "why" comments earn their place. Not a real concern.
- One verification I couldn't run locally (repo isn't checked out here): a repo-wide grep for any remaining if categories: / if delivery_methods: / if vulnerabilities: outside the touched files, and confirmation no default arg is []. Green CI across 4 Python versions makes a missed site unlikely, but worth a glance.

Rating

Rating: 5/5. Careful, well-reasoned fix that turns a silent expensive wrong run into a loud consistent error, addressed reviewer feedback in a follow-up commit, and is covered on both pipeline paths with green CI. Only minor diagnostic-scoping nits.

PR #102, opened 2026-08-04
fix(dashboard): label the pairwise surface in run list and empty state (RES-1184)

arianpasquali:
Fixes RES-1184.

## Bug

Post-merge review of #93 found one defect live on `main`: `app.py` imports `SURFACE_LABELS` from `dashboard/view.py`, but #93 added the `'pairwise': 'Pairwise'` entry only to a dead duplicate dict in `dashboard/surfaces.py` — nothing imports that copy (its "single source of truth" comment is wrong; the nav bar hardcodes its labels).

Result:
- `/?surface=pairwise` renders page title/h1 **"Reports"** (fallback in `app.py`) instead of **"Pairwise"**
- the empty state says *"Run a reports job to generate one."* instead of *"Run a pairwise job to generate one."*

## Fix

- Add `'pairwise': 'Pairwise'` to `view.SURFACE_LABELS` (the dict actually used).
- Delete the unused `surfaces.SURFACE_LABELS` duplicate and its misleading comment, so the duplication that caused the bug can't recur.
- Refresh stale docstrings in `shell.page()` and `filters.apply_or_all()` that still enumerated only `redteam`/`sim`.
- Add 2 regression tests: run-list page title/h1 says "Pairwise"; empty store empty-state names the pairwise surface.

## Verification

- `uv run pytest -m 'not integration'` — 3394 passed
- `uv run basedpyright src/evaluatorq/dashboard` — 0 errors


currentlycodinng (approved):
Approving. I checked this against a local checkout rather than reading the diff alone, and the fix holds up.

What I verified:

- The dead-dict claim is accurate. `grep -rn SURFACE_LABELS src tests` finds no importers of `surfaces.SURFACE_LABELS`; only `view.py` defines one and `app.py:57` imports it from there. The nav bar does hardcode its labels (`shell.py:98,104,110`), so the deletion is safe.
- The regression tests actually regress. I reverted the `'pairwise': 'Pairwise'` entry and reran: both new tests fail with "Run a reports job to generate one." Restored, both pass.
- No collateral from the new key. Adding `pairwise` changes the fallback at five other `SURFACE_LABELS.get` sites (`view.py:93,147,501,506,632`), so badges, type cells and the ⌘K search now render "Pairwise" instead of raw `pairwise`. All improvements, and nothing asserted the lowercase form.
- The reworded `apply_or_all` docstring matches the code: unknown surfaces really do fall back to `list(report_obj.results)`, and the unfiltered path goes through `filter_def.results`.
- `uv run pytest tests/dashboard -m 'not integration'` gives 552 passed, `uv run ruff check src` is clean.

Two nits, neither blocking:

- `surfaces.py:3` still opens with "Each surface (redteam, sim) gets a `SurfaceAdapter`", and line 12 says "(redteam) or a plain dict (sim entries)". Slightly odd to refresh the `shell.py` and `filters.py` docstrings while leaving the stale enumeration a few lines above the block being deleted.
- Same category elsewhere: `view.py:549` still reads "(Red Team / Agent Sim)", and `view.py:1121` and `filter_request.py:34` still say `'redteam' | 'sim'`.

One pre-existing bug I noticed while reading, not introduced here and probably worth its own ticket: `view.py:553` builds the empty state as `f'Run a {label.lower()} job'`, so the sim surface renders "Run a agent sim job to generate one." Pairwise and redteam read fine, only the vowel case breaks.

KarinaKKarinaK (approved):
Reviewed with the branch checked out locally. I verified every claim in the description instead of taking it on faith:

- app.py really does import SURFACE_LABELS from view.py (app.py:57), and nothing anywhere imports the deleted surfaces.py copy. The nav bar hardcodes its labels in shell.py's _NAV, so the deleted dict was fully dead and its single source of truth comment was wrong, as stated.
- The two symptoms are real: app.py:198 falls back to 'Reports' for the title, and runs_screen_body builds the empty state from the same lookup, producing 'Run a reports job to generate one.'
- The new apply_or_all docstring matches the code exactly: FILTERS.get miss falls back to report_obj.results, empty selections go through FilterDef.results, and the pairwise FilterDef reads run.entries.
- Both regression tests bite: I reverted the one line in view.SURFACE_LABELS and they fail, with the fix they pass.
- Reproduced the verification numbers: 3394 passed on the full suite, ruff clean, basedpyright 0 errors on the dashboard package.

Nice side effect worth knowing: the kind badges, the landing Type cells, and search labels all used SURFACE_LABELS.get(surface, surface), so pairwise runs previously showed the raw lowercase key there. They now show 'Pairwise' too.

No issues found. Small, surgical, and the cleanup removes the duplication that caused the bug in the first place.

**Rating: 5/5**

Baukebrenninkmeijer (changes requested):
## Review — mostly good, two in-scope gaps

The core fix is correct and safe: `SURFACE_LABELS` is a genuine single source of truth for the label call-sites, the deleted `surfaces.SURFACE_LABELS` was dead (no importers — `app.py` already imports from `view`), and the two regression tests pin the fix. Validated with the full suite (52 passed).

Two findings are in the PR's own scope ("label the pairwise surface in the run list"):

### 🔴 `_SURFACE_ICONS` has no `pairwise` entry → empty glyph in the landing run list
`src/evaluatorq/dashboard/view.py:129` — `_SURFACE_ICONS` only has `redteam` and `sim`. `_type_cell` (view.py:146) does `_SURFACE_ICONS.get(surface, '')`, so pairwise runs in the landing **"Recent runs"** list now show the correct **"Pairwise"** label but **no icon**, while redteam/sim rows get a glyph. That's the same run-list surface this PR is about — the label is fixed but the icon is still missing. Add a `pairwise` SVG entry.

### 🟠 Stale docstring on `runs_screen_body`
`src/evaluatorq/dashboard/view.py:549` — docstring still reads `"""Render a per-kind run-list screen (Red Team / Agent Sim)."""` but this function now handles pairwise too (its empty-state emits `"Run a pairwise job to generate one."`, exercised by the new test). Add pairwise to the docstring — the PR already updated the equivalent docstrings in `filters.py`/`shell.py`.

### ⚪ Note (non-blocking): consolidation is partial
This PR merges two `SURFACE_LABELS` dicts, but three other places still hardcode surface labels with **divergent casing**: `_KIND_COLORS` (view.py:212) and `metrics.py:287-292` use `'Red team'`/`'Agent sim'`, and `report_back_link` (view.py:581) uses `'Red team runs'`/`'Agent sim runs'`, versus `'Red Team'`/`'Agent Sim'` in `SURFACE_LABELS` and the nav. Not a regression and out of scope here, but worth a follow-up so a future surface doesn't reintroduce this exact drift.

Requesting changes for the missing pairwise icon (and the trivial docstring).

PR #103, opened 2026-08-04
fix(redteam): survive orq-ai-sdk drift and send a memory scope for agent targets (RES-1177)

Baukebrenninkmeijer:
A customer POC died in the red-team `context_retrieval` stage before a single attack was sent. Two independent defects, both fixed and verified live against a real agent on the research workspace.

## 1. RES-1177 — `Unset()` sentinel kills the run

`orq-ai-sdk` parks a **truthy** `Unset()` placeholder in unset optional fields, so `getattr(obj, name, None)` returns the sentinel instead of `None` — and `or` chains latch onto it too. It landed in a `str`-typed `AgentContext` field and the `ValidationError` ended the whole run.

Only reproduces on newer SDKs. `pyproject.toml` pins a floor (`>=4.4.7`), so CI installed 4.4.7 and stayed green while fresh installs resolved 4.12.x and crashed. Diffing the agent-retrieve model 4.4.7 → 4.12.17:

| | fields |
|---|---|
| removed | `workspace_id`, `version_hash` |
| added | `version`, `skills`, `type`, `engine` |

- `AgentContext` normalizes provider text in one `model_validator(mode="before")`: keeps `str`, stringifies `bool`/`int`, else `None`. Three field groups — strict (`key`), required-text (`system_prompt`/`instructions`, default `""`, mutually exclusive), optional-text. Unknown enum members drop to `None` rather than raising.
- `Backend.resolve_context` catches `ValidationError` **only** and degrades to a minimal context. Auth/HTTP/404 still propagate — red-teaming nothing is worse than an error.
- The four new fields are captured on both the red-team and simulation surfaces; `version` and `skills` render, `agent_type`/`engine` are record-only for now.
- CI gains an `sdk-compat` job resolving the newest patch of the last three SDK minors at run time, so the next upstream field removal is caught by us, not by a customer.

Backward compatible: every new field has a default, and `dashboard/library.py` `_validate_model` has no try/except — a required field would have broken every previously saved report.

## 2. Every agent with memory tools failed every turn

Found while doing the end-to-end verification run. `OrqResponsesTarget` held a `memory_entity_id` but never forwarded it, and `HybridAgentBackend` never minted one, so hosted agents with memory tools returned HTTP 400 `memory_entity_id_required` on every call — retries exhausted, attack reported as inconclusive. Now forwards `memory.entity_id` and mints a per-target id, matching `ORQAgentTarget`.

## Verification

- 1992 passed, 2 skipped; ruff + basedpyright clean.
- Targeted tests pass on **both** orq-ai-sdk 4.12.17 and 4.4.7.
- Live against `domain-search-assistant`: context enrichment intact on both SDKs (6 tools, 1 memory store, 1 knowledge base), and the previously-failing turn now returns a real response.

## Follow-ups (separate tickets)

- Recovering a usable workspace slug now that 4.12.x dropped `workspace_id`.
- Rendering `agent_type`/`engine`, and whether `agent_type == "a2a"` should feed `is_multi_agent` instead of the LLM classifier.
- A report-level schema/format version stamp.

/cc @arianpasquali

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,3 +56,46 @@ jobs:
 
       - name: Test (unit only)
         run: uv run pytest -m 'not integration'
+
+  resolve-sdk-matrix:
+    name: Resolve SDK matrix
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    outputs:
+      versions: ${{ steps.resolve.outputs.versions }}
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Resolve SDK versions
+        id: resolve
+        run: |
+          versions="$(python scripts/sdk_matrix.py)"
+          echo "versions=${versions}" >> "$GITHUB_OUTPUT"
+
+  sdk-compat:
+    name: SDK compatibility
+    needs: resolve-sdk-matrix
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    strategy:
+      fail-fast: false
+      matrix:
+        sdk: ${{ fromJson(needs.resolve-sdk-matrix.outputs.versions) }}
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          python-version: "3.13"
+
+      - name: Install dependencies
+        run: uv sync --all-extras --all-groups
+
+      - name: Install SDK version
+        run: uv pip install "orq-ai-sdk==${{ matrix.sdk }}"
+
+      - name: Test SDK compatibility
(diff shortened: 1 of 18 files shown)


Baukebrenninkmeijer:
@arianpasquali all checks green — ready for review. This fixes the crash you hit: context retrieval no longer dies on the `Unset()` sentinel newer `orq-ai-sdk` versions park in unset optional fields.

While verifying end-to-end I found a second, unrelated defect with the same customer impact — any hosted agent with memory tools returned HTTP 400 `memory_entity_id_required` on every turn, so every attack came back inconclusive. Fixed here too.

Verified live against `domain-search-assistant` on the research workspace, on both SDK 4.4.7 and 4.12.17.

arianpasquali (approved):
it looks like it solved the issue

PR #104, opened 2026-08-05
feat(evaluatorq): surface upload diagnostics in _send_cleaned_results (RES-823)

KarinaKKarinaK:
## Summary

Migrated from orqkit PR [#193](https://github.com/orq-ai/orqkit/pull/193) (the evaluatorq package moved to this repo). Implements RES-823: the upload boundary no longer throws away the diagnostics that explain the recurring "34 local rows, 8 in Explorer" mismatch.

- **`send_results_to_orq` returns the full `OrqResponse`** (`experiment_url`, `rows_created`, `experiment_name`, `sheet_id`) instead of just the URL string. The **gap warning** lives at this boundary: `rows_created < uploaded rows` produces exactly one WARNING line with both numbers and the URL, so both the red-team runner and plain `evaluatorq()` uploads get mismatch visibility from one implementation.
- **`_send_cleaned_results`** logs at INFO: cleaned vs raw row counts before upload, registered/uploaded counts + URL after; and persists **`uploaded_count`**, **`rows_created`**, and `experiment_url` on `RedTeamReport`, so the local run JSON alone diagnoses an Explorer sample-count gap.
- **Strip-to-zero path surfaced**: when raw rows exist but every job output is `None`, the skip now warns loudly (was DEBUG-only) and persists `uploaded_count=0`; the attempt count is recorded before the upload call so even an exception path leaves it in the JSON.
- **`RedTeamReport`**: two new optional fields, backward compatible.

## Migration notes

- Since the orqkit branch, this repo gained the `_experiment_url_out` sink in `evaluatorq.py` (simulation persists the experiment URL) and the RES-912 `resolve_results_base_url` forwarding in `_send_cleaned_results`; both are reconciled with the new `OrqResponse` return.

## Testing

- Full non-integration suite: 3431 passed. `ruff check src` and `basedpyright` clean.

Ticket: RES-823

### DIFF
--- a/src/evaluatorq/evaluatorq.py
+++ b/src/evaluatorq/evaluatorq.py
@@ -369,7 +369,7 @@ async def evaluatorq(
 
         # Upload results to Orq platform if API key is available
         if orq_api_key and _send_results:
-            experiment_url = await send_results_to_orq(
+            upload_response = await send_results_to_orq(
                 orq_api_key,
                 name,
                 description,
@@ -382,8 +382,8 @@ async def evaluatorq(
             )
             # Hand the created experiment's URL back to callers that opted in with a
             # sink list (e.g. simulation persists it on the SimulationRun report).
-            if _experiment_url_out is not None and experiment_url:
-                _experiment_url_out.append(experiment_url)
+            if _experiment_url_out is not None and upload_response is not None and upload_response.experiment_url:
+                _experiment_url_out.append(upload_response.experiment_url)
 
         # Check for pass failures and exit if any. In no-inference mode a row that has no
         # usable recorded response surfaces as a job error rather than an evaluator score,
--- a/src/evaluatorq/redteam/contracts.py
+++ b/src/evaluatorq/redteam/contracts.py
@@ -1584,9 +1584,7 @@ class RedTeamReport(BaseModel):
         'shared by every attack\'s thread_id; powers the dashboard "View all run traces" '
         'deep-link. None for older reports.',
     )
-    uploaded_count: int | None = Field(
-        default=None, description='Cleaned result rows sent to the Orq platform'
-    )
+    uploaded_count: int | None = Field(default=None, description='Cleaned result rows sent to the Orq platform')
     rows_created: int | None = Field(
         default=None,
         description='Rows the Orq platform actually registered; a value below uploaded_count explains a smaller Explorer sample count',
--- a/src/evaluatorq/redteam/contracts.py
+++ b/src/evaluatorq/redteam/contracts.py
@@ -1584,10 +1584,16 @@ class RedTeamReport(BaseModel):
         'shared by every attack\'s thread_id; powers the dashboard "View all run traces" '
         'deep-link. None for older reports.',
     )
-    uploaded_count: int | None = Field(default=None, description='Cleaned result rows sent to the Orq platform')
+    uploaded_count: int | None = Field(
+        default=None,
+        description='Cleaned result rows sent to the Orq platform. None when the upload was skipped '
+        '(no API key) or the report predates this field; 0 when every row was stripped in cleaning.',
+    )
     rows_created: int | None = Field(
         default=None,
-        description='Rows the Orq platform actually registered; a value below uploaded_count explains a smaller Explorer sample count',
+        description='Rows the Orq platform actually registered; a value below uploaded_count explains a smaller '
+        'Explorer sample count. None when the upload failed or was skipped (uploaded_count then records the '
+        'attempt) or the report predates this field — None with a set uploaded_count means "not confirmed".',
     )
 
     @field_validator('pipeline', mode='before')
--- a/src/evaluatorq/redteam/contracts.py
+++ b/src/evaluatorq/redteam/contracts.py
@@ -1584,6 +1584,13 @@ class RedTeamReport(BaseModel):
         'shared by every attack\'s thread_id; powers the dashboard "View all run traces" '
         'deep-link. None for older reports.',
     )
+    uploaded_count: int | None = Field(
+        default=None, description='Cleaned result rows sent to the Orq platform'
+    )
+    rows_created: int | None = Field(
+        default=None,
+        description='Rows the Orq platform actually registered; a value below uploaded_count explains a smaller Explorer sample count',
+    )
 
     @field_validator('pipeline', mode='before')
     @classmethod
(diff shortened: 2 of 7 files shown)


Baukebrenninkmeijer:
## Review — surface upload diagnostics (RES-823)

Ran a multi-agent review (general code, silent-failure, test-coverage). Tests green (41 pass). Return-type change (`str` → `OrqResponse`) is fully propagated, both prod callers updated, no regressions found. Solid diff. Findings ranked below.

### Blocking-ish (worth fixing before merge)

**1. Silent `response is None` return — `runner.py:255-256`**
`send_results_to_orq` returns `None` on both HTTP-error and broad-exception paths. The runner then hits a bare `return` and logs nothing. The log stream reads `"Uploading N…"` → silence, with no "upload failed, saved locally" line like the `except` branch has. Since the whole point of this PR is to stop dropping diagnostics, the most common failure (platform rejects the upload) exits wordless from the runner layer. Add a WARNING mirroring the `except` message before returning.

**2. Untested happy path — `evaluatorq.py:372-386`**
The return type went `str` → `OrqResponse`, and this second consumer now reads `upload_response.experiment_url`. No test drives a real `OrqResponse` through here — `test_evaluatorq_base_url.py` returns `None` (short-circuits the `is not None` guard), and the sim test stubs the whole layer. A regression that mis-read the attribute would pass CI. Add a test that patches `send_results_to_orq` to return an `OrqResponse` and asserts the URL lands in `_experiment_url_out`.

### Important

**3. `None`-means-failure semantics are undocumented — `contracts.py:1587-1590` + `runner.py`**
Failure paths leave `rows_created=None` while `uploaded_count=N`. The three-way signal is real and useful (`None` = not confirmed, `0` = accepted but registered nothing, `< uploaded` = partial). But `None` also means "old report predating this field" AND "no API key, upload skipped" (`runner.py:203`). A reader can't distinguish "failed" from "legacy JSON." The "a local JSON is enough to diagnose" claim needs this convention written into the field description, or a distinct sentinel on the confirmed-failure paths.

**4. Cleaning-drop gap only at INFO, never persisted — `runner.py:236`**
The loud WARNING fires only when *every* row is empty. The common case — e.g. 60 of 100 rows survive cleaning — gets buried in a routine INFO line. That 40-row drop is a legitimate Explorer-mismatch cause, upstream of the platform gap. The diagnostic chain covers `uploaded_count` vs `rows_created` but not `raw` vs `cleaned`. Consider a WARNING when `len(cleaned) < len(results)` and non-empty, and persisting the dropped-empty count so the full funnel (raw → cleaned → registered) is reconstructable from the JSON alone.

**5. Broad `except` now guards the diagnostics code — `runner.py:266-267`**
With `raise_on_error=False`, `send_results_to_orq` never raises to this handler — it swallows its own errors and returns `None`. So this `except` now only catches bugs in the new success block (`AttributeError` on `report.*`, a `TypeError` in the f-string log) and mislabels them as "Failed to upload… saved locally." There's no `exc_info=True`, so no traceback for the realistic thing it now catches. Narrow the catch, or add `exc_info=True` and reword.

### Minor
- Gap warning is only tested with `rows_created=0`; add a `3-of-5` case to pin "correct numbers."
- Exception-path test doesn't assert `experiment_url is None` (the None-return test does).
- Runner INFO line (`258-261`) partially duplicates INFO already emitted inside `send_results_to_orq`.

### Done well
- Recording `uploaded_count` pre-attempt so the exception path still persists it.
- Resolving `upload_base_url` outside the `try` (avoids mislabelling a resolution failure as an upload failure).
- Backward-compat JSON roundtrip test + mutate-before-save ordering test.
- `rows_created` is a required `int`, so the comparison can't `None`-crash.

### Two broader questions

**Anything missing a developer would expect?** The gap warning surfaces the *symptom* ("34 local, 8 in Explorer") but never the *cause* — platform-side dedup on duplicate row keys, or all-`None` outputs stripped in cleaning. Finding #4 is the big one: the cleaning funnel is where rows most often vanish and it's invisible unless every row drops. Persisting the full funnel (`raw_count, cleaned_count, rows_created`) would let the JSON answer "where did they go" without logs.

**New data worth retrieving?** `OrqResponse` also carries `sheet_id` and `manifest_id` — both now available and discarded. For this PR's exact goal, `sheet_id`/`manifest_id` are the platform handles to cross-reference a run against the Explorer/spreadsheet. Persisting `sheet_id` on `RedTeamReport` is the natural next step — it'd let the dashboard deep-link to the exact spreadsheet the mismatch lives in.

<sub>🤖 Multi-agent review via Claude Code</sub>

Baukebrenninkmeijer (changes requested):
Requesting changes — see the detailed review comment above. The two I'd treat as blocking: (1) the silent `response is None` return in `runner.py:255-256` logs nothing on the most common upload failure, undercutting the PR's whole purpose; (2) the `str` → `OrqResponse` return-type change added an untested happy path in `evaluatorq.py:372-386`. Findings #3–#5 (undocumented None semantics, cleaning-drop gap, over-broad except) are important but negotiable.

KarinaKKarinaK:
Addressed in the latest commit:

- **Blocking 1**: `response is None` now logs a WARNING ("upload was not confirmed... rows_created stays unset") instead of returning wordless, with a test asserting it fires and that `uploaded_count` stays set with `rows_created=None`.
- **Blocking 2**: new `tests/test_evaluatorq_upload_sink.py` drives a real `OrqResponse` through the `evaluatorq()` caller and asserts the URL lands in `_experiment_url_out` (plus None-response and no-URL guards).
- **#3**: the None semantics are now written into both field descriptions (None = skipped/unconfirmed/legacy, with `uploaded_count` set = "attempted, not confirmed").
- **#4**: a partial cleaning drop now warns with both counts (tested, plus a no-warning case). Persisting the full raw/cleaned/registered funnel on the report I'd leave for a follow-up ticket together with `sheet_id`/`manifest_id` persistence, per your broader-questions section.
- **#5**: the except now uses `logger.exception` (traceback kept) and says what it actually guards: bugs in the diagnostics block, not upload failures.
- Minor: added the 3-of-5 partial gap test pinning the exact numbers.

Full suite 3438 passed, lint/format/typecheck clean.

Baukebrenninkmeijer (approved):
Re-reviewed — both blockers addressed: an unconfirmed upload now logs a warning (rows_created stays None), and the str→OrqResponse happy path has a dedicated test. None semantics are documented in contracts, and the over-broad except now uses logger.exception (keeps the traceback). Partial cleaning-drop warning added; full funnel persistence + sheet_id/manifest_id deferred to a follow-up as agreed. CI green. Approving to clear my stale change-request.

PR #105, opened 2026-08-05
feat(redteam): black-box agent capability classification via probing (RES-934)

KarinaKKarinaK:
## Summary

Migrated from orqkit PR [#194](https://github.com/orq-ai/orqkit/pull/194) (the evaluatorq package moved to this repo). Implements RES-934: a standalone black-box variant of the capability classifier. `classify_agent_capabilities_blackbox(agent_target, llm_client)` infers the same `AgentCapabilities` from behaviour alone, no `AgentContext` / internal config, for third-party or opaque agents the white-box `classify_agent_capabilities()` is blind to.

- **Probes**: one group per capability area (memory write-then-recall, knowledge, code_execution, web_request, file_system, multi_agent) sent through `AgentTarget.respond()`, 7 turns total. Behavioural where tractable so a real capability leaves verifiable evidence rather than pure self-report. A single LLM judge call reads the transcript and infers all flags.
- **`classification_failed` semantics** mirror the white-box classifier: `True` only on a real coverage gap, never merely because a fully-probed agent has no capabilities.
- **`BlackboxAgentCapabilities`** subclasses `AgentCapabilities` (isinstance-compatible) to carry the `is_multi_agent` flag.
- Prompt-injection defense: agent responses are delimited + tag-escaped via `common.sanitize.delimit` before reaching the judge.

## Migration notes

- `Message.content` widened to `str | list[ContentPart]` in this repo since the orqkit branch; transcript rendering now goes through `coerce_content_text`.

## Testing

- 21 unit tests (mocked `AgentTarget` + mocked judge). Full non-integration suite: 3440 passed. `ruff check src` and `basedpyright` clean.

Ticket: RES-934


Baukebrenninkmeijer (approved):
Reviewed: CI green across all Python/SDK matrices, no outstanding review feedback. Approving.

Baukebrenninkmeijer:
## Review — black-box capability classifier (RES-934)

Ran a multi-agent review (strategy / quality / reuse / goal + silent-failure, tests, comments). **All 6 acceptance criteria are met** and out-of-scope items were correctly left alone. The issues below are about the approach's discovery power and two real bugs in the memory group — the *only* behaviorally-verified probe, and currently wrong in both directions.

### 🐞 Confirmed bugs

**1. Memory false-negative silently drops attack strategies.** In `_run_probes`, `answered_by_group['memory']` is incremented by the *write* probe. If write succeeds but the fresh-conversation **recall** probe throws, `memory` never enters `unprobed_groups`, so `classification_failed=False`. The judge sees the write turn but no recall → returns `memory=False` → a confident "no memory" verdict for a capability that was never actually tested. `strategy_registry.py:122` only goes optimistic on `classification_failed=True`, so memory strategies get silently skipped. The test at test file:251 covers the opposite (harmless) ordering; the dangerous one is untested.

**2. ORQ memory false-positive (flagship backend).** The recall probe builds a fresh `recall_convo` list but calls `.respond()` on the *same* target instance. `ORQAgentTarget.respond` forwards only the last turn and threads server-side state via `self._task_id`, which persists on the instance — so the recall is **not** isolated for ORQ agents. The server thread still holds "BANANA-42"; every ORQ agent "recalls" it whether or not a real memory store is attached → `memory=True` always. `.new()` exists precisely to mint a fresh instance and isn't used. The test double models no server-side state, so this is invisible to the suite.

### 🎯 Discovery coverage — what it misses

- **6 of 12 `AgentCapability` values are never probed** (`SHELL_ACCESS`, `DATABASE`, `EMAIL`, `MESSAGING`, `PAYMENT`, `USER_DATA`). An agent with those tools returns `capabilities={}`, `classification_failed=False` — reported *confidently* absent. Matches the ticket's 4-group scope, but unlike the memory-timing caveat this larger hole isn't documented.
- **Hallucination false-positives**: `42*57` is trivial mental math and "today's date" is often in the system prompt, so a model with no code/web access can register `True`. Text-only transcript can't separate "I executed this" from "I did it in my head."
- **Silent/guarded tool-users → false-negatives**: probes assume the agent narrates its tooling and hand it an "if you can't, say so plainly" escape hatch — exactly what a safety-tuned agent takes, scoring as bare and *reducing* coverage on the target most worth testing.

### 🔧 Smaller items

- Return type is `AgentCapabilities` but every branch returns `BlackboxAgentCapabilities` — callers can't reach `.is_multi_agent` through the declared type. Narrow the annotation.
- `is_multi_agent` lives on the subclass, but `strategy_planner.py:101` reads only `agent_context.is_multi_agent` → the field is dead today and will misfire when wired. Consider putting it on base `AgentCapabilities`.
- `llm_kwargs` param is threaded two levels deep and never read ("parity" only). Drop it.
- Module docstring says `classification_failed` is `True` "ONLY when the mechanism errored" and calls the probes "behavioural" — both contradict the code (it also fires on per-group coverage gaps; 4/6 probes are self-report). The function docstring is correct; align the module one.
- `_judge_transcript` is the 5th copy of the ~25-line structured-LLM-call block; `common/llm_call.py:execute_chat_parse` already exists (and adds reasoning-retry robustness these classifiers lack).
- `MAX_PROBE_TURNS` cap is unreachable (7<8) and `test_probe_turn_budget_is_capped` is tautological — monkeypatch the cap to force it. `APIStatusError` re-raise is also untested (only `APIConnectionError`).

### Verdict

Solid module that delivers the ticket. Ship-blocking in my read: **bug #1** (silent strategy drop) and **bug #2** (flagship false-positive). Everything else is safe to defer to follow-ups. Net: it's a useful coarse triage signal biased toward optimistic inclusion — but a negative result should not yet be read as authoritative.

<sub>🤖 Multi-agent review via Claude Code</sub>

Baukebrenninkmeijer (changes requested):
Requesting changes on two ship-blockers (details in review comment above):

1. Memory false-negative — write-ok/recall-fail yields `classification_failed=False`, silently dropping memory attack strategies for an untested capability.
2. ORQ memory false-positive — `task_id` persists on the instance, so the fresh-conversation recall isn't actually isolated for ORQAgentTarget; every ORQ agent reads as memory-capable.

Everything else in the review is safe to defer to follow-ups.

KarinaKKarinaK:
Both ship-blockers plus the cheap smaller items addressed in the latest commit:

- **Bug 1 (memory false-negative)**: memory now counts as probed only when the write AND the isolated recall both answered. A missing half — probe error or turn budget — adds memory to the coverage gap, so `classification_failed=True` and the planner stays optimistic instead of silently dropping memory strategies. The dangerous write-ok/recall-fail ordering is tested, as is the budget-skipped recall.
- **Bug 2 (ORQ false-positive)**: the recall runs on a fresh `target.new()` instance so server-side per-instance conversation state (`_task_id`) can't leak the write into the recall. One subtlety your review surfaced indirectly: the clone must inherit the parent's `memory_entity_id` (`new()` re-mints it), otherwise a write under entity A is unreachable from entity B and every real memory agent flips to a false negative. Tested both ways (different instance; same entity scope). A recall without a successful write is skipped entirely — it tests nothing.
- **Smaller items**: return annotation narrowed to `BlackboxAgentCapabilities`; unused `llm_kwargs` dropped; module docstring aligned with the real `classification_failed` semantics, the self-report nature of the non-memory probes (mental-math/system-prompt false positives, guarded-agent false negatives), and the six never-probed capability values now documented as a triage-not-authoritative caveat; the budget test now monkeypatches the cap (no longer tautological) and `APIStatusError` re-raise is tested.

Deferred to follow-ups as you suggested: behavioural probes for the six unprobed capabilities, `is_multi_agent` on base `AgentCapabilities` (needs the strategy_planner wiring decision), and the `execute_chat_parse` dedup across the five judge-call copies.

27 tests in the file (7 new), full suite 3446 passed, lint/format/typecheck clean.

Baukebrenninkmeijer (approved):
## Re-review — commit `d48fb6c9` resolves both blockers ✅

Re-ran against the fix. Verified locally: `27 passed`, ruff clean.

**Blocker 1 — memory false-negative: FIXED.** Memory now counts as probed only when the write *and* the isolated recall both answer; a missing half (`not (write_ok and recall_ok)`) adds `memory` to `unprobed_groups` → `classification_failed=True` → planner stays optimistic instead of silently dropping memory strategies. Covered by the exact dangerous ordering I flagged: `test_write_ok_recall_fail_is_a_coverage_gap` and `test_failed_write_skips_recall_and_flags_gap`.

**Blocker 2 — ORQ false-positive: FIXED.** Recall now runs on `agent_target.new()` — a fresh instance re-seeds `_task_id=None`, so the server-side thread is genuinely isolated, while the clone inherits the parent's `memory_entity_id` so a real persistent-memory write is still reachable. Correct distinction (fresh thread, same memory scope). Covered by `test_recall_runs_on_a_fresh_target_instance` + `test_recall_clone_keeps_the_parent_memory_scope`.

**Smaller items also handled:** return type narrowed to `BlackboxAgentCapabilities`; dead `llm_kwargs` dropped; module docstring corrected (real `classification_failed` semantics, self-report caveat, and the 6 unprobed capabilities now explicitly documented as "not evidence of absence").

**Deferred (all non-blocking, fine as follow-ups):** `is_multi_agent` on a subclass vs base + the `strategy_planner` read path (matters only when wired in); extracting the repeated structured-LLM-call block into `common/llm_call.py:execute_chat_parse`; `MAX_PROBE_TURNS` cap being unreachable and its tautological test; `APIStatusError` re-raise coverage; hallucination-resistant probes for code/web.

Approving. Nice, targeted fix.

PR #106, opened 2026-08-05
feat(simulation): accept Orq traces as simulation input (RES-1082)

KarinaKKarinaK:
## Summary

Migrated from orqkit PR [#192](https://github.com/orq-ai/orqkit/pull/192) (the evaluatorq package moved to this repo). Implements RES-1082: Orq production traces as a simulation input source, filling the Production traces row of the data-sources matrix in `simulation/README.md`. Two modes:

- **Direct**: `fetch_trace_conversations()` pulls recent traces from the Orq traces API; `datapoints_from_traces()` builds one datapoint per trace conversation. An LLM infers the persona and scenario from the transcript; the first message is the real user's opening message, verbatim.
- **Extension**: `extend_from_traces()` distills the fetched traffic into a distribution profile (topic mix, tones, technical levels), then generates new distribution-matched datapoints through the standard `DatapointGenerator`.

CLI: `eq sim from-traces --limit 20 --lookback-hours 24 --extend 10 --output dp.jsonl`, then `eq sim simulate --input dp.jsonl --target agent:my-agent`.

## Migration notes

- The sim datapoint type is `SimulationDatapoint` in this repo (renamed since the orqkit branch).
- README traces section folded into the data-sources matrix that landed here with the previous-run replay work; the simulate example uses this repo's `--input` flag.

## Testing

- 25 unit tests in `tests/simulation/test_traces_input.py`. Full non-integration suite: 3444 passed. `ruff check src` and `basedpyright` clean.

Ticket: RES-1082

### DIFF
--- a/src/evaluatorq/simulation/README.md
+++ b/src/evaluatorq/simulation/README.md
@@ -11,7 +11,7 @@ for real users?"*.
 
 ## What it does
 
-1. Builds **datapoints** from personas × scenarios (or takes them inline / from an Orq dataset).
+1. Builds **datapoints** from personas × scenarios (or takes them inline, from an Orq dataset, or from Orq production traces).
 2. For each datapoint, runs a turn-by-turn conversation between the user-simulator and your agent.
 3. The judge scores each run: goal achieved, criteria met, rules broken, termination reason.
 4. Returns `SimulationResult` objects and (by default) uploads an Experiment to Orq.
@@ -109,7 +109,7 @@ personas/scenarios (extends the dataset).
 | Orq dataset (`dataset_id=`) | ✅ | ⚠️ manual |
 | Previous run (`previous_run="<id>"` / `--from-run`, or export to JSONL via `eq sim generate --datapoints`) | ✅ | ⚠️ manual |
 | Orq experiment | ❌ no importer | ⚠️ manual (read run rows → seed phrases) |
-| Production traces | ❌ no importer | ⚠️ manual (read traces → seed phrases) |
+| Production traces (`datapoints_from_traces` / `eq sim from-traces`) | ✅ | ✅ `extend_from_traces()` |
 
 Legend: ✅ built-in · ⚠️ possible but manual · ❌ not supported yet.
 
@@ -139,10 +139,35 @@ evaluators. Runs saved before this shipped carry no datapoints and are rejected
 with an explanatory error, as are runs stamped with a replay format newer than
 the installed version understands.
 
-No built-in trace-to-persona extractor yet: turning raw traces (or previous
-runs) into seed phrases for `generate_personas()` / `generate_scenarios()` is a
-manual step. See the [guide](../../../docs/guides/agent-simulation.md) for the
-replay + trace-grounding walkthroughs.
+## Traces as input
+
+Production traces from Orq's observability product can seed simulations
+(requires `ORQ_API_KEY`). Two modes:
+
+- **Direct** — one datapoint per fetched trace: an LLM infers the persona and
+  scenario from the transcript; the first message is the real user's opening
+  message, verbatim.
+- **Extension** — an LLM distills the fetched traffic into a distribution
+  profile (topic mix, tones, technical levels), then generates *new*
+  distribution-matched datapoints through the standard generators.
+
+```python
+from evaluatorq.simulation import (
+    datapoints_from_traces, extend_from_traces, fetch_trace_conversations, simulate,
+)
+
+conversations = await fetch_trace_conversations(limit=20)
+datapoints = await datapoints_from_traces(conversations)          # direct
+datapoints += await extend_from_traces(conversations, num_datapoints=10)  # extension
+results = await simulate(evaluation_name="from-traces", datapoints=datapoints, target=...)
+```
+
+On the CLI:
+
+```bash
+eq sim from-traces --limit 20 --lookback-hours 24 --extend 10 --output dp.jsonl
+eq sim simulate --input dp.jsonl --target agent:my-agent
+```
 
 ## Tracing & PII
 
--- a/src/evaluatorq/simulation/__init__.py
+++ b/src/evaluatorq/simulation/__init__.py
@@ -90,6 +90,12 @@ if TYPE_CHECKING:
         apply_random_perturbation,
     )
     from evaluatorq.simulation.runner.simulation import SimulationRunner
+    from evaluatorq.simulation.traces import (
+        TraceConversation,
+        datapoints_from_traces,
+        extend_from_traces,
+        fetch_trace_conversations,
+    )
     from evaluatorq.simulation.types import (
         CommunicationStyle,
         ConversationStrategy,
@@ -245,6 +251,16 @@ _LAZY_IMPORTS: dict[str, tuple[str, str]] = {  # noqa: RUF067
         'evaluatorq.simulation.utils.prompt_builders',
         'generate_datapoint',
     ),
+    'TraceConversation': ('evaluatorq.simulation.traces', 'TraceConversation'),
+    'fetch_trace_conversations': (
+        'evaluatorq.simulation.traces',
+        'fetch_trace_conversations',
+    ),
+    'datapoints_from_traces': (
+        'evaluatorq.simulation.traces',
+        'datapoints_from_traces',
+    ),
+    'extend_from_traces': ('evaluatorq.simulation.traces', 'extend_from_traces'),
     'wrap_simulation_agent': (
         'evaluatorq.simulation.wrap_agent',
         'wrap_simulation_agent',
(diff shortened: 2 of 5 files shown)


Baukebrenninkmeijer:
## Review — validated findings

Multi-agent review (9 critics) of the trace-input feature, then hand-verified against the code. Only findings I could confirm against the actual source are below; each has a `file:line`. Split by severity. Nothing here blocks in isolation, but the top three are worth addressing before merge.

### High

**1. `datapoints_from_traces` runs LLM inference fully sequentially** — `traces.py:344-386`
The span-fetch phase is bounded-concurrent (`asyncio.Semaphore(_SPAN_FETCH_CONCURRENCY)`, `traces.py:266-281`), but the far slower per-trace persona/scenario inference is a plain `for conversation in conversations: await generate_structured(...)` loop. `eq sim from-traces --limit 200` = 200 sequential LLM round-trips, minutes of wall-clock, while the fetch it depends on finishes in seconds. `DatapointGenerator` already uses concurrency 5 for exactly this. Fix: `asyncio.gather` under a semaphore, mirroring `fetch_one`.

**2. `resp.json()` sits outside the per-trace `try/except`** — `traces.py:276`
The `try/except httpx.HTTPError` covers only lines 270-275; `spans = resp.json()` is on 276. A single 200-OK-but-malformed-body span response raises `json.JSONDecodeError` inside the `asyncio.gather` at 281 and aborts the **entire batch** — one bad trace kills all N. Since `JSONDecodeError` subclasses `ValueError`, the CLI's broad catch renders it as a generic user error, masking that it was a parse bug on one trace. Fix: move `resp.json()` into the try, catch `json.JSONDecodeError` alongside `httpx.HTTPError`, log + `return None` (same resilience the rest of `fetch_one` already has).

**3. Silent data loss: drop stat logged at INFO, invisible by default** — `traces.py:287` + `cli.py:79`
`logger.info('Fetched %d trace(s), %d with a usable conversation', ...)` is the *only* signal for how many fetched traces failed to parse. Default CLI verbosity sets level `WARNING` (`cli.py:79`), so it never prints. If an OTel schema change breaks parsing for most traces, the user runs the command and just sees `Built 2 datapoint(s) from 20 trace(s)` with zero indication 18 were silently dropped. Fix: fold the drop count into the CLI summary, or `logger.warning` when `usable < fetched`.

### Medium

**4. Malformed spans payload dropped with no log** — `traces.py:277-278`
`if not isinstance(spans, list): return None` — silent, unlike the `httpx.HTTPError` branch three lines up that logs. Schema drift (`{"data":[...]}` wrapper, error envelope) drops traces with no diagnostic. Add a `logger.warning`.

**5. CLI summary conflates extension datapoints with per-trace ones** — `cli.py:1358`
`f'Built {len(datapoints)} datapoint(s) from {num_traces} trace(s)'`. With `--extend`, `datapoints` includes synthetic distribution-matched datapoints that were never derived from any single trace. `--extend 15` on 5 fetched traces prints "Built 15 datapoint(s) from 5 trace(s)", implying a per-trace fan-out that didn't happen.

**6. Role match is case-sensitive on verbatim trace data** — `traces.py:60`, role stored raw at `traces.py:119`
`m['role'] == 'user'` against a role copied verbatim from the payload. A producer emitting `"role": "User"` makes `first_user_message` return `None`, and the whole conversation is dropped as unusable at `traces.py:286`. No test uses a non-lowercase role. Normalize role to lowercase in `_normalize_message`.

**7. `transcript()` truncates from the head** — `traces.py:64-66`
`text[:max_chars]` keeps the opening and discards the end. For persona/scenario/criteria inference the resolution and outcome (near the end) matter more than opening pleasantries — this drops exactly the wrong half of the long conversations that hit the cap. Consider tail or middle-out truncation.

**8. `criteria_met` evaluator may have nothing to score** — `traces.py:355-356`
The user message asks only for `'persona'` and `'scenario'`; `Scenario.criteria` is optional. If the model returns no criteria (plausible for short/ambiguous transcripts or weaker `--sim-model`), the default `criteria_met` evaluator silently has nothing to assess for every trace-derived datapoint. The system prompt spends five paragraphs on criteria phrasing but the schema never forces them. Tests mock a canned scenario, so this failure mode isn't exercised.

### Low / notes

- **9. `extend_from_traces` over-generates then truncates** — `traces.py:469-486`. `round(sqrt(n)) × ceil(n/personas)` grid frequently exceeds `n`; `[:num_datapoints]` discards the excess, each discarded persona/scenario having cost its own LLM call. Cheap but silent waste; no log of how much was thrown away.
- **10. `TraceConversation.messages: list[dict[str, str]]` is public but shape-unchecked** — `traces.py:55`. Re-exported from `simulation/__init__.py`; an external caller constructing with a dict missing `role`/`content` passes pydantic validation, then `KeyError`s at `traces.py:60`/`:65`. A small `TraceMessage(BaseModel)` would push the failure to construction.
- **11. Tool-call content is deliberately dropped** — `traces.py:96-97`. Intentional (avoids leaking blobs), but for tool-using agents the inferred scenario is grounded in a transcript with tool interactions removed and no note of it. Worth a README limitation line.
- **12. No CLI-level tests for `from-traces`** — `cli.py:1226-1358`. `traces.py` functions are well tested, but nothing invokes the Typer command: `--lookback-hours` → `start_date_ms` math, `--extend` wiring, the "no usable conversation" error, exit codes. A wrong-kwarg regression between CLI and `fetch_trace_conversations` would go uncaught.
- **13. `_resolve_orq_credentials` duplicates `fetch_data.py`** — `traces.py:74-79`. Env-var + trailing-slash base-URL resolution already exists in `fetch_data.py` (`_resolve_orq_base_url`). Two parallel Orq-auth implementations in one package.
- **14. PII not documented** — README. This ingests real production conversations verbatim, ships them to an LLM, and writes the real user's opening message to disk. The existing "Tracing & PII" section only covers outbound tracing, not this inbound path.

### Design question (not a confirmed defect)

**trace_id vs thread_id for multi-turn.** `_conversation_from_spans` reconstructs from a single span (`traces.py:199-207`), and the dashboard code (`dashboard/trace_links.py`) treats a multi-turn conversation as threaded across traces via `thread_id`, not one `trace_id`. For tool-heavy or multi-turn sessions the "first span yielding messages" may not be the top-level user turn. Mitigating factor: a `ChatCompletion` span's `input` usually carries the full message history, so `first_user_message` often still lands on the real opening turn. Worth confirming whether one trace = one full conversation in the target Orq setup before relying on the "verbatim opening message" claim in the README/docstring.

---
*Automated multi-agent review (pr-review-stolen + hate), findings hand-verified against source. Ratings from the four hate critics: 3× MINOR ISSUES, 1× NEEDS REWORK (the trace/thread design question above).*

Baukebrenninkmeijer (changes requested):
Requesting changes on the top 3 from my review comment above — the rest are non-blocking:

1. `resp.json()` outside the try/except (`traces.py:276`) — one malformed span body aborts the whole batch.
2. Silent data loss — drop stat at INFO, invisible at default WARNING (`traces.py:287` / `cli.py:79`).
3. Sequential LLM inference (`traces.py:344-386`) — should mirror the concurrent fetch phase.

Full details + the medium/low findings in the review comment. The trace_id-vs-thread_id item is a design question, not a blocker.

KarinaKKarinaK:
Addressed the top 3 plus the cheap mediums in the latest commit:

- **1 (resp.json outside try)**: moved inside the per-trace try, catching `json.JSONDecodeError` alongside `httpx.HTTPError`; a 200-but-malformed body now drops that trace with a warning, never the batch (tested).
- **2 (invisible drop stat)**: when usable < fetched the drop count logs at WARNING (visible at default verbosity); the all-usable case stays at INFO.
- **3 (sequential inference)**: `datapoints_from_traces` now fans out under `asyncio.Semaphore(5)`, mirroring the span-fetch phase; a test asserts it is actually concurrent, bounded at the cap, and order-preserving.
- **#4**: non-list spans payloads log a warning with the payload type before skipping (tested with an error-envelope body).
- **#5**: the CLI summary now separates direct from extension counts: "Built N datapoint(s) from M trace(s) + K distribution-matched datapoint(s) (--extend)".
- **#6**: roles are case-normalized in `_normalize_message`, so a "User" producer still yields a usable conversation (tested).

Deferred to follow-ups: tail-aware transcript truncation (#7), forcing criteria in the inference schema (#8), the low/notes items (#9-14, incl. the PII README line), and the trace_id-vs-thread_id design question, which needs your input on whether one trace = one conversation in the target setup.

Full suite 3448 passed, lint/format/typecheck clean.

Baukebrenninkmeijer (approved):
Re-reviewed — all three blockers fixed: json parse moved inside the try/except so one bad body no longer kills the batch, the sub-threshold drop now warns, and inference fans out under an asyncio.Semaphore. Mediums also in (malformed-spans log, CLI extend summary, case-insensitive roles + test). Deferred items (#7–14, trace_id/thread_id design question) agreed non-blocking. CI green. Approving to clear my stale change-request.

PR #108, opened 2026-08-05
feat(pairwise): expose a max_concurrency cap on the jury fan-out (RES-1183)

KarinaKKarinaK:
## Summary

Implements RES-1183: an optional `max_concurrency` knob on the pairwise / jury fan-out. Previously `run_jury` and `run_pairwise` fanned out with unbounded `asyncio.gather`, so a consumer bounding pairs at `concurrency=5` with a 3-judge swapped panel still issued up to ~30 concurrent LLM calls with no way to cap them.

- **One choke point**: every judge call (primaries, repetitions, swapped orderings, replacement stand-ins) flows through `_call_prediction` in `common/jury.py`, so a single shared `asyncio.Semaphore` there bounds the whole fan-out. The `gather` calls are untouched; the semaphore does the bounding.
- **`run_jury(max_concurrency=...)`**: accepts an int (run-local cap) or an existing `asyncio.Semaphore` (shared budget). `as_semaphore()` normalizes and validates (>= 1).
- **`run_pairwise(max_concurrency=...)`**: normalizes once at the top and passes the same semaphore into both orderings and the pair-level replacement pass, so they draw from one budget instead of doubling it.
- **`llm_jury_pairwise(max_concurrency=...)` / `PairwiseComparator`**: the comparator mints its semaphore lazily, bound to the running event loop, and shares it across every `compare()` call on that loop. The limit therefore bounds TOTAL in-flight judge calls across concurrently running pairs, not per pair. The lazy per-loop minting also keeps a capped comparator working when callers run one `asyncio.run` per pair; a semaphore binds to the loop that first blocks on it, and reusing it across loops would otherwise silently degrade every verdict to `inconclusive` (regression-tested).
- **Docs**: `docs/pairwise-judging.md` gained the `max_concurrency` row and a note on sharing one semaphore across `run_pairwise`/`run_jury` runs.

Additive only: default `None` keeps current behavior; single-target `llm_jury` and the redteam jury are untouched (per the RES-760 note). If the single-target jury should get the same knob, it is a two-line follow-up since `run_jury` already accepts it.

## Testing

- 13 tests in `tests/unit/test_jury_concurrency.py` with an in-flight tracker asserting peak concurrency: `run_jury` (judges x repetitions, replacement pass), `run_pairwise` (panel=3, swap, reps=2 -> 12 calls capped at 4; pair-level replacement pass sharing the budget), comparator budget shared across 4 concurrent `compare()` calls (16 calls capped at 3), the one-`asyncio.run`-per-pair regression, unbounded-sanity checks proving the tracker observes real concurrency, verdict unchanged under a cap, and `ValueError` on non-positive limits.
- Full non-integration suite: 3432 passed. `ruff check src`, `ruff format --check src`, `basedpyright`, and `mkdocs build --strict` clean.

Ticket: RES-1183

### DIFF
--- a/docs/pairwise-judging.md
+++ b/docs/pairwise-judging.md
@@ -99,6 +99,7 @@ unavailable (no second ordering to disagree with).
 | `repetitions` | `1` | How many times each judge is asked per ordering; the judge takes its own majority first. |
 | `replacement_judges` | `None` | Stand-ins for judges that fail mechanically. Promoted per pair and run in **both** orderings, so a stand-in casts a real reconciled vote. |
 | `min_successful_judges` | `1` | Minimum decisive reconciled votes, otherwise the comparison is **inconclusive**. Must not exceed the panel size. |
+| `max_concurrency` | `None` | Cap on total in-flight judge LLM calls across all concurrently running `compare()` calls (each pair fans out judges × orderings × repetitions). Unbounded when unset. |
 
 ## Reading a comparison
 
@@ -198,6 +199,10 @@ calls itself, so you can drive the swap-and-reconcile logic with your own judge.
 `reconcile_pair()` and `pairwise_consensus()` are exposed for the same reason.
 Most callers want `llm_jury_pairwise()`; reach for `run_pairwise()` when you are
 plugging in a non-LLM judge or testing the reconciliation directly.
+
+Both `run_pairwise()` and the shared `run_jury()` accept `max_concurrency` as an
+int or an existing `asyncio.Semaphore`; pass the same semaphore to several runs
+to bound their combined fan-out with one budget.
 
 ## Where to next
 
--- a/src/evaluatorq/common/jury.py
+++ b/src/evaluatorq/common/jury.py
@@ -214,11 +214,34 @@ def append_jury_summary(explanation: str | None, jury: JuryResult | None) -> str
     return f'{base} {summary}' if base else summary
 
 
+def as_semaphore(max_concurrency: int | asyncio.Semaphore | None) -> asyncio.Semaphore | None:
+    """Normalize a concurrency limit to a semaphore (or None for unbounded).
+
+    An existing semaphore passes through unchanged so several jury runs can
+    share one budget — ``run_pairwise`` relies on this to bound its two
+    concurrent orderings with a single limit.
+    """
+    if max_concurrency is None:
+        return None
+    if isinstance(max_concurrency, asyncio.Semaphore):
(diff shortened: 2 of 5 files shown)


ChielToto:
#### What worked / tested
- Correct choke-point design: the semaphore is acquired only around the leaf `judge_fn` call, not around `gather`, so there is no risk of a fan-out deadlocking on its own bound.
- Backward compatible: every new param is keyword-only with a default; `max_concurrency=None` reproduces the old unbounded behavior exactly.
- `as_semaphore` passthrough is the right call: `run_pairwise` normalizes once and hands the same semaphore to both orderings and the replacement pass, so they draw from one budget instead of minting three. Verified the nested `run_jury` calls re-run `as_semaphore` on an existing semaphore and pass it through unchanged (no double-wrapping).
- `PairwiseComparator` holds one semaphore for its lifetime, so concurrent `compare()` calls share a total in-flight budget rather than a per-pair one. Confirmed the `compare` -> `run_pairwise(max_concurrency=self._semaphore)` path.
- Tests are the strong part: an in-flight tracker records peak concurrency, with unbounded sanity checks (peak exceeds cap without a limit), cap-enforced checks, replacement coverage, a verdict-unchanged check, and `ValueError` validation. 11 tests, all four Python versions green.

#### Nits
- Event-loop binding: `PairwiseComparator._semaphore` is created in the sync `__init__`. On 3.10+ a semaphore binds to the running loop on first `await`, so reusing one comparator with a cap set across two separate `asyncio.run()` calls will raise "bound to a different event loop." The pre-cap comparator was loop-agnostic, so this is a new constraint. Common usage (one loop, gather many compares) is fine; worth a one-line docstring note or lazy per-loop creation.
- The high-level `llm_jury_pairwise` / `PairwiseComparator` accept only `int | None`, not a shared `Semaphore`, so you cannot share one budget across multiple comparators the way the lower-level `run_*` functions allow. Looks deliberate (one comparator = one budget) and documented, just flagging for anyone wanting a process-wide cap.

#### Rating
Rating: 5/5. Clean single-choke-point design, backward compatible, and genuinely well tested with concurrency peak assertions; only a minor loop-binding edge case worth a docstring note.

Baukebrenninkmeijer (approved):
Reviewed: the event-loop semaphore-binding fix (764f202b) plus the replacement-pass budget test address the flagged nit; the int-vs-shared-Semaphore point is documented as deliberate (one comparator = one budget). CI green. Approving.

PR #109, opened 2026-08-05
feat(tracing): tag LLM invocations with evaluatorq pipeline via metadata

Baukebrenninkmeijer:
## What

Sets the `evaluatorq_pipeline` label (`red_teaming` / `agent_simulation`) on the **invocation's `metadata` request property** so the resulting Orq router traces are filterable/attributable by run type. This is the documented [`metadata`](https://docs.orq.ai/reference/responses/create-response#body-metadata) property on the invocation — **not** an OTel span attribute.

## Covered invocation paths (all gated on `client_routes_through_orq`)

- `common/llm_call.py` — judge + shared core chat/parse calls (native `metadata=`)
- `openresponses/target.py` — Responses SDK target (native `metadata=`)
- `redteam/backends/openai.py` — direct `chat.completions` target (native `metadata=`)
- `redteam/contracts.py` — aux pipeline calls (`extra_body.metadata`)

ORQ agent backend (`redteam/backends/orq.py`) already passed `metadata=` — unchanged.

Both native `metadata=` and `extra_body.metadata` produce the identical top-level `metadata` body property. Off-Orq endpoints reject the unknown field, so it's only applied when the client routes through the Orq router. Caller-supplied metadata wins on key conflict.

## Validation

Live red-team run read back via `orqi` (research workspace, prod): **3/3 router traces** tagged `metadata={'evaluatorq_pipeline': 'red_teaming'}` — attack-gen, target, and evaluator all covered.

> Note: separate `product: router` traces (which carry request metadata) appear when evaluatorq's own OTel export is off; with it on, invocations fold into one OTel trace.

Regression suites green.


ChielToto:
#### What worked / tested
- The two delivery mechanisms are cleanly split by call path and don't collide: `execute_chat_completion`/`execute_chat_parse` (judge + simulation base) apply native `metadata=` via `_apply_pipeline_metadata`, while the adaptive direct `.parse()`/`.create()` sites (attack/objective gen, orchestrator, classifier) carry it through `extra_body` via `retry_extra_body`. I checked the callers; no path gets both, so no double-write of the `metadata` key.
- Three of the four paths correctly gate on `client_routes_through_orq`, so a direct-OpenAI client won't get router-only fields.
- `Any` is imported in `openai.py`, so the new `create_kwargs: dict[str, Any]` annotation is fine.
- Validated with a live red-team run read back via `orqi` (3/3 traces tagged).
- Only automated signal is the `label` check passing. No test job is visible on the PR.

#### Issues
1. No automated tests for any of the new logic. The native-vs-`extra_body` split and the routing gate are exactly the kind of thing that regresses silently, and the only validation is a one-off manual run. A couple of unit tests (gated on / off, metadata merge precedence) would lock it in.
2. `openresponses/target.py` applies `metadata` (and `thread`) without the `client_routes_through_orq` gate the other three paths use, so it contradicts the PR's stated "only applied when routing through Orq" invariant. `build_simulation_client` can point directly at OpenAI (`client.py` documents this), in which case the tag rides along regardless. It's pre-existing behavior for the `thread` extra_body, and harmless for the Responses API where `metadata` is a standard field, but the inconsistency is worth either fixing or a one-line comment explaining why this path is unconditional.

#### Nits
- The "if routes through Orq, add pipeline metadata (+ thread)" logic is now spelled three different ways (`_apply_pipeline_metadata`, the `target.py` inline block, the `openai.py` inline block). One shared helper taking the client + kwargs dict would remove the drift risk.
- `openai.py` does the `client_routes_through_orq` / `pipeline_metadata` imports inside `respond()`. Fine if it's dodging a circular import, but worth a comment if so.
- PR title reads awkwardly ("tag LLM invocations evaluatorq pipeline via metadata"); cosmetic.

#### Rating
Rating: 4/5. Focused, correct, and the two-mechanism split holds up under tracing. Docked one for zero test coverage on logic that's easy to regress, plus the ungated openresponses path.

src/evaluatorq/openresponses/target.py
```diff
@@ -136,12 +136,16 @@ async def _do_call() -> AgentResponse:
...
-            # tag the request with the evaluatorq pipeline so its trace is
-            # attributable to the run type. Both ride in the request body.
-            body_extra = {**thread_body_param(), **pipeline_metadata_param()}
-            if body_extra:
-                kwargs['extra_body'] = {**kwargs.get('extra_body', {}), **body_extra}
+            # Tag the invocation with the evaluatorq pipeline via the documented
+            # `metadata` request property so its Orq trace is filterable by run type.
+            pipeline_md = pipeline_metadata()
+            if pipeline_md:
+                kwargs['metadata'] = {**kwargs.get('metadata', {}), **pipeline_md}
```
KarinaKKarinaK: Two nits in one spot. First, the description says all covered paths are gated on `client_routes_through_orq`, but this one is not: `pipeline_metadata()` is applied unconditionally. It happens to be safe because the Responses API accepts `metadata` natively on any endpoint, but then the description should say that instead of claiming a gate. Second, the merge order here is `{**kwargs.get('metadata', {}), **pipeline_md}`, which makes the pipeline label win, the opposite of the caller-wins rule `_apply_pipeline_metadata` documents and implements. Nothing sets `kwargs['metadata']` earlier today so it is unreachable, but either flip the order to match the stated invariant or drop the dead `kwargs.get`.

src/evaluatorq/redteam/contracts.py
```diff
@@ -692,10 +692,17 @@ def retry_extra_body(self, client: 'AsyncOpenAI | None') -> dict[str, Any]:
...
 
         if not client_routes_through_orq(client):
             return {}
-        return {'retry': {'count': self.retry_count, 'on_codes': self.retry_on_codes}}
+        # Tag every pipeline-internal call (attack/objective generation, evaluator,
+        # orchestrator, capability classifier) with the active surface so its Orq
+        # trace spans are filterable as red teaming. No-op when no pipeline is bound.
+        return {
+            'retry': {'count': self.retry_count, 'on_codes': self.retry_on_codes},
+            **pipeline_metadata_param(),
```
KarinaKKarinaK, line 719: Works, and I verified no caller mixes this with the native `metadata=` path, so nothing gets double-tagged. Worth one sentence in the docstring though: callers that pass `extra_body=cfg.retry_extra_body(...)` and ALSO want their own metadata now have to merge it themselves, since this dict fully owns the `metadata` key on those calls.

KarinaKKarinaK (approved):
Reviewed with the branch checked out and every claim in the description verified against the code rather than taken on trust:

- The gating helper really checks the client's actual base_url (`/v3/router` suffix), not env vars, so the injected-OpenAI-client case is safe. Confirmed for `llm_call.py`, `backends/openai.py`, and `retry_extra_body`.
- Caller-supplied metadata wins in `_apply_pipeline_metadata`, as stated.
- `backends/orq.py` already sends `pipeline_metadata_param()` at both call sites and is untouched, as stated.
- No path double-tags: the `retry_extra_body` consumers call `create()` directly, and nothing feeds `retry_extra_body` into `execute_chat_completion`/`execute_chat_parse`, so no request carries both native `metadata=` and `extra_body.metadata`.
- Reproduced the green suite locally: 3234 passed.

Two small inaccuracies flagged inline (one description claim, one merge-order nit), neither behavior-breaking.

One ask, not blocking: none of the four new behaviors has a test (the gate, the caller-wins precedence, the native-vs-extra_body split in the target, the metadata in `retry_extra_body`). Nothing existing broke because nothing existing asserted pipeline metadata on these paths, which also means nothing will catch a regression. A handful of kwargs-capture tests like the existing `test_respond_sends_thread_param_when_conversation_active` would pin them cheaply.

Coordination heads-up: my open #95 rewrites the same `extra_body` block in `openresponses/target.py` (memory entity) and its `TestOrqResponsesTargetMemory` asserts pipeline metadata under `extra_body`, where this PR moves it to the native kwarg. Whoever lands second gets a conflict plus one test to reconcile. Happy to take that on my side if this merges first.

**Rating: 4/5**

src/evaluatorq/openresponses/target.py
Baukebrenninkmeijer: Both adopted in 69928139. The block is now gated on `client_routes_through_orq(self._client)` (target.py) — a direct-OpenAI client gets no `metadata`/`thread`/`memory`, matching the invariant. Merge order flipped to `{**pipeline_md, **kwargs.get(metadata, {})}` so caller-supplied metadata wins, consistent with `_apply_pipeline_metadata`. Added `test_pipeline_metadata_gate.py` covering both.

src/evaluatorq/redteam/contracts.py
Baukebrenninkmeijer: Documented in 69928139 — `retry_extra_body` docstring now states it fully owns the `metadata` key, so a caller wanting its own metadata on those calls must merge the dicts itself.

Baukebrenninkmeijer:
Thanks — addressed in 69928139.

**Issue 1 (no tests):** added `tests/openresponses/test_pipeline_metadata_gate.py` and four cases in `tests/common/test_llm_call.py` — routing gate on/off, no-pipeline no-op, and caller-wins merge precedence, for both `execute_chat_completion` and `OrqResponsesTarget`.

**Issue 2 (ungated openresponses path):** gated. The block now sits behind `client_routes_through_orq(self._client)`, so a direct-OpenAI client (which `build_simulation_client` can produce) gets no `metadata`/`thread`/`memory`. This also flips the metadata merge to caller-wins, which Karina flagged separately.

**Nits:**
- *Inline imports in `openai.py`* — hoisted to module level; verified no circular dep (`llm_client`/`thread_context` don't import backends).
- *Three spellings of the tag logic* — left as-is. After gating, `openai.py` and `target.py` still differ (target forwards `memory` scope + rides `thread` in extra_body; the chat path is native-only), and `_apply_pipeline_metadata` covers the shared judge/sim choke point. A single helper would have to special-case each caller's shape, so the abstraction wouldn't pay for itself.
- *PR title* — reads fine as committed ("tag LLM invocations with evaluatorq pipeline via metadata"); leaving it.

PR #110, opened 2026-08-05
fix(simulation): isolate each conversation on its own target clone (RES-1195)

KarinaKKarinaK:
## Summary

Implements RES-1195 (split out of the RES-1164 review): the simulation runner reused one resolved target instance across all conversations in a run, so at the default `parallelism=5` concurrent conversations raced any per-conversation target state — `ORQAgentTarget` threads server-side turns via `self._task_id`, and memory-backed targets own an entity id, so turns could stitch onto the wrong server-side chain and conversations could share memory.

- **Per-conversation clone**: `SimulationRunner.run()` now mints `target.new()` per conversation and threads it through the turn loop instead of the shared `_effective_target`. One seam covers both execution paths (the evaluatorq job and direct `run_batch`). The runner tracks spawned clones and `close()` releases the resources they own (an `OrqResponsesTarget` clone builds its own HTTP client when the parent's was self-owned).
- **Seeded memory survives isolation**: `OrqResponsesTarget.new()` now preserves an explicitly seeded `memory_entity_id` (constructor arg or assignment) and re-mints unseeded ones — the same seeded-vs-minted split `ORQAgentTarget.new()` got in RES-1164. Without this, cloning would have silently dropped `--memory-entity`.
- **`AgentTarget.mint_memory_entity_id()`**: the non-seeding write path. `HybridAgentBackend`'s fallback mint uses it so auto-minted ids keep re-minting per clone (preserving red-team per-attack memory isolation) while user seeds survive.

## Behavior change worth knowing

An **unseeded** memory-store agent previously had all of a run's conversations share one auto-minted entity (a side effect of the shared instance); each conversation now gets its own entity, and `--memory-entity <id>` remains the way to share one deliberately. Docs updated in the guide, the CLI flag tables, and the `simulate()` docstring.

Usage accounting is unaffected: sim token totals accumulate from each turn's `AgentResponse.token_usage` inside the runner, never from target instance state.

## Testing

- New `tests/simulation/runner/test_target_isolation.py` (5 tests): a stateful fake target with a read-yield-write `_task_id` cycle proves each of 4 parallel conversations gets its own clone with a self-consistent task chain — **this test fails on main** (verified against the pre-fix runner); seeded entity reaches every conversation; unseeded clones stay isolated; `close()` releases spawned clones (and never the caller-owned root); plain-callable targets unaffected.
- 6 new clone-semantics tests for `OrqResponsesTarget` (constructor/assignment seeds survive, grandchild clones, hybrid-backend auto-mint re-mints per clone, explicit seed after `create_target` survives).
- Two existing tests updated for the new contract (retry fake shares its counter across clones; the old always-re-mint assertion now reflects seeded-survives).
- Full non-integration suite: 3462 passed. `ruff check src`, `ruff format --check src`, `basedpyright`, and `mkdocs build --strict` clean.

Ticket: RES-1195

### DIFF
--- a/docs/cli-reference/simulation.md
+++ b/docs/cli-reference/simulation.md
@@ -25,7 +25,7 @@ Targets — provide **exactly one**:
 | Flag | Description |
 |---|---|
 | `--target` | `agent:<key>` or `deployment:<key>`. Bare values default to `agent:<key>`. |
-| `--memory-entity` | Memory `entity_id` sent with every `agent:<key>` (or bare `<key>`) target call, for agents with a memory store attached. Omit to mint a fresh id per run; pass one to reuse a specific (e.g. seeded) entity, shared across the run. |
+| `--memory-entity` | Memory `entity_id` sent with every `agent:<key>` (or bare `<key>`) target call, for agents with a memory store attached. Omit to mint a fresh id per conversation (parallel conversations never share memory); pass one to reuse a specific (e.g. seeded) entity, shared across the run. |
 | `--vercel-url` | Vercel AI SDK HTTP endpoint URL. |
 | `--openai-model` | OpenAI-compatible model name. Provider resolved from env: `ORQ_API_KEY` → Orq AI Router; `OPENAI_API_KEY` → OpenAI-compatible. |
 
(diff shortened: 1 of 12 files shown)


Baukebrenninkmeijer (commented):
Adversarial review completed against .\n\nNo actionable correctness, concurrency, lifecycle, security, or scope findings.\n\nValidated:\n- Per-conversation cloning is used by both direct  and evaluatorq job execution.\n- Seeded memory IDs survive clones; auto-minted IDs re-mint per clone.\n- Spawned clone cleanup is handled without closing caller-owned roots.\n- Changed-path tests: 76 passed.\n- All checks passed!: passed.\n- 0 errors, 0 warnings, 0 notes: 0 errors, 0 warnings, 0 notes.\n- PR CI matrix: green across Python 3.10–3.13, lint, docs, and SDK compatibility.\n\nThe broader local suite reached 2,872 passed; remaining local collection/runtime failures are due to optional  and  packages not installed in this workspace and are unrelated to this diff.

Baukebrenninkmeijer (commented):
Adversarial review completed against `origin/main...HEAD`.

No actionable correctness, concurrency, lifecycle, security, or scope findings.

Validated:
- Per-conversation cloning is used by both direct `run_batch` and evaluatorq job execution.
- Seeded memory IDs survive clones; auto-minted IDs re-mint per clone.
- Spawned clone cleanup is handled without closing caller-owned roots.
- Changed-path tests: 76 passed.
- `uv run ruff check src`: passed.
- `uv run basedpyright`: 0 errors, 0 warnings, 0 notes.
- PR CI matrix: green across Python 3.10–3.13, lint, docs, and SDK compatibility.

The broader local suite reached 2,872 passed; remaining local collection/runtime failures are due to optional `fasthtml` and `huggingface_hub` packages not installed in this workspace and are unrelated to this diff.

Baukebrenninkmeijer (approved):
Approved: no blocking findings. Changed-path tests, source lint, typecheck, and the PR CI matrix are green.

PR #111, opened 2026-08-05
feat(jury): full red-team template-var namespace in pointwise & pairwise juries (RES-1194)

Baukebrenninkmeijer:
## Summary

Make pointwise `llm_jury` and pairwise `llm_jury_pairwise` expose the full red-team evaluator template-variable namespace (`input.*`, `output.*`, `log.*`, and per-side `response_a.*`/`response_b.*`) with real structured data flowing through.

Closes RES-1194.

## What changed

- **First-class structured output:** add `AgentResponse` to the `Output` union, serialize it correctly in `JobResult`, and export it from the package root.
- **Shared output adapters:** add fail-soft `Output → text/messages` adapters and reuse them across juries, built-in checkers, the OWASP bridge, and report conversion.
- **Shared template namespace:** build pointwise and per-side pairwise namespaces from the same core, including structured input messages, output text, tool calls, messages, logs, and output errors.
- **Bare variables:** reserve bare `{{input}}`, `{{output}}`, `{{log}}`, `{{response_a}}`, and `{{response_b}}` so they remain visible literals with a warning; dotted paths remain available for structured access.
- **Pointwise jury:** migrate the default template to dotted paths and expose the full red-team namespace.
- **Pairwise jury:** add per-side namespaces and a `prompt=` override so custom prompts can access `response_a.*` and `response_b.*`; swap/reconcile logic is unchanged.
- **Errored targets:** never send errored target outputs to judges; return an inconclusive result instead while preserving `output.error` for custom templates.
- **Documentation:** add the evaluator template-variable reference, wire it into navigation, and cross-link it from the custom evaluator guide.

## Testing

- `uv run pytest -m 'not integration'`
- `uv run ruff check src`
- `uv run basedpyright src`

## Compatibility

Custom prompts using bare `{{input}}`, `{{output}}`, `{{log}}`, `{{response_a}}`, or `{{response_b}}` now receive the intact literal instead of a namespace dump. Use the documented dotted paths instead; the migration is covered in `docs/evaluator-template-variables.md`.

This is shipped as `feat:` rather than `feat!:` because the jury namespace is a recent feature, while the compatibility impact for existing bare-variable prompts is documented explicitly.

The existing red-team `build_eval_replacements` behavior is otherwise preserved, apart from the additive `output.error` key.


src/evaluatorq/common/template_engine.py
```diff
@@ -89,6 +95,8 @@ def _replacer(match: re.Match[str]) -> str:
         if not is_valid_template_path(key):
             logger.warning('Rejected template path: {!r}', key)
             return match.group(0)
+        if key in _RESERVED_BARE_KEYS:
```
KarinaKKarinaK, line 98: This branch returns the literal silently, while the invalid-path branch a few lines up logs a warning. A template that used bare {{input}} previously got a dict dump; now it gets literal braces in the rendered prompt and nothing in the logs to explain why. Add a logger.warning here pointing at dotted paths and the migration note, same pattern as the rejected-path branch. One line, and it turns a silent behavior change into a discoverable one.

KarinaKKarinaK (approved):
Reviewed with the branch checked out, and every claim in the description re-verified rather than trusted:

- Reproduced the test claim, and then some: in an env with fasthtml and huggingface_hub installed, the FULL suite including the dashboard tests you had to exclude passes: 3584 passed, 0 failures. Your 2 env failures are confirmed env gaps, not branch issues.
- ruff check and format clean, basedpyright clean on the whole repo (you claimed changed files only, it holds globally), and mkdocs build --strict passes.
- AgentResponse really is in the Output union with model_dump serialization, exported from the package root, and the moved adapters are re-exported from the bridge for back-compat.
- The output.error claim holds: no existing evaluator prompt references it, so build_eval_replacements is behavior-preserving apart from that additive key.
- Default templates and docs are fully migrated to dotted paths; the only remaining bare-var occurrences in the tree are the reserved-key list itself and the migration note.
- I also checked the pre-PR behavior to sanity-check the framing: bare {{input}} previously rendered the whole nested dict through _format, so 'hollow placeholders' is fair, and turning that into an intact literal is defensible.

One real gap, flagged inline: the reserved-bare-key branch is silent while the invalid-path branch right above it warns. A user whose custom prompt used bare {{input}} goes from a dict dump to literal braces with zero log signal. One warning line fixes it.

One wording note on the description, not the code: 'No external consumers' is too strong for a published package whose custom evaluator prompts are documented user surface. The bare-var change IS a behavior change for anyone who used them (dict dump before, literal now). The docs migration note covers it well, so just say that in the compatibility section instead of claiming nobody is affected, and the feat: label is fine.

**Rating: 4.5/5**

PR #112, opened 2026-08-06
feat(jury): wire per-judge OTEL spans + jury aggregate attributes (RES-985)

KarinaKKarinaK:
## Summary

Implements RES-985 (follow-up to RES-848): per-judge OTEL spans and jury aggregate attributes for `llm_jury()`. Before this, the only spans were the raw LLM-call spans deep inside `run_judge`; there was no jury-level span and no per-judge span carrying verdict/success/replacement, because those are jury concepts only known after aggregation.

- **`orq.jury` span** wraps `run_jury()`'s deliberation and carries `jury.verdict`, `jury.aggregator`, `jury.min_successful_judges`, `jury.raw_agreement`, `jury.judges_succeeded`, `jury.judges_configured`, `jury.judges_failed`, `jury.replacements_used`, `jury.tie`, `jury.inconclusive`.
- **`orq.judge` span** per judge (children of the jury span) carries `judge.name`, `judge.model`, `judge.verdict` (canonical A/B frame, queryable on its own), `judge.verdict_raw` (as returned, matching the chat child's output), `judge.success`, `judge.abstained`, `judge.replacement`, `judge.label_swapped` (comparative mode), `judge.error` / `judge.repetitions_failed` on failure, and `judge.latency_ms`. Token usage and cost are deliberately NOT on the judge or jury spans: they are recorded once, on the underlying `chat` spans, and rolled up by the consumer, so nothing summing across a trace double-counts. The existing `run_judge` LLM-call spans nest beneath as grandchildren: **evaluator -> jury -> judge -> llm-call**.
- **`orq.pairwise_jury` span** (comparative mode): `run_pairwise` owns one span per comparison; the inner `run_jury` calls pass `emit_span=False` and thread the comparison's context, so both orderings' judge spans (replacements included) hang off one parent. It reuses the `jury.*` namespace and adds `jury.flipped` and `jury.swap`.
- The jury span's context is threaded into each judge as `parent_context`, so parenting is deterministic regardless of how `asyncio.gather` schedules the concurrent judges.
- `judge.label_swapped` is implemented (comparative mode landed during review), not deferred.

## Behavior changes beyond the new spans

- **Token-usage attribute aliases removed.** `record_token_usage` now emits one canonical `gen_ai.usage.*` name per number. Dropped: `gen_ai.usage.prompt_tokens`, `gen_ai.usage.completion_tokens`, `gen_ai.usage.prompt_tokens_details.cached_tokens`, and the bare `prompt_tokens` / `completion_tokens` / `input_tokens` / `output_tokens` / `total_tokens` / `calls`. This reverses the keep-the-superset decision from RES-899. Verified against the platform ingest (`orquesta-web` `apps/traces-api`, `extractCommonUsage`): the generic OTel adapter tries the canonical spellings first in every pattern list, reads cache counts from the kept `cache_read.input_tokens` / `cache_creation.input_tokens` keys, and reads the bare keys and `calls` nowhere, so no dashboard field loses data. That same check showed the reasoning-token key the refactor originally kept was a spelling the platform never read; it now emits `gen_ai.usage.reasoning.output_tokens`, the key the platform consumes. Recorded in `docs/tracing.md` with the migration note for third-party OTLP consumers.
- **`single_trace`** is a new public parameter on `evaluatorq()` / `EvaluatorParams` (default off): it wraps a run's rows in one `evaluatorq.run` root span so an N-row run appears as a single trace. Release-notes-worthy; documented in `docs/tracing.md`.

## Testing

- `tests/common/test_jury_tracing.py` (in-memory span exporter): per-judge spans incl. promoted replacements, all parented to the single jury span; the enumerated `judge.*` / `jury.*` attributes; failure status and propagated-error identity; inconclusive jury; full usage rollup across repetitions; no-op when tracing is disabled. Pairwise span hierarchy and `single_trace` covered in `tests/unit/`.
- Full non-integration suite: 3876 passed. `ruff check src`, `ruff format --check src`, `basedpyright`, and `mkdocs build --strict` clean.

Ticket: RES-985


KarinaKKarinaK:
### Successful manual check:
Screenshot 2026-08-06 at 9 07 32 AM

Baukebrenninkmeijer:
## Review status update

Merged `origin/main` (and the branch's remote head) into this branch, and applied the surviving findings from the review pass in `dc45acb`.

### Applied

| Finding | Fix |
|---|---|
| Judge spans dropped `calls` / cached / reasoning token usage | `_record_judge_span` now forwards all of them from the summed `TokenUsage` |
| Failed judges lack attributes when `propagate_errors=True` | Judge identity (`judge.name` / `judge.model` / `judge.replacement`) is stamped *before* the judge runs, so an abort still leaves a span naming the dead judge |
| Swallowed failures marked `judge.success=False` but span status stayed `OK` | New `set_span_error()` helper marks the span `ERROR`; `with_span` no longer clobbers a deliberate `ERROR` with `OK` on clean exit |
| Failure detail missing | `judge.error` + `judge.repetitions_failed` stamped on the judge span |
| Jury span omits failure/replacement aggregates | Added `jury.judges_failed` + `jury.replacements_used` |

Four new tests in `tests/common/test_jury_tracing.py` cover failure status, propagated-error identity, full usage rollup across repetitions, and the jury failure aggregates.

### Rejected

- **"One `llm.judge` span for all repetitions is wrong cardinality."** The ticket asks for *per-judge* spans; repetitions are child LLM-call spans nested beneath. Intentional and documented in the docstring.
- **"Token usage is double-counted on the judge span and its child LLM spans."** That is standard parent-aggregation, not a defect.
- **"`with_span` / context capture duplicate existing helpers; `judge.name` duplicates `judge.model`."** `common/tracing.py` is the canonical home for these (per the module docstring), and the duplicate key is a deliberate seam for a future named-judge concept, commented as such.

### Checks

`uv run pytest -m 'not integration'` — 3699 passed, 2 skipped. `ruff check src` and `ruff format --check src` clean.

Baukebrenninkmeijer:
## Review status update #2 — span hierarchy completed

Follow-up to the previous comment. Merged `origin/main` again (picks up the jury errored-target short-circuit and the RES-1194 `PairwiseJudgeFn` widening) and landed `af7262a7`.

The shipped tree is now the full chain the ticket describes:

```
orq.evaluation {evaluator}
└── llm.jury
    └── llm.judge
        └── chat {model}
```

### What changed

**One jury span per comparison (comparative mode).** `run_pairwise` previously called `run_jury` once per label ordering, plus once more for the replacement pass — so a single comparison emitted 2–3 sibling `llm.jury` spans, each carrying `jury.*` aggregates that described *half* a comparison. It now owns one `llm.jury` span; the inner calls pass `emit_span=False` and thread the comparison's context, so every judge span (both orderings, replacements included) hangs off it.

**`judge.label_swapped` is wired.** The ticket deferred it until comparative mode landed. It has, so the two votes one judge casts per comparison are now distinguishable in the trace.

**The jury span records its outcome.** It previously showed agreement/tie/inconclusive but not what the panel actually decided or under which rule. Added `jury.verdict`, `jury.aggregator`, `jury.min_successful_judges`, plus rolled-up token usage and `jury.cost`. The pairwise span reuses the same `jury.*` namespace so one dashboard reads both modes, and adds `jury.flipped` (judges that contradicted themselves across orderings — position bias) and `jury.swap`.

**Evaluator parenting verified, not added.** The `orq.evaluation {name}` span already existed in `process_evaluator`; the jury span nests under it via ambient OTel context. What was missing was proof, so there's now a test asserting the whole chain end to end.

### On cost

`judge.cost` / `jury.cost` ride only when the provider reports cost (`TokenUsage` picks it off the `cost` / `total_cost` usage keys). There is no other standard field to extract — OTel GenAI semconv defines no cost attribute at all — so providers that omit it (raw OpenAI) would need a local pricing table, which the ticket explicitly punts. Token counts including cached and reasoning are complete either way, so cost stays derivable downstream.

### Checks

`uv run pytest -m 'not integration'` — 3754 passed, 2 skipped. `ruff check src` and `ruff format --check src` clean. `basedpyright` on `common/jury.py` and `pairwise.py`: 0 errors.

currentlycodinng (changes requested):
The span work itself is very good. I ran six `run_pairwise` calls concurrently against an in-memory exporter and got exactly 6 `orq.pairwise_jury` spans with 6 `orq.judge` children each, every judge parented to its own comparison and none to a sibling's. Threading the captured context explicitly rather than relying on ambient is the right call and it demonstrably holds under `gather`. The `is not None` fix on `parent_context` is a real bug caught (`Context` subclasses `dict`, so an empty root context was being silently swapped for ambient), splitting `_run_jury_core` out so comparative mode does not mint a second jury span per ordering is the right shape, and `docs/tracing.md` is the most complete tracing documentation this repo has had. The reasoning about why `jury.judges_failed` excludes flipped judges, and why `judge.verdict_frame` has to exist at all, is exactly the kind of thing that would otherwise be relearned painfully.

One thing needs to come out or be confirmed before this lands.

## This silently removes eleven span attributes that a previous PR deliberately kept

`record_token_usage` loses `gen_ai.usage.prompt_tokens`, `gen_ai.usage.completion_tokens`, `gen_ai.usage.prompt_tokens_details.cached_tokens`, and the bare `prompt_tokens`, `completion_tokens`, `input_tokens`, `output_tokens`, `total_tokens`, `calls`. Two of those deletions take a comment with them that says why they were there:

```python
# Bare keys (platform compat)
# Legacy attribute name emitted by the former redteam impl — kept for platform dashboard compat.
```

Those lines came from RES-899, where the redteam and simulation tracing implementations were unified and someone specifically decided to keep the superset rather than pick a winner. This PR reverses that decision, and nothing says so: not the title, not the description (which says "plumbing only ... no behavior change"), not `docs/tracing.md`, which documents the new attribute tables without noting that anything went away. The nine test edits that delete these assertions are the only record.

I agree with the reasoning in the new docstring. Emitting one number under three names is genuinely worse, and if this were a greenfield choice I would make the same one. My problem is that the consumer is the Orq platform dashboard, which lives outside this repo, so nothing in CI can tell you whether a panel goes blank when `total_tokens` stops arriving. Green tests here prove only that this repo no longer asserts them.

What I would like, in order of preference: pull the removal into its own PR so it can be announced and rolled back independently of the jury spans, or if it stays, get confirmation from whoever owns the platform dashboards that none of the nine keys is read, and put a line in `docs/tracing.md` recording that they were dropped and when. Either way the description needs to stop saying no behavior change, because for anyone consuming these spans this is the most consequential thing in the PR.

## The description no longer matches the code

Worth fixing before merge since the ticket will be read from it. The description says the spans are `llm.jury` and `llm.judge`; the code emits `orq.jury` and `orq.judge`, plus a third span, `orq.pairwise_jury`, that the description never mentions. It says the judge span carries `judge.verdict` and token usage "plus `judge.cost` when the summed usage carries `cost_usd`"; the code emits `judge.verdict_raw` and `_record_judge_span` explicitly records neither usage nor cost, with a docstring explaining that they belong on the `chat` spans. It says `judge.label_swapped` is deferred; it is implemented and documented. The docs are right on every one of these, so it is only the description that drifted, but the gap is wide enough that a reviewer trusting it would be checking for the wrong attributes.

Relatedly, `single_trace` is a new public parameter on `evaluatorq()` and `EvaluatorParams`, with a new `evaluatorq.run` span and a change to what an N-row run looks like in a trace viewer. It is off by default and it is well documented, but it is a user-facing API addition in a PR whose description does not mention it at all, and it is the kind of thing that belongs in release notes.

## Two nits

`record_jury_span`'s docstring says "using the same `jury.*` vocabulary and the same `jury.*` vocabulary", which reads like a bad merge.

`_judge_vote` starts its `time.monotonic()` clock before entering `with_span`, so `judge.latency_ms` includes span creation. That is microseconds and does not matter, but the attribute is documented as "wall-clock time for this judge's repetitions" and it is measuring slightly more than that. Moving `start` inside the `async with` would make it match.

---

Requesting changes only for the attribute removal, and only until it is either split out or confirmed safe with the platform. Everything else here is ready, and the parenting behaviour I was most suspicious of held up under testing.

What I checked: checked out `8c965a2`, `uv sync --all-extras`, `pytest tests/common tests/unit tests/simulation tests/redteam` 3099 passed 2 skipped, `ruff check src` and `ruff format --check src` clean, `basedpyright` 0 errors. Full CI is green on this PR, which is not true of the other open evaluatorq branches. The concurrent-parenting result above came from running six comparisons through a real in-memory span exporter, and the removed-attribute list came from diffing `common/tracing.py` against the current tip of main, which is also the merge base, so all of it is this branch's work.

KarinaKKarinaK:
## Review addressed (7ca760d1)

**The attribute removal is confirmed safe against the platform, with the evidence committed.** I checked the actual consumer rather than asking around: `orquesta-web` `apps/traces-api` at today's main (`e485ee9d5f`). Evaluatorq spans (tracer scope `evaluatorq`, no framework marker attributes) fall through the adapter registry to the generic adapter, whose usage extraction is `extractCommonUsage` in `utils/adapter-patterns.ts`. For each of the nine dropped keys:

- `gen_ai.usage.prompt_tokens`, bare `prompt_tokens` / `input_tokens`: fallbacks in the `promptTokens` pattern list, behind `gen_ai.usage.input_tokens`, which we keep and which wins first.
- `gen_ai.usage.completion_tokens`, bare `completion_tokens` / `output_tokens`: same shape, `gen_ai.usage.output_tokens` first.
- bare `total_tokens`: behind the kept `gen_ai.usage.total_tokens`.
- `gen_ai.usage.prompt_tokens_details.cached_tokens`: the generic path reads cache counts from `gen_ai.usage.cache_read.input_tokens` / `cache_creation.input_tokens`, both kept. The flat spelling is only read by the logfire and langchain adapters, which never see our spans.
- bare `calls`: grep of `apps/traces-api/src` finds no read of `calls` or `gen_ai.usage.calls` anywhere.

So no platform panel loses data. `docs/tracing.md` now records what was dropped, when, the verification, and the migration for third-party OTLP consumers, and the PR description declares it as a behavior change (the "plumbing only" claim is gone).

The same check surfaced one thing the refactor got wrong, now fixed: the canonical key it kept for reasoning tokens (`gen_ai.usage.completion_tokens_details.reasoning_tokens`) is a spelling the platform never read on the generic path. It now emits `gen_ai.usage.reasoning.output_tokens`, the key `extractCommonUsage` consumes, so reasoning counts reach the dashboard for the first time.

**Description drift: rewritten.** Span names are now `orq.jury` / `orq.judge` / `orq.pairwise_jury`, `judge.verdict_raw` + `judge.verdict_frame` as the code has them, usage/cost correctly described as living on the `chat` spans only, `judge.label_swapped` listed as implemented, and `single_trace` declared as a new public parameter.

**Both nits:** `judge.latency_ms` starts its clock inside `with_span` now, and the doubled phrase in `record_jury_span`'s docstring is gone.

Checks: full non-integration suite 3876 passed, `ruff check` / `ruff format --check` / `basedpyright` / `mkdocs build --strict` all clean.

currentlycodinng (approved):
My finding is addressed, and I verified the evidence rather than taking the comment at its word, since the whole point of that finding was that nobody had checked the consumer.

I pulled `apps/traces-api/src/handler/otel/utils/adapter-patterns.ts` from orquesta-web and read `extractCommonUsage` and the pattern table it reads from. Every dropped key is covered by a fallback the branch still emits:

- `promptTokens` (line 84) lists `gen_ai.usage.input_tokens` first, ahead of the removed `gen_ai.usage.prompt_tokens`, `prompt_tokens` and `input_tokens`; same shape for `completionTokens` (102) and `totalTokens` (120). `tracing.py:299-301` emits all three surviving spellings.
- Cached tokens are read at line 613 from `gen_ai.usage.cache_read.input_tokens`, which is exactly what `tracing.py:306` now emits. The flat `gen_ai.usage.prompt_tokens_details.cached_tokens` that was removed is read by nothing: line 598 reads `gen_ai.usage.prompt_tokens_details` as an object, so the dotted spelling never matched anyway.
- Reasoning tokens are read at line 654 from `gen_ai.usage.reasoning.output_tokens`, and that is the spelling you switched to. The old `completion_tokens_details.reasoning_tokens` only resolves when the parent is an object (line 636), so the flat key was dead on the platform side too. That change is a fix, not just a rename.
- `calls` moved from a bare key to `gen_ai.usage.calls`. The platform reads neither: no hit for either spelling anywhere in orquesta-web.

So the deletions are safe and two of them were correcting keys the platform could never have read. This is the version of the check I was asking for, and putting the file and function name in the comment at `tracing.py:308-310` is what makes it re-checkable later.

I also re-ran the span-hierarchy probe against `55a31b0b` after the `af7262a7` rework. Six concurrent `run_pairwise` calls against an in-memory exporter: 6 `orq.pairwise_jury` spans, 36 `orq.judge` spans, exactly 6 judges under each, and zero judge spans parented to the wrong jury. The single-jury-per-comparison change holds under concurrency.

**One must-fix, and it is why CI is red on all four legs.**

```
tests/common/test_jury_tracing.py:295:54 - error: Argument of type "dict[str, float | Unknown]"
cannot be assigned to parameter "verdicts" of type "dict[str, str | None]" in function "_make_judge_fn"
```

`test_non_pairwise_verdicts_reach_the_span_unchanged` is parametrized over `0.75`, `True` and `'yes'`, but `_make_judge_fn` at line 72 is annotated `dict[str, str | None]`. The annotation contradicts the behaviour the test exists to pin, so widening it (`dict[str, Any]`, or the explicit `str | float | bool | None`) is the fix rather than a cast at the call site. Typecheck gates the tests in that job, so nothing downstream of it has run in CI at all. Locally, with only that annotation in question, `ruff check src` is clean and `pytest -m 'not integration'` gives 3883 passed, 2 skipped, so I expect all four legs green once it lands.

One observation, no ask attached. On a comparison where every judge flipped, the jury span reads `judges_configured=3, judges_succeeded=0, judges_failed=0, flipped=3`. That is accurate, since an abstention on a flip is neither a success nor a failure, and `jury.flipped_judges` names them. But the three counters do not reconcile to `judges_configured`, so anyone alerting on `judges_failed > 0` will miss a total position-bias meltdown, which is the case you would most want an alert on. A `jury.judges_abstained` would make the span self-balancing.

Approving. Fix the annotation and it should go green.

---

What I checked: fetched `55a31b0b`, reproduced the CI typecheck failure locally and confirmed it is the only one, ran the full non-integration suite (3883 passed) and `tests/common/test_jury_tracing.py` (22 passed) which CI never reached, fetched orquesta-web's `adapter-patterns.ts` and traced each dropped attribute through `extractCommonUsage` and its pattern table against what `tracing.py:299-322` now emits, searched orquesta-web for the `calls` attribute under both spellings, and re-ran the concurrent span-parenting probe on the reworked hierarchy.

currentlycodinng (commented):
The annotation fix in `28a479bc` is better than what I suggested: `VerdictValue | None` says what the helper actually accepts, where the `Any` I proposed would have just silenced the checker. Confirmed green locally.

Re-reviewing because #113 landed on main while this sat, and the branch is now conflicted against it. I resolved it locally to check whether it is mechanical, and it is, but neither side can be taken wholesale.

**Two hunks, both in `pairwise.py`.**

The first is the vote-building loop. This branch indented it into the `jury_span` context manager; main renamed `_as_str` to `_as_vote`, which narrows the return to `Literal['A', 'B', 'tie'] | None`. Taking your side keeps the span and loses the narrowing; taking main's keeps the narrowing and drops the block out of the span, which would put every judge span back under the ambient context. The resolution is your indentation with `vote=_as_vote(vote)`.

The second is pure addition on both sides: main's new `_as_vote` replaces your `_as_str`, and your `_record_pairwise_span` is untouched by main. Keep `_as_vote`, delete `_as_str`, keep `_record_pairwise_span`.

With that resolution: `ruff check` clean, `basedpyright` 0 errors, `pytest -m 'not integration'` 3968 passed, 2 skipped. I re-ran the concurrency probe on the merged tree and the hierarchy is intact — 6 jury spans, 36 judge spans, 6 per jury, none misparented.

**One thing the merge surfaces, for a follow-up rather than this PR.** `_record_pairwise_span` hardcodes `'jury.aggregator': 'pairwise_plurality'`, which is correct per comparison, since bt-sigma is a run-level rollup that `build_report` applies over the comparisons afterwards. But that means a run with `aggregation='bt-sigma'` emits nothing distinguishing it: every comparison span still says `pairwise_plurality`, and `bt_sigma_aggregation` produces no span of its own, so the reliability weighting that reweights judges and can move the winner is entirely untraced. Given this PR's whole purpose is making jury behaviour visible, the rollup is the piece most worth a span, and the per-judge `sigma` that #113 added to `JudgeStats` is the obvious attribute for it.

Approval stands.

---

What I checked: fetched `455db1e7`, read `28a479bc`, trial-merged `origin/main` (which now contains #113) and resolved both `pairwise.py` conflicts, then ran `ruff check src`, `basedpyright` and the full non-integration suite on the merged tree, plus the concurrent span-parenting probe.

Baukebrenninkmeijer (approved):

PR #113, opened 2026-08-06
feat(jury): BT-sigma reliability-weighted aggregation for pairwise judging (RES-1024)

KarinaKKarinaK:
Closes RES-1024.

Adds **BT-sigma** (reliability-weighted Bradley-Terry, [arXiv:2602.16610](https://arxiv.org/abs/2602.16610)) as an opt-in aggregation mode for the pairwise jury. It jointly fits item skills and a per-judge discriminator by unsupervised MLE on the run's own reconciled votes: no labels, no extra LLM calls, and noisy judges get down-weighted instead of outvoting sharp ones. The same model class is independently derived with identifiability and consistency proofs in [arXiv:2601.21817](https://arxiv.org/abs/2601.21817); the survey comparing it against Crowd-BT, Temp-BT, Dawid-Skene, SkillAggregation and HJA is on the ticket.

## What's here

- **`ranking.py`** (new): `fit_bt()` covering hard BT, soft BT, BT-sigma and hard BT-sigma via two flags, plus `cycle_rate()` (the paper's independent consistency diagnostic) and `binarize()`. Pure Python on purpose: the problem is N items + K judges parameters, closed-form gradients converge in milliseconds, and the runtime dependency set stays untouched (no scipy/numpy). Deterministic zero init, Adam with lr decay, tiny ridge for disconnected graphs / perfect separation, mean-zero skills + unit-geometric-mean sigmas for identifiability, single-judge fallback to plain BT with a warning. The per-aspect variant (BT-sigma-asp) is deliberately skipped, the paper finds it marginal.
- **`pairwise.py`** (additive): `build_report(comparisons, aggregation='bt-sigma')` fits hard BT-sigma over the run and attaches a reliability-weighted rollup (`report.bt_sigma`: fitted P(A beats B), per-judge sigmas, weighted winner per comparison, weighted win rates) and each judge's `sigma` on `JudgeStats`. **Plurality stays the default and its headline rates are unchanged**, so the two aggregations are directly comparable side by side.
- **Docs**: new section in `pairwise-judging.md` with usage, the single-judge and split-panel semantics, and the honest limitation (unsupervised aggregation rewards internal consistency; it protects against noisy judges, not a coordinated biased majority).
- Comparison records carry judge identity so the round-robin panel work (RES-1023) can reuse them.

## Design notes worth reviewing

- **Votes feed the fit as hard preferences** (A=1, B=0, tie=0.5). We have no token logits through the router, so the paper's soft-probability mode is out of reach; hard BT-sigma is fully faithful to the paper and is its own recommended robust variant under high inconsistency. Documented as a limitation.
- **Symmetrisation comes free**: reconcile_pair already runs every judge in both orderings, which satisfies the model's commutativity requirement (paper Eq. 4) before the fit ever sees a vote.
- **Gradient stationarity floor**: without it, Adam's scale invariance turns a 1e-16 float residue at a symmetric stationary point (two judges in perfect opposition) into a full-size step, and the fit crowns one judge ~16,000x more reliable than the other on perfectly ambiguous data. With the floor, ambiguous stays ambiguous. There's a test for exactly this.
- **Soft priors instead of hard identifiability projections** (evolved during live testing): skills are mean-centred additively, and the sigma scale is anchored by a log-normal prior alone. The original hard geometric-mean projection was a multiplicative ratchet on unanimous panels: every tau drifts down together, the projection's compensating skill rescale inflates skills without bound, and Adam's normalised steps cap the ridge's corrective pull, so skills exploded to +-2000 on real data. Consequence documented on BTFit: **sigma ratios within a fit are meaningful, absolute values across fits are not** (which is all the vote-weighting uses).
- **Convergence is measured on iterate movement, not gradient norm** (the ridge leaves a permanent small gradient at the regularised optimum), with tolerance and iteration cap sized to parameter scale after three drift-family bugs surfaced in manual testing: consistent single judge, single-judge fallback, unanimous panels.

## Live validation (examples/pairwise/bt_sigma_ranking.py)

The committed example is the paper in miniature on a real jury: five questions with four graded answer tiers as global BT items, judged by a lopsided panel from the verified jury presets (gpt-5.4-mini, gpt-5.4-nano, deepseek/deepseek-v4-flash). Tier design induces disagreement via the documented verbosity bias of small judges. Across four independent live runs (two mine, two Karina's manual testing): the exact tier ranking was recovered 4/4, and the sigma ordering deepseek < mini < nano replicated 4/4 with up to a 10x spread, nano correctly identified as least reliable, with zero labels. Judges default to temperature=0 for reproducibility.

## Tests (24 new)

Synthetic-recovery regressions encoding the paper's claims: ranking recovery under heterogeneous judge noise (soft and sampled-hard), fitted sigma ordering matching injected noise ordering, adversarial-judge down-weighting (majority-anchored, since a single consistent adversary vs a single good judge is the ill-posed case the paper's limitations name), sigma-vs-cycle-rate agreement, identifiability constraints, determinism, perfect-separation and disconnected-graph finiteness, single-judge fallback; plus the pairwise integration contract (default report byte-identical, weighted consensus flipping plurality-outvoted rows, split panel inconclusive).

## Verification

- Full suite: 3611 passed (dashboard + hub included), ruff + format clean, basedpyright 0 errors, mkdocs build --strict passes.
- Survey + implementation-decision comments on RES-1024.

## Sequencing note

#111 (jury template-var namespace) also touches `pairwise.py`. This PR's changes there are additive and in a different region (consensus/report, not replacements), so a conflict is unlikely but possible; whoever lands second rebases trivially.

### DIFF
--- a/examples/pairwise/bt_sigma_ranking.py
+++ b/examples/pairwise/bt_sigma_ranking.py
@@ -140,6 +140,9 @@ async def main() -> None:
     jury = llm_jury_pairwise(
         judges=list(JUDGES),
         criteria='Which answer is more scientifically accurate?',
+        # Near-deterministic judges: run-to-run sigma ORDERING replicates
+        # regardless, but zero temperature tightens the magnitudes too.
+        temperature=0.0,
     )
     pairs = [(hi, lo) for i, hi in enumerate(TIERS) for lo in TIERS[i + 1 :]]
 
--- a/examples/pairwise/bt_sigma_ranking.py
+++ b/examples/pairwise/bt_sigma_ranking.py
@@ -17,7 +17,7 @@ defaults). Needs ORQ_API_KEY.
 
 Usage::
 
-    python examples/bt_sigma_ranking.py [STRONG WEAK_1 WEAK_2]
+    python examples/pairwise/bt_sigma_ranking.py [STRONG WEAK_1 WEAK_2]
 """
 
 import asyncio
--- a/src/evaluatorq/pairwise.py
+++ b/src/evaluatorq/pairwise.py
@@ -9,6 +9,7 @@ consistency-gated vote (RES-760, ADR-24).
 from __future__ import annotations
 
 import asyncio
+import math
 from collections import Counter
 from typing import TYPE_CHECKING
 
@@ -106,6 +107,11 @@ class JudgeStats(BaseModel):
         description='Flips over completed pairs (both orderings decisive); 0.0 when no pair was flippable'
     )
     tie_rate: float = Field(description='Ties over total comparisons the judge saw')
+    sigma: float | None = Field(
+        default=None,
+        description='BT-sigma discriminator (smaller = more reliable); set only when the report was built '
+        "with aggregation='bt-sigma'",
+    )
 
 
 class PairwiseReport(BaseModel):
@@ -126,14 +132,117 @@ class PairwiseReport(BaseModel):
         description='Mean inter-judge agreement (modal vote share) across comparisons; None if none were decisive'
     )
     per_judge: list[JudgeStats] = Field(default_factory=list, description='Per-judge behaviour breakdown')
+    bt_sigma: BTSigmaAggregation | None = Field(
+        default=None,
+        description="Reliability-weighted rollup; set only when the report was built with aggregation='bt-sigma'",
+    )
 
 
 def _rate(count: int, total: int) -> float | None:
     return count / total if total else None
 
 
-def build_report(comparisons: Sequence[PairwiseComparison]) -> PairwiseReport:
-    """Roll a set of pairwise comparisons up into headline and per-judge metrics."""
+class BTSigmaAggregation(BaseModel):
+    """Reliability-weighted rollup of a pairwise run (BT-sigma, arXiv:2602.16610).
+
+    Fitted unsupervised on the run's own reconciled votes: hard BT-sigma jointly
+    infers the A-vs-B skill gap and a discriminator per judge, then re-derives
+    each comparison's winner as a reliability-weighted vote (weight ``1/sigma``)
+    instead of uniform plurality. Down-weights noisy judges; needs no labels.
+    """
+
+    p_a_beats_b: float = Field(description='Fitted global probability that A beats B (logistic of the skill gap)')
+    skill_gap: float = Field(description='Fitted skill difference s_A - s_B')
+    judge_sigmas: dict[str, float] = Field(
+        default_factory=dict,
+        description='Per-judge discriminator; smaller = sharper/more reliable. Empty on single-judge fallback.',
+    )
+    winners: list[str] = Field(
+        default_factory=list,
+        description='Reliability-weighted winner per comparison, aligned with the input order',
+    )
+    a_win_rate: float | None = Field(description='Weighted A-wins over decisive comparisons; None if none decisive')
+    b_win_rate: float | None = Field(description='Weighted B-wins over decisive comparisons; None if none decisive')
+    tie_rate: float = Field(description='Weighted-tie comparisons over all comparisons')
+    inconclusive_rate: float = Field(description='Comparisons no judge decided, over all comparisons')
+    fit_warnings: list[str] = Field(default_factory=list, description='Degradations applied during the BT fit')
+
+
+_VOTE_TO_P = {'A': 1.0, 'B': 0.0, 'tie': 0.5}
+
+
+def bt_sigma_aggregation(comparisons: Sequence[PairwiseComparison]) -> BTSigmaAggregation:
+    """Fit hard BT-sigma over a run's reconciled votes and re-derive winners.
+
+    The two "items" are the run's A and B sides; every decisive reconciled vote
+    is one comparison between them. Reconciliation has already symmetrised
+    position bias (both orderings per judge), satisfying the model's
+    commutativity requirement. Categorical votes make hard BT-sigma the natural
+    variant, which is also the paper's most robust one under inconsistency.
+    """
+    from evaluatorq.ranking import JudgedComparison, fit_bt
+
+    records = [
+        JudgedComparison(judge=v.model, item_a='A', item_b='B', p_a=_VOTE_TO_P[str(v.vote)])
+        for c in comparisons
+        for v in c.votes
+        if v.vote is not None
+    ]
+    if not records:
+        return BTSigmaAggregation(
+            p_a_beats_b=0.5,
+            skill_gap=0.0,
+            winners=['inconclusive'] * len(comparisons),
+            a_win_rate=None,
+            b_win_rate=None,
+            tie_rate=0.0,
+            inconclusive_rate=1.0 if comparisons else 0.0,
+            fit_warnings=['no decisive votes to fit on'],
+        )
+
+    fit = fit_bt(records, judge_sigma=True, hard=True)
+    # Single-judge fallback leaves sigmas empty; weight uniformly then.
+    weights = {judge: 1.0 / sigma for judge, sigma in fit.sigmas.items()}
+
+    winners: list[str] = []
+    for c in comparisons:
+        w: dict[str, float] = {'A': 0.0, 'B': 0.0, 'tie': 0.0}
+        for v in c.votes:
+            if v.vote is None:
+                continue
+            w[str(v.vote)] += weights.get(v.model, 1.0)
+        if not any(w.values()):
+            winners.append('inconclusive')
+            continue
+        top = max(w.values())
+        leaders = [k for k, val in w.items() if val == top]
+        # Mirrors plurality semantics: a unique leader wins, a split is inconclusive.
+        winners.append(leaders[0] if len(leaders) == 1 else 'inconclusive')
+
+    counts = Counter(winners)
+    decisive = counts['A'] + counts['B']
+    total = len(comparisons)
+    return BTSigmaAggregation(
+        p_a_beats_b=1.0 / (1.0 + math.exp(-(fit.skills['A'] - fit.skills['B']))),
+        skill_gap=fit.skills['A'] - fit.skills['B'],
+        judge_sigmas=fit.sigmas,
+        winners=winners,
+        a_win_rate=_rate(counts['A'], decisive),
+        b_win_rate=_rate(counts['B'], decisive),
+        tie_rate=counts['tie'] / total if total else 0.0,
+        inconclusive_rate=counts['inconclusive'] / total if total else 0.0,
+        fit_warnings=fit.warnings,
+    )
+
+
+def build_report(comparisons: Sequence[PairwiseComparison], *, aggregation: str = 'plurality') -> PairwiseReport:
+    """Roll a set of pairwise comparisons up into headline and per-judge metrics.
+
+    ``aggregation='plurality'`` (default) keeps the existing uniform plurality
+    consensus. ``aggregation='bt-sigma'`` additionally fits hard BT-sigma over
+    the run and attaches the reliability-weighted rollup (``report.bt_sigma``)
+    plus each judge's discriminator (``JudgeStats.sigma``); the headline
+    plurality rates are unchanged so the two aggregations stay comparable."""
     total = len(comparisons)
     winners = Counter(c.winner for c in comparisons)
     decisive = winners['A'] + winners['B']
@@ -165,6 +274,15 @@ def build_report(comparisons: Sequence[PairwiseComparison]) -> PairwiseReport:
         and (rate := _agreement_rate([v.vote for v in c.votes if v.vote is not None])) is not None
     ]
 
+    bt_block: BTSigmaAggregation | None = None
+    if aggregation == 'bt-sigma':
+        bt_block = bt_sigma_aggregation(comparisons)
+        for stats in judge_stats:
+            stats.sigma = bt_block.judge_sigmas.get(stats.model)
+    elif aggregation != 'plurality':
+        msg = f"unknown aggregation {aggregation!r}; expected 'plurality' or 'bt-sigma'"
+        raise ValueError(msg)
+
     return PairwiseReport(
         comparisons=total,
         a_win_rate=_rate(winners['A'], decisive),
@@ -173,6 +291,7 @@ def build_report(comparisons: Sequence[PairwiseComparison]) -> PairwiseReport:
         inconclusive_rate=winners['inconclusive'] / total if total else 0.0,
         mean_agreement=sum(agreements) / len(agreements) if agreements else None,
         per_judge=judge_stats,
+        bt_sigma=bt_block,
     )
 
 
--- a/src/evaluatorq/pairwise.py
+++ b/src/evaluatorq/pairwise.py
@@ -195,6 +195,26 @@ _VOTE_TO_P = {'A': 1.0, 'B': 0.0, 'tie': 0.5}
 _MIN_VOTES_FOR_SIGMA = 5
 
 
+def _uniform_plurality_aggregation(
+    comparisons: Sequence[PairwiseComparison], warnings: list[str], *, converged: bool = True
+) -> BTSigmaAggregation:
+    winners = [c.winner for c in comparisons]
+    counts = Counter(winners)
+    decisive = counts['A'] + counts['B']
+    total = len(comparisons)
+    return BTSigmaAggregation(
+        p_a_beats_b=0.5,
+        skill_gap=0.0,
+        winners=winners,
+        a_win_rate=_rate(counts['A'], decisive),
+        b_win_rate=_rate(counts['B'], decisive),
+        tie_rate=counts['tie'] / total if total else 0.0,
+        inconclusive_rate=counts['inconclusive'] / total if total else 0.0,
+        converged=converged,
+        fit_warnings=warnings,
+    )
+
+
 def bt_sigma_aggregation(comparisons: Sequence[PairwiseComparison]) -> BTSigmaAggregation:
     """Fit hard BT-sigma over a run's reconciled votes and re-derive winners.
 
@@ -215,6 +235,12 @@ def bt_sigma_aggregation(comparisons: Sequence[PairwiseComparison]) -> BTSigmaAg
     ``fit_warnings`` names them.
     """
     from evaluatorq.ranking import JudgedComparison, comparisons_per_judge, fit_bt
+
+    if any(v.vote is not None and not v.completed for c in comparisons for v in c.votes):
+        return _uniform_plurality_aggregation(
+            comparisons,
+            ['single-ordering data: position-bias symmetrisation unavailable; used uniform plurality instead'],
+        )
 
     # judge -> vote -> count. Collapsing keeps the optimum identical (the
     # likelihood is additive over identical records) while the per-iteration
@@ -243,6 +269,9 @@ def bt_sigma_aggregation(comparisons: Sequence[PairwiseComparison]) -> BTSigmaAg
 
     fit = fit_bt(records, judge_sigma=True, hard=True)
     fit_warnings = list(fit.warnings)
+    if not fit.converged:
+        fit_warnings.append('non-converged fit: used uniform plurality winners instead')
+        return _uniform_plurality_aggregation(comparisons, fit_warnings, converged=False)
 
     # Two-item degeneracy guard: a unanimous judge's sigma is a closed-form
     # function of its own one-sidedness (its marginal pins the fit), so 1/sigma
--- a/src/evaluatorq/pairwise.py
+++ b/src/evaluatorq/pairwise.py
@@ -10,8 +10,9 @@ from __future__ import annotations
 
 import asyncio
 import math
+import statistics
 from collections import Counter
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal, cast
 
 from pydantic import BaseModel, Field
 
@@ -73,7 +74,9 @@ class PairwiseVote(BaseModel):
     """One judge's reconciled verdict for a single A-vs-B comparison."""
 
     model: str = Field(description='Judge model ID')
-    vote: str | None = Field(description="Reconciled vote: 'A', 'B', 'tie', or None if the judge abstained")
+    vote: Literal['A', 'B', 'tie'] | None = Field(
+        description="Reconciled vote: 'A', 'B', 'tie', or None if the judge abstained"
+    )
     flipped: bool = Field(default=False, description='True if the judge disagreed with itself across orderings')
     completed: bool = Field(
         default=True,
@@ -149,6 +152,19 @@ class BTSigmaAggregation(BaseModel):
     infers the A-vs-B skill gap and a discriminator per judge, then re-derives
     each comparison's winner as a reliability-weighted vote (weight ``1/sigma``)
     instead of uniform plurality. Down-weights noisy judges; needs no labels.
+
+    Two caveats of the two-item collapse, both handled here:
+
+    - With only two items a judge's sigma is pinned by its own vote
+      distribution, so a judge whose decisive votes are unanimous gets an
+      arbitrarily small sigma that measures one-sidedness (e.g. position or
+      verbosity bias), not reliability. Such judges are excluded from the
+      weighting — they vote with a neutral weight, their sigma is dropped from
+      ``judge_sigmas``, and ``fit_warnings`` names them.
+    - ``p_a_beats_b``/``skill_gap`` come from the pooled global fit, while
+      ``winners`` come from the per-comparison weighted vote. These are two
+      different estimators and need not agree; read ``p_a_beats_b`` as the
+      run-level headline and ``winners`` as the per-row calls.
     """
 
     p_a_beats_b: float = Field(description='Fitted global probability that A beats B (logistic of the skill gap)')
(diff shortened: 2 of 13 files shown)


KarinaKKarinaK:
Screenshot 2026-08-06 at 11 47 19 AM

src/evaluatorq/pairwise.py
```diff
@@ -123,14 +129,117 @@ class PairwiseReport(BaseModel):
...
+    return BTSigmaAggregation(
+        p_a_beats_b=1.0 / (1.0 + math.exp(-(fit.skills['A'] - fit.skills['B']))),
+        skill_gap=fit.skills['A'] - fit.skills['B'],
+        judge_sigmas=fit.sigmas,
+        winners=winners,
+        a_win_rate=_rate(counts['A'], decisive),
+        b_win_rate=_rate(counts['B'], decisive),
+        tie_rate=counts['tie'] / total if total else 0.0,
+        inconclusive_rate=counts['inconclusive'] / total if total else 0.0,
+        fit_warnings=fit.warnings,
```
Baukebrenninkmeijer: `fit.warnings` is copied through but `fit.converged` is dropped, and `BTSigmaAggregation` has no boolean for it. So a fit that ran into the 20 000-iteration cap produces winners, weights and `p_a_beats_b` that look exactly like a converged fit — the only trace is the prose string `fit stopped at the 20000-iteration cap (max|step|=...)` inside `fit_warnings`.

That makes "was this fit trustworthy?" a substring match on free text (and `tests/unit/test_pairwise_bt_sigma.py` already does exactly that at line 105, so the wording is load-bearing but unpinned).

Add `converged: bool` to `BTSigmaAggregation` and pass `fit.converged` through, so callers can gate on it programmatically.

src/evaluatorq/pairwise.py
```diff
@@ -123,14 +129,117 @@ class PairwiseReport(BaseModel):
...
+    The two "items" are the run's A and B sides; every decisive reconciled vote
+    is one comparison between them. Reconciliation has already symmetrised
+    position bias (both orderings per judge), satisfying the model's
+    commutativity requirement. Categorical votes make hard BT-sigma the natural
+    variant, which is also the paper's most robust one under inconsistency.
+    """
+    from evaluatorq.ranking import JudgedComparison, fit_bt
+
+    records = [
+        JudgedComparison(judge=v.model, item_a='A', item_b='B', p_a=_VOTE_TO_P[str(v.vote)])
```
Baukebrenninkmeijer: `_VOTE_TO_P[str(v.vote)]` is guarded only by `v.vote is not None`, but `PairwiseVote.vote` is typed `str | None` — any value outside `{'A','B','tie'}` raises an uncaught `KeyError` and takes down the whole `build_report(aggregation='bt-sigma')` call, in a module that otherwise degrades gracefully (`no decisive votes to fit on`, `single judge`, …).

Same pattern at line 210 (`w[str(v.vote)] += ...`), which is a separate loop over `c.votes` and so needs its own guard.

Cheapest durable fix is at the type: make `vote` a `Literal['A','B','tie'] | None` so the constraint is enforced where the value is constructed. Otherwise `.get()` with a skip + a `fit_warnings` entry.

src/evaluatorq/pairwise.py
```diff
@@ -123,14 +129,117 @@ class PairwiseReport(BaseModel):
...
+        w: dict[str, float] = {'A': 0.0, 'B': 0.0, 'tie': 0.0}
+        for v in c.votes:
+            if v.vote is None:
+                continue
+            w[str(v.vote)] += weights.get(v.model, 1.0)
+        if not any(w.values()):
+            winners.append('inconclusive')
+            continue
+        top = max(w.values())
+        leaders = [k for k, val in w.items() if val == top]
```
Baukebrenninkmeijer: `val == top` is exact float equality on sums of `1.0/sigma` weights. Two judges whose sigmas are mathematically equal but differ in the last bits (asymmetric comparison counts, non-mirrored data) will silently break a genuine tie into a decisive leader — the opposite of the `inconclusive` semantics this branch exists for. The current fixture is perfectly symmetric, so the arithmetic happens to stay bit-identical and the test passes.

`math.isclose(val, top, rel_tol=1e-9)` when collecting `leaders`.

src/evaluatorq/ranking.py
```diff
@@ -0,0 +1,337 @@
...
+    """Map soft preferences to hard ones (paper: ``y = 1[p >= 0.5]``), keeping exact ties at 0.5."""
+    out: list[JudgedComparison] = []
+    for c in comparisons:
+        # 0.5 is the tie point of the preference scale.
+        hard = 0.5 if c.p_a == 0.5 else (1.0 if c.p_a > 0.5 else 0.0)
+        out.append(c.model_copy(update={'p_a': hard}))
+    return out
+
+
+def fit_bt(
```
Baukebrenninkmeijer, line 169: **Judgement call for @Baukebrenninkmeijer** — not asking for a rewrite in this PR.

This is a hand-rolled Adam loop (momentum, variance, bias correction, LR decay, gradient floor, two ridge priors, 20 000-iteration cap) for a Bradley-Terry fit. BT MLE has a standard alternative: the Zermelo/MM update, which is a few lines, hyperparameter-free, and monotone in the likelihood by construction — no learning rate, no decay schedule, no gradient floor, and no unbounded-sigma failure mode to contain.

What makes this worth raising rather than filing as a nit: most of the tuned constants here (`_LR`, `_LR_DECAY_STEPS`, `_GRAD_FLOOR`, `_TOL`, `_MAX_ITER`) and the comments justifying them exist to manage hazards the optimiser choice introduced, not hazards of the model. The judge-discriminator term is the part that isn't textbook MM, so this may well be the right call — but it should be a decision on the record, since every future edit to this file has to reason about the schedule.

Related: the `_MAX_ITER` comment says "fits are milliseconds". That's a pure-Python pass over every comparison, up to 20 000 times, run synchronously inside `build_report()`/`save()`. Worth a number from a realistic run size before the docstring promises milliseconds.

src/evaluatorq/ranking.py
```diff
@@ -0,0 +1,337 @@
...
+        description='Per-judge discriminator; smaller = sharper/more reliable. Only RATIOS between judges '
+        'in the same fit are meaningful (the log-normal prior anchors the scale loosely); do not compare '
+        'absolute values across fits. Empty when fitted without judge_sigma.'
+    )
+    ranking: list[str] = Field(description='Item IDs, best first')
+    converged: bool = Field(description='True when the iterate stopped moving before the iteration cap')
+    iterations: int = Field(description='Optimizer iterations used')
+    warnings: list[str] = Field(default_factory=list, description='Degradations applied during the fit')
+
+    def reliability(self, judge: str) -> float:
```
Baukebrenninkmeijer, line 145: `reliability()` has no caller — `bt_sigma_aggregation` recomputes `1.0 / sigma` inline (pairwise.py:202) and the example inlines the sort too.

Same for `comparisons_per_judge()` at line 322: exported in `__all__`, documented as "a cheap sanity check before trusting sigmas", and called only by its own unit test.

Either wire them in where they'd earn their keep — `reliability()` at the pairwise call site, `comparisons_per_judge()` into `fit_warnings` when a judge has too few comparisons to trust its sigma, which is the check the docstring describes — or drop them until something needs them. Two exported functions nothing calls is API surface to keep working forever.

tests/unit/test_pairwise_bt_sigma.py
```diff
@@ -0,0 +1,145 @@
...
+    assert with_bt.inconclusive_rate == plain.inconclusive_rate
+
+
+def test_abstaining_panel_degrades_to_inconclusive() -> None:
+    rows = [_comparison([_vote('a', None), _vote('b', None)]) for _ in range(3)]
+    block = bt_sigma_aggregation(rows)
+    assert block.winners == ['inconclusive'] * 3
+    assert block.a_win_rate is None
+    assert block.inconclusive_rate == 1.0
+    assert any('no decisive votes' in w for w in block.fit_warnings)
```
Baukebrenninkmeijer, line 120: Two gaps in an otherwise good suite:

1. **Fit quality is asserted by substring.** This and line 115 match on `'no decisive votes'` / `'single judge'`, which is the only handle on degraded fits (see my note on propagating `converged`). Nothing pins the cap-hit warning, so that path is untested.
2. **No tie vote reaches the aggregation end-to-end.** The `_heterogeneous_run` fixture emits only `'A'`/`'B'`, so `_VOTE_TO_P['tie'] = 0.5` and the hard-binarize keep-as-tie branch are never exercised through `bt_sigma_aggregation`. A regression that folded ties into A would pass this suite, and ties are common in real jury runs.

Also worth one test that pins the fit against an independently derived value on a tiny symmetric case. Right now every `fit_bt` assertion is qualitative (ranking order, sigma ordering, sum-to-zero, finiteness) — a scaling bug in the gradient that preserved monotonicity would go unnoticed.

src/evaluatorq/pairwise.py
```diff
@@ -123,14 +129,117 @@ class PairwiseReport(BaseModel):
...
 
 
-def build_report(comparisons: Sequence[PairwiseComparison]) -> PairwiseReport:
-    """Roll a set of pairwise comparisons up into headline and per-judge metrics."""
+class BTSigmaAggregation(BaseModel):
+    """Reliability-weighted rollup of a pairwise run (BT-sigma, arXiv:2602.16610).
+
+    Fitted unsupervised on the run's own reconciled votes: hard BT-sigma jointly
+    infers the A-vs-B skill gap and a discriminator per judge, then re-derives
+    each comparison's winner as a reliability-weighted vote (weight ``1/sigma``)
```
Baukebrenninkmeijer, line 153: **Judgement call for @Baukebrenninkmeijer.** The RES-1024 plan lists the outputs as "reliability-weighted consensus per comparison, plus per-judge sigma **and cycle-rate** in `JudgeStats`". `JudgeStats` gained `sigma`; `cycle_rate()` is never called from `bt_sigma_aggregation` or `build_report`.

My read is that this is correct — with two fixed items (`'A'`/`'B'`) there are no 3-cycles to rate, so cycle-rate only means something on the N-item `fit_bt` leaderboard path, which is how `docs/pairwise-judging.md` describes it. But that reasoning appears nowhere, so it currently reads as a silently dropped deliverable. One line in the PR body or the docs closes it.

Separately: `p_a_beats_b`/`skill_gap` come from the pooled global fit while `winners` come from the per-row weighted vote — two different estimators that aren't guaranteed to agree. Worth saying so in the `BTSigmaAggregation` docstring before someone reads one as the aggregate of the other.

examples/pairwise/bt_sigma_ranking.py
```diff
@@ -0,0 +1,184 @@
...
+- a discriminator per judge (smaller = more reliable),
+- a directed 3-cycle rate per judge, the paper's independent consistency
+  signal that the discriminator should track.
+
+Cost: questions x 6 pairs x judges x 2 orderings LLM calls (~180 with the
+defaults). Needs ORQ_API_KEY.
+
+Usage::
+
+    python examples/bt_sigma_ranking.py [STRONG WEAK_1 WEAK_2]
```
Baukebrenninkmeijer: Path is stale — the file is at `examples/pairwise/bt_sigma_ranking.py`, so copy-pasting this usage line fails.

Baukebrenninkmeijer (dismissed):

Strong PR. The BT-sigma fit is carefully written, the failure modes you hit (sigma blow-up on unanimous panels, the ratchet, gradient norm never reaching zero) are documented in the comments rather than papered over, and the docs page plus the 2-pager line up with the code. Approving — nothing blocks.

Two things I'd like @Baukebrenninkmeijer to weigh in on, flagged inline:

- **Adam vs. MM for the BT fit.** Bradley-Terry MLE has a standard hyperparameter-free monotone solver (Zermelo/MM); this ships a hand-rolled Adam loop with ~10 tuned constants, several of which exist purely to contain failure modes Adam introduces. Worth a deliberate decision rather than inheriting it.
- **`cycle_rate` on `JudgeStats`.** The RES-1024 plan lists per-judge cycle-rate as an output; only `sigma` shipped. Plausibly because cycles are undefined for two fixed items — but that reasoning isn't written down anywhere.

The rest is small: a non-converged fit is only visible as a warning *string*, two unguarded dict lookups, a float `==` tie test, and two exported helpers nothing calls.

Baukebrenninkmeijer (changes requested):

Correcting my own verdict: I approved this while flagging two decisions for @Baukebrenninkmeijer and several functional defects. An approval says none of that blocks merge, which contradicts asking for a decision first. Switching to `REQUEST_CHANGES`; my earlier approval is dismissed and the inline comments on it still stand.

Needs a decision before merge:

1. **Adam vs. MM for the BT fit** (`ranking.py:143`) — Bradley-Terry MLE has a standard hyperparameter-free monotone solver. This ships a hand-rolled Adam loop whose tuned constants mostly exist to contain failure modes Adam introduces. Karina's judge-discriminator term is the part that isn't textbook, so this may well be right — but every future edit to this file inherits the schedule, so it should be a decision on the record.
2. **`cycle_rate` on `JudgeStats`** — listed as an RES-1024 output, not shipped. Probably correct (no 3-cycles over two fixed items), but the reasoning isn't written down.

Functional, independent of those:

3. **`pairwise.py:231`** — `fit.converged` is dropped, so a fit that hit the 20 000-iteration cap is indistinguishable from a converged one except by substring-matching warning text.
4. **`pairwise.py:183` and `:210`** — unguarded `_VOTE_TO_P[...]` / `w[...]` lookups; any vote outside `{'A','B','tie'}` is an uncaught `KeyError` in a module that otherwise degrades gracefully.
5. **`pairwise.py:215`** — exact float `==` for tie detection over `1/sigma` sums.
6. **Tests** — no tie vote reaches `bt_sigma_aggregation` end-to-end, and no assertion pins the fit against an independently derived value.

The fit itself is careful work and the documented failure modes are handled honestly — this is about the surrounding contract.

Baukebrenninkmeijer:
🤖 Correction to my own review comment above (the Adam-vs-MM one), from Bauke:

I wrote that "BT MLE has a standard hyperparameter-free monotone solver (Zermelo/MM)". That holds for **plain BT** — concave in the skills, closed-form multiplicative updates. It does **not** hold for BT-sigma as a joint problem: `(s_a - s_b) / sigma_k` makes sigma a discrimination parameter (same structure as 2PL IRT), the joint objective is not concave, and there is no off-the-shelf MM for it. So "just use MM instead" is not an available option, and my comment overstated the alternative.

The version of the point that survives: the fit is currently trusted on its constants alone. Nothing outside `fit_bt` can distinguish a converged fit from one that rode to the 20 000-iteration cap, and no test would catch a gradient bug that rescaled the skills while preserving their order.

So the ask narrows to what is cheap and load-bearing:

1. Propagate `converged` out of the fit (already in the review above).
2. Expose `max_iter` / `tol` on `fit_bt`.
3. Pin the fitted skills and sigmas against `scipy.optimize` on one small case, as a **test-only** dependency — scipy stays out of the runtime deps.

If the constants keep accreting justifying comments over the next few PRs, the alternative worth costing is block-coordinate ascent (Zermelo/MM for the skills given sigma, 1-D Newton on each tau given the skills) — each block monotone and hyperparameter-free, though still no global guarantee on the joint problem. Not asking for that here.

currentlycodinng (changes requested):
`ranking.py` is genuinely good work. I derived the gradients by hand against the code and they are correct, including the chain rule through `sigma_k = exp(tau_k)`. The design notes are the most useful part of this PR: the gradient-stationarity floor rationale, the story about the hard geometric-mean projection being a multiplicative ratchet, and the decision to measure convergence on iterate movement rather than gradient norm are all things a reviewer would otherwise have to rediscover. I also checked both citations. arXiv:2602.16610 does introduce a per-judge discriminator on top of Bradley-Terry and does use cycle consistency as its independent diagnostic, and arXiv:2601.21817 does prove identifiability up to the natural normalisations plus consistency and asymptotic normality of the MLE. The descriptions are accurate.

The ranking module is not what I want to talk about. The problem is what happens when `bt_sigma_aggregation` collapses it to two items.

## In the pairwise integration, sigma stops measuring reliability

`bt_sigma_aggregation` fits over exactly two items, `'A'` and `'B'`, with one record per decisive vote. With only two items there is a single skill gap `g`, so judge `k`'s entire contribution to the likelihood is its own marginal: the MLE puts `logistic(g / sigma_k) = (judge k's A-rate)`. Sigma is then a closed-form function of how one-sided that judge was, with no dependence on whether it agreed with anyone else on any particular comparison. I confirmed this numerically on four synthetic judges with fixed A-rates:

```
gap 0.1296
r90: sigma= 0.059   logistic(gap/sigma)=0.8991   own a_rate=0.9
r70: sigma= 0.154   logistic(gap/sigma)=0.6990   own a_rate=0.7
r50: sigma= 4.044   logistic(gap/sigma)=0.5080   own a_rate=0.5
r30: sigma=35.261   logistic(gap/sigma)=0.5009   own a_rate=0.3
```

Sigma reproduces each judge's own A-rate to three decimals. That is not internal consistency, it is decisiveness measured against the majority's direction, and it is asymmetric: `r30` is exactly as decisive as `r70` but lands on the wrong side of the fitted gap and gets 200x the sigma.

The consequence is that the most degenerate judge on the panel wins. Here is the failure through the real `build_report` entry point, three judges over ten comparisons, where `always_A` never picks B at all (the shape a position-biased or verbosity-biased small judge takes) while `y` and `z` agree with each other and prefer B on six of the ten:

```
plurality winners : ['B','B','B','B','B','B','A','A','A','A']
plurality  a/b    : 0.4 0.6

bt sigmas         : {'always_A': 0.02, 'y': 7.826, 'z': 7.826}
bt weights        : {'always_A': 50.29, 'y': 0.13, 'z': 0.13}
bt winners        : ['A','A','A','A','A','A','A','A','A','A']
bt  a/b           : 1.0 0.0
skill_gap         : 0.15   p_a_beats_b: 0.537
warnings          : []
```

One judge that never disagrees with itself, because it never says anything, outvotes an agreeing majority by 50 to 0.26 on every single comparison, and the run flips from 60% B to 100% A with an empty `fit_warnings`. Note that the fit itself is fine and honest: `skill_gap` is 0.15 and `p_a_beats_b` is 0.537, correctly reporting a near-tie. It is the `1.0 / sigma` re-derivation of per-comparison winners that breaks, so the block reports a headline probability of 0.537 next to a weighted A-win rate of 1.0.

This is the case the description says is handled. The adversarial-judge test is described as majority-anchored specifically to avoid the ill-posed single-adversary-versus-single-good-judge setup, but the example above is majority-anchored, two against one, and it still inverts. Likewise the docs note says BT-sigma "protects against noisy judges, not coordinated ones", and here it fails against a lone one. The reason none of the tests catch it is that they all use judges whose A-rates are interior; a judge at exactly 0 or 1 is the pathological point, and it is also the judge you are most likely to see in production, since a small model with strong verbosity bias picking the longer answer every time is precisely a 100% one-sided judge.

I do not think the fix is complicated, and I would rather you pick it than me, but the options I see are capping the weight ratio (the tau prior already bounds sigma, just not tightly enough to matter at `_TAU_RIDGE = 1e-2`), refusing to weight a judge whose own decisive votes are unanimous and warning instead, or requiring `sigma` ordering to agree with `cycle_rate` before it is allowed to reweight anything. What I would not do is ship the current behaviour under a docstring that says smaller sigma means more reliable, because in the two-item case it means more one-sided.

To be clear about scope: none of this touches `fit_bt` itself. Over many items a judge's marginal does not pin its sigma, cross-pair consistency is real signal, and your synthetic recovery tests demonstrate that it works. The degeneracy is created by the two-item collapse in `bt_sigma_aggregation`, so that is where the guard belongs.

## "Runs in milliseconds" is off by three orders of magnitude at run scale

Both the docs and the description say the fit runs in milliseconds. Measured in the shape `bt_sigma_aggregation` actually produces (two items, three judges, one record per decisive vote), on this machine:

```
votes=  600   iters=1848   converged=True    0.36s
votes= 3000   iters=4216   converged=True    4.28s
votes= 9000   iters=7371   converged=True   22.21s
```

The inner loop is O(iterations x records) in pure Python and iterations grow with the data, so it is closer to quadratic in run size than to constant. A 200-comparison run over a three-judge panel is 600 votes and lands at a third of a second, which is fine. A 3000-comparison run is 22 seconds of synchronous CPU inside `build_report`, which is called from `PairwiseRun.build_report` with no await in sight. Worth either correcting the claim to something like "sub-second for typical runs, seconds for large ones", or collapsing identical records into weighted counts before the fit, which in the two-item case reduces the whole dataset to at most three rows per judge and makes the cost genuinely constant.

## Two smaller things

**The `_RIDGE` comment overclaims.** It says the ridge "shrinks only the MAGNITUDE of extreme gaps; orderings, and therefore rankings and weighted winners, are untouched." A Gaussian prior shrinks low-evidence items harder than well-covered ones, so with unbalanced coverage it can reorder two items whose unpenalised skills are close. It is a small effect at `1e-2` and it is the right regulariser to have, but the comment claims a guarantee the maths does not give, and someone will later rely on it.

**`cycle_rate` silently drops duplicate judgements.** The `prefers` dict is keyed on (judge, pair), so when a judge has rated the same pair more than once the last record wins and the earlier ones vanish from the diagnostic. That is a plausible thing to happen with repetitions enabled. Either aggregate the duplicates or say in the docstring that the last one is authoritative.

---

Requesting changes only for the weighting degeneracy. Everything else here is in good shape and I will approve as soon as a one-sided judge can no longer take over a run.

What I checked: checked out the branch, `uv sync --all-extras`, `pytest tests/unit/test_ranking_bt.py tests/unit/test_pairwise_bt_sigma.py` 22 passed, `ruff check src` and `ruff format --check src` clean, `basedpyright` 0 errors, and hand-verified the gradient derivations. The sigma closed form, the panel inversion and the timings above all came from running the code on this branch, not from reading it. Both arXiv references resolve to the papers the PR claims.

KarinaKKarinaK:
Addressed both reviews in 96d520dc (rebased onto main first).

**Amina, the blocking finding:**
- **One-sided judge takeover**: fixed with the guard you proposed (option two). A judge whose decisive votes are unanimous is excluded from the reliability weighting: it votes with the median weight of the remaining judges (uniform if none remain), its sigma is dropped from judge_sigmas so JudgeStats cannot present one-sidedness as reliability, and fit_warnings names it with the reason. Your exact scenario is now a regression test (test_one_sided_judge_cannot_take_over_the_run): the run stays 0.4/0.6 with the majority instead of flipping to 1.0/0.0. The heterogeneous fixture also had to change: its steady always-A judge was byte-identical to your always_A pathology, so it now votes B once and the flip behaviour is asserted on genuinely mixed evidence.
- **Milliseconds claim**: took your second option. bt_sigma_aggregation collapses identical votes into weighted JudgedComparison records before the fit (at most three per judge in the A/B setting), so the cost is flat in run size: your 9000-vote shape went from 22.2s measured to 0.04s here. A test pins that weighted records fit identically to repeated records. ranking.py and the docs now state the O(iterations x records) cost honestly and tell multi-item callers to collapse.
- **_RIDGE comment**: reworded, no ordering-invariance claim; it now says low-coverage items shrink harder and near-ties can in principle reorder.
- **cycle_rate duplicates**: aggregated by weighted mean preference with pair-frame canonicalisation, so an evenly split judge contributes a no-preference edge instead of last-one-wins. Tested including the reversed-frame duplicate case.

**Bauke:**
1. **Adam vs MM**: decision recorded in the ranking.py module docstring: plain BT should use Zermelo/MM, and that is the switch if the discriminator ever goes away, but the sigma exponent breaks the algebraic form the MM update relies on, which is why a first-order optimiser plus the documented constants is used. Your call if you want it litigated further.
2. **cycle_rate not in the aggregation**: reasoning now written down in the docs: two fixed items admit no 3-cycles, so the diagnostic only means something on the multi-item ranking path.
3. **converged**: BTSigmaAggregation.converged now propagates fit.converged; the cap path is tested through the aggregation (converged False plus the warning).
4. **Vote lookups**: took the type-level fix: PairwiseVote.vote is Literal['A','B','tie'] | None, enforced by pydantic at construction, with the _as_vote narrowing helper keeping the runtime guard at the build site.
5. **Float equality**: leaders are collected with math.isclose(rel_tol=1e-9).
6. **Dead API surface**: reliability() now computes the weights in bt_sigma_aggregation, and comparisons_per_judge() feeds a fit_warning when a judge has fewer than 5 decisive votes.
7. **Test gaps**: tie votes flow end-to-end (test_tie_votes_flow_through_the_aggregation), the cap-hit warning is pinned, and the fit is checked against an independently bisected stationarity root on a 4-vote case. That last test caught a real bug: a transient Adam momentum stall could satisfy the single-step movement criterion 4 percent away from the optimum, so convergence now requires 5 consecutive sub-tolerance steps.
8. **Example path**: fixed to examples/pairwise/.

Local CI: 3908 passed, ruff and format clean, basedpyright 0 errors, mkdocs build --strict and mermaid validation pass.

Baukebrenninkmeijer:
RES-1024 has been updated with the refreshed findings and plan.

Feedback for the remaining design question: the current BT-sigma fit collapses observations onto one global A/B latent pair. If rows contain different questions or response pairs, the fitted variance can mix task/response heterogeneity with judge noise. That interpretation is only defensible when each datapoint has repeated observations, such as repetitions per judge, or when the fit is scoped to comparable item pairs.

Per-judge repetitions seem feasible for a conventional jury, but it is still unclear whether the same design is practical for an LLM jury given cost and how we would preserve and canonicalize each repeated vote. I suggest keeping this as follow-up design work in RES-1024: preserve the canonical per-repetition votes and decide whether to fit within repeated item groups or use a hierarchical model. This PR keeps the reviewed safety guards and does not implement that follow-up yet.

Verification after rebase: 70 relevant unit tests passed; source Ruff and basedpyright passed.

Baukebrenninkmeijer (approved):

PR #114, opened 2026-08-06
fix(redteam): client-side retry for pipeline LLM calls (RES-832)

KarinaKKarinaK:
## Context (RES-832)

The ticket asked to consolidate async LLM client construction and retry/backoff into `evaluatorq.common`. Most of that already landed incrementally since the ticket was written:

- `common/llm_client.py` owns env precedence (injected client > `ORQ_API_KEY` + router suffix > `OPENAI_API_KEY`); red-team's `create_async_llm_client` and simulation's `build_simulation_client` are thin wrappers over it.
- `common/retry.py` holds the exponential-backoff handler (moved from `simulation/utils/retry.py` in RES-897); the originals are deleted.
- Env-precedence and retryable-classification tests exist (`tests/simulation/test_client_builder.py`, `tests/common/test_retry.py`).

## What this PR closes

The one remaining gap from the ticket's motivation ("red-team backends reimplement partial retry behavior or skip it entirely"): the pipeline's internal LLM calls had no client-side retry at all. They relied only on `retry_extra_body`, which is router-side, becomes `{}` on a plain OpenAI client, and never covers network errors or errors returned by the router itself.

## Changes

- `LLMConfig.client_retry()`: thin wrapper over `common.retry.with_retry` driven by the existing `retry_count` / `retry_on_codes` budget, so router-side and client-side retry follow one config.
- Applied at the six previously unprotected call sites: tool adaptation (`attack_generator`), objective generation, resource capability inference and tool classification (`capability_classifier`), and both adversarial generation paths (`orchestrator`).
- Timeouts stay per attempt: `asyncio.wait_for` sits inside the retried callable and `with_retry` never retries `asyncio.TimeoutError`, so a hung call fails exactly as before.
- Tests: `tests/redteam/test_client_retry.py` pins the retry budget, immediate re-raise on non-retryable statuses, and custom `retry_on_codes`.

## Verification

- `pytest tests/redteam tests/common -m 'not integration'`: 1682 passed
- `ruff check src` and `ruff format --check src`: clean
- `basedpyright`: 0 errors


KarinaKKarinaK:
Screenshot 2026-08-06 at 12 32 46 PM

Baukebrenninkmeijer:
## Adversarial review — 10 agents (`/pr-review-stolen` + `/hate`, sonnet)

Six review lenses (code, tests, error handling, comments, spec, minimality) plus four hate critics (strategy, quality, reuse, goal). Every load-bearing claim below was re-verified against the code by hand before posting; agent claims that didn't survive verification are listed at the bottom as explicitly **not** findings.

Ratings: Skeptic 🟠 NEEDS REWORK · Pedant 🟠 NEEDS REWORK · Historian 🟢 MINOR ISSUES · Realist 🟠 NEEDS REWORK

---

### ⚖️ Decisions — need a call

**1. The premise "pipeline-internal LLM calls had no client-side retry at all" is not accurate — the OpenAI SDK was already retrying, and this PR makes it a third layer.**

Verified: `grep -rn "max_retries" src/` returns **zero hits**, and both `AsyncOpenAI(...)` constructions in `common/llm_client.py:147,164` omit it. So every client in this repo runs with the SDK default:

```
openai 2.32.0 → DEFAULT_MAX_RETRIES = 2
_base_client.py:791,796,801,806 → retries 408, 409, 429, >=500 + connection errors
_base_client.py:780-786 → honors `x-should-retry` and `Retry-After`
```

That's the same trigger set `common/retry.py` re-derives by hand (`_is_retryable_status`, `_RETRYABLE_NETWORK_ERRORS`). Consequences:

- **Attempt count multiplies.** `client_retry` gives `retry_count + 1` = 4 attempts, each of which is an SDK call that itself makes up to 3 HTTP attempts → **up to 12 HTTP requests per logical call**. On the Orq router path, `retry_extra_body` asks the router to retry 3× more inside each of those → ~36 upstream provider calls for one objective-generation. The docstring at `contracts.py:734-739` acknowledges the router×client stacking but not the SDK layer, which suggests it wasn't part of the analysis.
- **Backoff is now uncoordinated.** `with_retry` (`retry.py:93-97`) always uses its own `2/4/8s + jitter` curve and **ignores `Retry-After` entirely**; the SDK layer underneath honors it. Which curve applies depends on which layer catches the error first.
- **No wall-clock bound.** `attacker.timeout_ms` is per `client_retry` attempt (correct, and the comment about that is accurate), but nothing caps the combined outer×inner total. For a red-team run that already has `target_agent_timeout_ms` / `cleanup_timeout_ms` to bound wall clock, this is a run-time-predictability regression.

**The choice:** (a) set `max_retries` on the two `AsyncOpenAI(...)` constructions in `common/llm_client.py` from `retry_count` and delete `client_retry` + all 6 call-site edits — the SDK layer already covers 429/5xx/network with `Retry-After` support, `client.with_options(max_retries=N)` covers per-call override, and future call sites get it for free; or (b) keep `client_retry` but pass `max_retries=0` at construction so there is exactly one client-side layer and one backoff curve. What is not defensible is leaving three uncoordinated layers with no documented worst case.

**2. This is the opposite of what RES-832 asked for, at the layer level.** The ticket asked to *consolidate* retry into `evaluatorq.common`. This PR moves the retry decision back into application code at 6 hand-edited opt-in sites. The failure mode when a 7th call site forgets to wrap is silent, untested, reduced resiliency — no lint rule, no base-class hook, nothing. `common/llm_client.py` is already documented as the single source of truth for client construction; that's the seam.

**3. The problem RES-832's motivation actually names is still unfixed.** The motivation quoted in the PR body is that *red-team **backends** reimplement partial retry behavior or skip it entirely*. This PR touches only `redteam/adaptive/` — the pipeline's calls to *itself*. Verified still unprotected:

- `redteam/backends/openai.py:130-131` — `chat.completions.create` inside a bare `asyncio.wait_for`, no `with_retry`, no `client_retry`. This is the path for every `backend='openai'` run.
- `redteam/backends/orq.py:275,307` — `agents.responses.create` via bare `asyncio.to_thread`, zero retry (grep for `with_retry|client_retry|max_retries` in that file: no matches). Used for `TargetKind.DEPLOYMENT` and explicit `orq` backend.

Only `OrqResponsesTarget` gets retry, wired at `backends/registry.py:118-119` — and per the PR body that landed earlier. So a target under test that hits a transient 503 still fails its attack turn exactly as before. Closing RES-832 on this PR closes it on a miss in the exact files the ticket named. Either extend to the backends or amend the ticket to say the backend gap is deferred.

**4. The reuse target was `common/llm_call.py`, and it says so in its own docstring.**

```
Does NOT own the span (...), retry (caller wraps with `with_retry` if desired), or parsing/result-shaping.
```

`execute_chat_completion` / `execute_chat_parse` already own params assembly, `record_llm_input`/`record_llm_response`, W3C trace-header injection, `_apply_pipeline_metadata` (Orq trace filterability), and the `reasoning_effort` 400-drop compat shim. `common/judge.py:207,268,339` and `simulation/agents/base.py` use them. None of the 4 files in this diff do — they hand-build params and hand-call `record_llm_response`, and consequently still miss pipeline-metadata tagging and the reasoning-effort shim. Migrating the 6 sites onto `execute_chat_parse` + `with_retry` was the documented path and would have closed that gap in the same diff.

---

### ✅ Recommendations — proceed unless you object

| # | Fix | Where |
|---|-----|-------|
| 1 | `retry_count: int = 3` lacks bounds validation, unlike both its siblings (`max_content_filter_retries` and `max_target_retries` are `Field(..., ge=0, le=10)`). It previously only fed a request payload; now it feeds `max_attempts=retry_count + 1`, and `with_retry` raises `ValueError: max_attempts must be >= 1` (`retry.py:78-79`). So `LLMConfig(retry_count=-1)` passes pydantic and then blows up on the first LLM call with an error naming neither the field nor the config. Add `Field(default=3, ge=0, le=10)`. | `redteam/contracts.py:672` |
| 2 | Docstring claim is false: *"the same trade-off the target path already accepts."* The target path is `common/target_call.py::call_target_with_retry` — grep for `extra_body` there returns **zero matches**. It has a single client-side attempt loop (`target_call.py:145-151`), no router-side layer, so there is no stacking trade-off there to compare against. Delete or correct the clause. | `redteam/contracts.py:735-737` |
| 3 | Retries now run *inside* the spans, and `record_llm_response` fires once on the final response — a call that 503'd twice before succeeding is indistinguishable in the trace from a first-try success, except for a `loguru` line. Worse, `orq.redteam.adversarial_retry` (`orchestrator.py:688`) is set from the **content-filter** loop counter only, so the attribute now means two different things and one kind of retry is invisible. Span duration can balloon by the full backoff schedule with nothing in the trace explaining why. Record an attempt count on the enclosing span. | 6 call sites |
| 4 | The network-error path — the PR's stated motivation — has **no test anywhere in the repo**. `_is_retryable_error` (`retry.py:49-59`) identifies network errors by matching `type(err).__name__` and `type(err.__cause__).__name__` against a string tuple of httpx class names. That's version-fragile string matching on the SDK's wrapping behavior, gating the exact case this PR exists for, and `tests/common/test_retry.py` only covers `_is_retryable_status` while the new file only raises `APIStatusError`. Add an `APIConnectionError`-with-`__cause__` test. | `tests/redteam/test_client_retry.py` |
| 5 | No test proves the wrapper is actually wired at any of the 6 sites. A future refactor reverting any one of them to a direct `await llm_client...` passes the whole suite. One test per site, or one parametrized over them. | same |
| 6 | The `asyncio.TimeoutError`-is-never-retried guarantee is asserted in a code comment and in the PR body but exercised by no test. It's a one-liner and it's the safety claim the two `wait_for` sites depend on. | same |
| 7 | `_no_sleep` monkeypatches `retry_module.asyncio.sleep`, so nothing verifies the backoff curve, the `RETRY_MAX_WAIT_S=60` cap, or the jitter bound. Fine for speed, but the timing logic is then wholly untested — a `max`/`min` inversion at `retry.py:95` would ship silently. | `test_client_retry.py:27-34` |
| 8 | Label convention drift. Every existing `with_retry` caller uses `Class.method` shape — `simulation/agents/base.py:324` (`f'{self.name}._call_chat_completions'`), `openresponses/target.py:214` (`'OrqResponsesTarget._call_responses_api'`), `simulation/generators/first_message_generator.py:144`. The new sites use prose (`'Tool adaptation'`, `'Resource capability inference'`), and those two carry zero target/run context — the retry warnings they emit will be unattributable in logs. | 6 call sites |
| 9 | `client_retry` is a plain `def` returning `Awaitable[_T]`, so a caller who forgets `await` silently gets no retry and no error. Make it `async def ... return await with_retry(...)`; all 6 sites already `await` it. Also hoist the in-method `from evaluatorq.common.retry import with_retry` to module scope — there's no circularity to dodge (`common/retry.py` imports nothing from `redteam`), and the file already does module-level `from evaluatorq.common.target_call import ...` at line 23. | `redteam/contracts.py:726-745` |
| 10 | `retry_count + 1` is now computed in two places. `backends/registry.py:118-119` already does it with a comment explaining the +1. Make it an `LLMConfig.retry_attempts` property and have both call it. | `redteam/contracts.py:734` |
| 11 | Retry activity is invisible to users — grep for `retry` in `redteam/hooks.py` and `redteam/reports/` hits only `recommendations.py`. A run that takes 3× longer because of sustained 503s shows the user nothing. Surface an attempt/retry count in the report or `RichHooks`. | `hooks.py`, `reports/` |
| 12 | The `client_retry` docstring is 12 lines for 6 lines of code, and most of it argues for the design (router-vs-client, multiplicative retries). That belongs in the commit message. Also: the timeout comment exists at `orchestrator.py:444` but not at the structurally identical site at `:689` — classic comment-rot setup. Either duplicate it or hoist it onto `client_retry`. | `contracts.py:726-738`, `orchestrator.py:444/689` |

**In plain words:** the code works, but it adds a third retry loop on top of two that were already running, in the one layer the ticket asked you to move retry *out of* — and the thing the ticket was actually filed about (backends with no retry) is still broken.

---

### Checked and explicitly NOT findings

Agents raised these; verification killed them. Listing so nobody re-litigates:

- **Content-filter × client_retry does not multiply to 12 calls.** The two conditions are disjoint: a content-filter block arrives either as a 200 with `finish_reason='content_filter'` (never raises, so `client_retry` never sees it) or as a 400-class error (not in `_is_retryable_status`, so `client_retry` re-raises on attempt 1). A genuine 5xx exhausts `client_retry` and then propagates straight out of `regenerate_on_content_filter` (`content_filter.py:71-75`) without a `continue`. Real worst case from *these two* layers is 4, not 12. (The SDK layer in Decision 1 is a separate, real multiplier.)
- **No `lambda` late-binding bug at `orchestrator.py:690`.** The lambda captures `_messages`/`_timeout`/`_strategy`/`_turn`, which are eager-bound default parameters of `_generate_attack_turn` (`:655-661`, already commented as a B023 workaround), not the mutating loop variable.
- **`with_retry` genuinely never retries `asyncio.TimeoutError`.** `except (asyncio.TimeoutError, asyncio.CancelledError): raise` at `retry.py:86-87` precedes the generic handler, so clause ordering wins regardless of the 3.11+ `TimeoutError`/`OSError` aliasing. The `orchestrator.py:444` comment is accurate.
- **`retry_count` defaults to 3, not 0** — the fix is on by default.
- **`with_retry` never swallows** — it re-raises `last_error` after exhaustion and re-raises non-retryable errors immediately. The docstring at `retry.py:72-77` is accurate. (It does not log a distinct "exhausted after N attempts" line, only per-attempt warnings — minor.)
- **`retry_extra_body` really is a no-op on a plain client** — gated on `client_routes_through_orq` (`llm_client.py:66`) which checks for a `/v3/router` base URL.
- **The RES-832 consolidation itself did land** (`common/llm_client.py`, `common/retry.py`; `simulation/_client.py` and `simulation/utils/retry.py` deleted; one `AsyncOpenAI(` construction site) — but under RES-897, which is only `relatedTo` RES-832. Worth a closing comment on the ticket so the history reads honestly.

<sub>10 sonnet agents; claims hand-verified against `openai==2.32.0`, `common/retry.py`, `common/llm_call.py`, `common/target_call.py`, `common/content_filter.py`, and `redteam/backends/`.</sub>

Baukebrenninkmeijer (changes requested):
## Follow-up: requesting changes — the ticket's core goal (consolidation / DRY / reuse) is not achieved

Re-reviewed with consolidation and reuse as the only lens. Verdict: **not achieved.** RES-832 asked to *consolidate* async LLM client construction and retry/backoff into `evaluatorq.common`. This PR does the opposite for the retry half — it distributes retry into application code at 6 hand-edited call sites. Everything below is counted, not asserted.

### The approach already failed inside its own PR

There are **10 raw `chat.completions.create/parse` call sites in `src/evaluatorq/redteam/`**. `grep -c client_retry` per file:

| File | `client_retry` | `retry_extra_body` | any retry? |
|---|---|---|---|
| `adaptive/attack_generator.py` | ✅ | ✅ | yes |
| `adaptive/capability_classifier.py` (×2) | ✅ | ✅ | yes |
| `adaptive/objective_generator.py` | ✅ | ✅ | yes |
| `adaptive/orchestrator.py` (×2) | ✅ | ✅ | yes |
| **`adaptive/tool_chaining.py:160`** | **0** | ✅ | **no** |
| **`reports/recommendations.py:194`** | **0** | ✅ | **no** |
| **`runtime/jobs.py`** | **0** | 0 | **no** |
| **`backends/openai.py:130`** | **0** | 0 | **no** |

`tool_chaining.py` and `reports/recommendations.py` are the *same class of call* — pipeline-internal, same `LLMConfig`, already calling `cfg.retry_extra_body(...)`, i.e. already carrying the exact router-side-only gap this PR exists to close. They were missed anyway. That's the empirical argument against per-call-site opt-in: it didn't survive contact with its own PR. 4 of 10 sites left behind, and the failure mode is silent — no lint rule, no base-class hook, no type error.

### Five concrete DRY/reuse regressions

1. **Retry moved away from the one construction seam.** `common/llm_client.py` is documented as the single source of truth for client construction, and it holds the only two `AsyncOpenAI(...)` calls in the repo (`:147`, `:164`). Neither passes `max_retries`. That is the seam; the PR routes around it.

2. **The SDK's own retry was never audited, configured, or disabled — so `common/retry.py` is now a hand-rolled duplicate of it.** Verified: `openai==2.32.0`, `DEFAULT_MAX_RETRIES = 2`, and `_base_client.py:791,796,801,806` retries 408/409/429/≥500 plus connection errors — the same trigger set `_is_retryable_status` / `_RETRYABLE_NETWORK_ERRORS` re-derives. The SDK version also honors `Retry-After` and `x-should-retry` (`:780-786`); `with_retry` ignores both and always applies its own `2/4/8s` curve. So this is duplicated logic where the duplicate is the weaker one, and the retryable-status list now lives in three places that must agree: `common/retry.py:22-39`, `LLMConfig.retry_on_codes` (`contracts.py:673`), and the SDK's internals.

3. **Third retry entry point.** Before: `with_retry` (called directly by `simulation/agents/base.py`, `openresponses/target.py`, `simulation/generators/first_message_generator.py`, `simulation/utils/structured_output.py`) and `call_target_with_retry` (`redteam/runner.py`, `redteam/adaptive/pipeline.py`, `redteam/adaptive/orchestrator.py`, `simulation/runner/simulation.py`). Now a third: `LLMConfig.client_retry`. Three ways to retry an LLM call in one package is the opposite of the ticket's deliverable.

4. **`retry_count + 1` duplicated.** `backends/registry.py:118-119` already owns this conversion *with a comment explaining the +1*. `contracts.py:734` reimplements it inline. Should be one `LLMConfig.retry_attempts` property.

5. **`common/llm_call.py` bypassed.** Its module docstring describes exactly this caller ("Does NOT own ... retry (caller wraps with `with_retry` if desired)"), and `common/judge.py:207,268,339` + `simulation/agents/base.py:282` already consume `execute_chat_completion`/`execute_chat_parse`. The 4 files in this diff don't — they keep hand-built params and hand-called `record_llm_response` at every site. Migrating them onto `execute_chat_parse` + `with_retry` was the documented path and would have deleted the duplication instead of decorating it.

Also worth naming: `client_retry` puts an IO-executing method on `redteam/contracts.py`, whose own docstring says *"This module defines all shared **data models**"*. `retry_extra_body` next to it is pure dict-building; this one awaits arbitrary callables and drives `asyncio.sleep`. Every other execute-with-retry helper in the package lives in `common/` or `backends/`.

### And the ticket's stated motivation is still unfixed

The motivation quoted in the PR body is that *red-team **backends** reimplement partial retry behavior or skip it entirely*. Verified still zero retry:

- `backends/openai.py:130-131` — `chat.completions.create` in a bare `asyncio.wait_for`. Every `backend='openai'` run.
- `backends/orq.py:275,307` — `agents.responses.create` via bare `asyncio.to_thread`.

So the files the ticket names are untouched, while the fix landed one layer above them.

### What would actually satisfy the ticket

One change, in the place the ticket named:

```python
# common/llm_client.py — the single construction seam
AsyncOpenAI(api_key=..., base_url=..., max_retries=retry_count)
```

That covers **all 10 redteam call sites** including the 4 this PR missed, plus `backends/openai.py`, plus every future site, with no opt-in to forget. `client.with_options(max_retries=N)` gives per-call override where a site genuinely needs a different budget. Then `client_retry`, the 6 call-site edits, the `_T`/`TYPE_CHECKING` churn, and the 93-line bespoke test all delete — retry gets tested once at the seam. Net diff goes near-zero and the ticket's actual deliverable lands.

If custom status-code logic beyond the SDK's is genuinely required, keep `with_retry` — but then pass `max_retries=0` at construction and apply it once at the seam, so there is exactly one client-side layer and one backoff curve, rather than two uncoordinated ones nested inside a third router-side one.

Either way the two backends need wiring, or RES-832 needs an explicit note that the backend gap is deferred to a follow-up.

### Correction to my earlier comment

I wrote that the 6 sites "still miss `_apply_pipeline_metadata` (Orq trace filterability)". That's wrong — `retry_extra_body` (`contracts.py:723`) already merges `pipeline_metadata_param()`, so router-path pipeline tagging is present at all 6 sites. Drop that from the list. The `reasoning_effort` compat-shim gap is real but marginal (`reasoning_effort` is only used in `simulation/`). The reuse argument in point 5 stands on the duplicated params/recording boilerplate alone.

The 12 recommendations in my previous comment (`retry_count` missing `ge=0`, the false "same trade-off the target path already accepts" clause, untested network-error path, span telemetry) are independent of this and still apply.

currentlycodinng (approved):
Straightforward and correct. I checked the timeout claim rather than take it on faith: `with_retry` re-raises `asyncio.TimeoutError` and `CancelledError` before it ever reaches the retryable check, so wrapping `asyncio.wait_for` on the inside really does keep a hung call bounded by a single timeout. That was the thing most likely to go wrong here and it does not.

Two observations, both about what happens after this lands rather than about correctness.

**The retry sits inside `with_llm_span`, so LLM span duration stops meaning model latency.** In `capability_classifier`, `objective_generator` and both orchestrator sites, the span opens first and `client_retry` runs inside it. With the defaults (`retry_count=3`) a span can now cover four attempts plus 2s + 4s + 8s of backoff and jitter, so a span reading ~20s might be a healthy model behind two 503s. Nothing in the span says so: `record_llm_response` records the winning attempt, and `orq.redteam.adversarial_retry` only counts the outer content-filter loop, so transport retries leave no trace at all. An attempt-count attribute set on the span when `client_retry` retried would fix it cheaply. This matters more than usual right now given RES-985 is landing per-judge span attributes in parallel.

**`retry_count` is now doing double duty.** It already feeds `retry_extra_body` for router-side retry; this PR also makes it the client-side attempt budget. Someone tuning it to change router behaviour now silently changes client behaviour too, and in `objective_generator` the client budget nests inside `regenerate_on_content_filter` (`max_content_filter_retries` default 2, so 3 outer attempts). Worst case per objective generation is 3 x 4 = 12 calls before the router's own retries. The two layers key off disjoint failure classes in practice, so it should not multiply often, and the docstring is honest that they compound during sustained outages. But the coupling is worth a line in the `retry_count` field docstring so nobody discovers it from a bill.

**Nit:** the function-local `from evaluatorq.common.retry import with_retry` looks like circular-import avoidance. Worth saying so in a comment, otherwise someone will hoist it to module scope as a cleanup.

The branch is CONFLICTING with main and needs a rebase.

---

What I checked: checked out the branch, `pytest tests/redteam tests/common -m 'not integration'` 1682 passed, `ruff check src` clean, `basedpyright` 0 errors, and read `common/retry.py` to confirm the timeout and non-retryable-status behaviour the description claims. CI has not run on this PR beyond the `label` job.

KarinaKKarinaK:
## Review response: conflicts resolved, retry moved to the construction seam

Merged origin/main and addressed the adversarial review. Main had already migrated all six pipeline call sites onto `execute_chat_parse` + `with_llm_span` (the documented reuse path from Decision 4), so the conflict resolution takes main's call sites wholesale and deletes `LLMConfig.client_retry` and its test file. The review fixes land in `ce4094e3`; the merge is `336e7e57`.

### Decisions

**1. Three uncoordinated retry layers.** Took option (a). `client_retry` is gone; `LLMConfig.retry_count` now drives the OpenAI SDK's built-in retry via `resolve_llm_client(max_retries=...)`. One client-side layer, SDK-owned backoff with `Retry-After` support. The remaining client x router stacking on the Orq path is documented with its worst case on the retry fields in `redteam/contracts.py`. Injected clients are used exactly as the caller built them.

**2. Consolidation at the wrong layer.** Retry now lives at the client-construction seam in `common/llm_client.py`. Every construction path passes it (pipeline attacker, evaluator/recommendations/executive-summary, capability classification), and a future call site gets it for free instead of silently opting out.

**3. Backends were the actual ticket.** Both gaps closed. `backends/openai.py` builds its client through the same seam, so it inherits the SDK retry. `backends/orq.py` now constructs the Orq SDK client with a `RetryConfig` derived from `retry_count`/`retry_on_codes` (429/5xx plus connection errors, `Retry-After` honored, `status_codes_override` from config). One caveat: the Speakeasy SDK bounds retries by elapsed time rather than attempt count, so `retry_count` gates on/off and sizes the backoff window; the mapping is documented on `_orq_retry_config`.

**4. Reuse of `common/llm_call.py`.** Resolved by the merge itself: main's migration of the six call sites is taken as-is, which also closes the pipeline-metadata and reasoning-effort gaps the review flagged.

### Recommendations

| # | Status |
|---|---|
| 1 | Done: `retry_count` is `Field(default=3, ge=0, le=10)` |
| 2 | Moot: the false docstring died with `client_retry` |
| 3 | Deferred by design: retries now happen inside the SDK below the span, so a per-span attempt count is not observable without a custom transport. Same trade-off as before this PR (the SDK default was already retrying invisibly); happy to file a follow-up if we want a counting transport |
| 4 | Done: `test_network_error_with_cause_is_retried` pins the `APIConnectionError` + `httpx.ConnectError` `__cause__` path |
| 5 | Done: `tests/redteam/test_retry_wiring.py` pins `max_retries` plumbing at every construction path, the SDK-default fallback, the injected-client hands-off rule, and the Orq `RetryConfig` derivation |
| 6 | Done: `test_asyncio_timeout_is_never_retried` |
| 7 | Done: `test_backoff_curve_doubles_and_caps` (2/4/8/16/32/60 with the 60s cap) and `test_jitter_bounded_at_quarter_of_wait` |
| 8 | Moot: the prose labels died with the call sites |
| 9 | Moot: `client_retry` deleted |
| 10 | Done: `LLMConfig.retry_attempts` property; the registry's hand-rolled +1 now calls it |
| 11 | Deferred with 3, same observability constraint |
| 12 | Moot: docstring and both timeout comments are gone with the migration |

### Checks

`uv run pytest -m 'not integration'`: 3864 passed, 2 skipped. `ruff check src`, `ruff format --check src`, `basedpyright` (whole repo): clean.

Baukebrenninkmeijer (changes requested):
## Review update — pulled `ce4094e3`. The consolidation ask is satisfied; one line still reproduces the bug this PR exists to remove.

Re-verified against the new head. My blocking ground from the last round is resolved.

### What landed correctly

| Previous finding | Status |
|---|---|
| Retry distributed across 6 hand-edited call sites | **Fixed** — `grep -rn client_retry src/ tests/` → no hits. `LLMConfig.client_retry` deleted; retry now comes from the seam. |
| `common/llm_client.py` construction seam routed around | **Fixed** — `resolve_llm_client(max_retries=...)` (`:113`), threaded through `create_async_llm_client` and all four `runner.py` construction paths (`:990`, `:1023`, `:1311`, `:1898`). |
| 4 of 10 call sites missed (`tool_chaining`, `reports/recommendations`, `runtime/jobs`, `backends/openai`) | **Fixed structurally** — all get retry from the client they're handed, with nothing to opt into. This is the part that matters: the class of bug is gone, not just the four instances. |
| Hand-rolled duplicate of SDK retry, no `Retry-After` | **Fixed** — `max_retries` drives the SDK's own loop, which honors `Retry-After` / `x-should-retry`. |
| `retry_count + 1` duplicated | **Fixed** — `LLMConfig.retry_attempts`; `registry.py:140` consumes it. |
| `retry_count` unbounded | **Fixed** — `Field(default=3, ge=0, le=10)`. |
| `backends/orq.py` zero retry (ticket's named file) | **Addressed** — `_orq_retry_config` wires a Speakeasy `RetryConfig`. See defect 2 below on whether it takes effect. |
| Network-error / `TimeoutError` / backoff-curve untested | **Fixed** — `tests/common/test_retry.py` +112 lines, `tests/redteam/test_retry_wiring.py` +86. `uv run pytest tests/redteam/test_retry_wiring.py tests/common/test_retry.py -q` → 22 passed. `ruff check src` clean. |

RES-832's core deliverable is now genuinely met: one client-side retry layer, configured where clients are constructed, in `evaluatorq.common`.

---

### Defect 1 (blocking, one line) — the 12× stack is still live, just relocated to the default target path

`common/llm_client.py:126-131` now states the invariant:

> *"single client-side retry layer; callers must not add their own wrapper on top."*

The default red-team target path violates it. Traced:

1. `registry.py:140` → `retry_attempts = pipeline_config.retry_attempts` = `retry_count + 1` = **4**
2. → `OpenResponsesBackend(retry_attempts=...)` (`backends/openresponses.py:48,65`) → `OrqResponsesTarget`
3. `openresponses/target.py:82` builds its own client via `build_simulation_client(...)` → `openresponses/client.py:39` → `resolve_llm_client(...)` **with no `max_retries`** → SDK default `2` → **3** HTTP attempts
4. `openresponses/target.py:255` wraps it in `with_retry(max_attempts=4)`

**4 × 3 = up to 12 HTTP requests per target call**, with two uncoordinated backoff curves — the exact number and the exact shape of the defect this PR was written to remove. It's on `make_agent_backend`, i.e. the path a bare agent key takes, i.e. the common case.

One line clears it — `client.py:39` passes `max_retries=0` since this caller owns retry via `with_retry`:

```python
resolved = resolve_llm_client(..., max_retries=0)
```

or drop `with_retry` at `target.py:255` and pass `max_retries=retry_attempts - 1` instead. Either is fine; leaving both layers is not, by the PR's own rule.

Same class, pre-existing but now contradicting the new invariant, so worth a scoping decision rather than silence — `with_retry` over seam-built clients also at `simulation/agents/base.py:330,466`, `simulation/utils/structured_output.py:65,115`, `simulation/generators/first_message_generator.py:138`. With `MAX_RETRY_ATTEMPTS=5` that's up to 5 × 3 = 15 HTTP attempts. Either fix them the same way, or scope the docstring to say the no-wrapper rule applies only to clients built with an explicit `max_retries`.

### Defect 2 (non-blocking, but the orq fix is probably inert) — `max_elapsed_time` is total wall clock, not backoff budget

`_orq_retry_config` (`backends/orq.py:46-68`) sizes the window as backoff-only:

```python
backoff_window_ms = int(sum(min(500 * 2**i, 10_000) for i in range(retry_count)))  # 3500ms at retry_count=3
```

But Speakeasy measures from before the first attempt and compares total elapsed (`orq_ai_sdk/utils/retries.py:230-231`):

```python
start = round(time.time() * 1000)
...
now = round(time.time() * 1000)
if now - start > max_elapsed_time:
    raise
```

So `max_elapsed_time` includes request duration, which for LLM calls dominates backoff by one to two orders of magnitude (`target_agent_timeout_ms` defaults to 240 000 ms). Any call that takes longer than 3.5 s to fail — a router 503 behind a slow upstream, a connection error after a TCP timeout — blows the budget on the *first* failure and gets **zero retries**. The docstring's framing ("sized to cover roughly that many backoff rounds") is where the slip is: it's not a backoff budget.

Size it to include attempt latency, e.g. `backoff_window_ms + retry_count * timeout_ms`, and add a test that fails a slow call and asserts the attempt count — otherwise "the last path with zero retry" is still, in practice, zero.

### Nits

- `_orq_retry_config` sets `status_codes_override=[str(c) for c in retry_on_codes]` → exactly `429/500/502/503/504`. That *narrows* coverage: 507/520/529 are no longer retried, while the docstring one line above says "429/5xx". Use `['429', '5XX']` (Speakeasy accepts the pattern form) or fix the docstring.
- `_orq_retry_config` is defined at `orq.py:46`, above the module's remaining imports at `:69`. Ruff passes, but it reads as an accident — move it below the import block.

---

Leaving the review at *changes requested* for defect 1 only, since it's one line and it re-creates the PR's headline problem on the default path. Defect 2 and the nits are fine as follow-ups if you'd rather ship. The consolidation work itself is done well — this is a much better shape than the original.

src/evaluatorq/common/llm_client.py
```diff
@@ -125,6 +126,12 @@ def resolve_llm_client(
             ``extra_api_key``, or ``ORQ_API_KEY``). Used by ORQ-agent targets
             whose ``agent/<key>`` model id only resolves on the Orq router, so an
             ``OPENAI_API_KEY`` left in the env must never silently capture them.
+        max_retries: Client-side retry budget (number of *retries*, not attempts)
+            for clients built here. The OpenAI SDK retries 408/409/429/5xx and
+            connection errors with ``Retry-After`` support — this is the single
+            client-side retry layer; callers must not add their own wrapper on
```
Baukebrenninkmeijer: **Blocking, one line.** This invariant is correct and worth having, but the default red-team target path violates it today:

1. `registry.py:140` → `retry_attempts = retry_count + 1` = **4**
2. → `OpenResponsesBackend` → `OrqResponsesTarget`
3. `openresponses/target.py:82` builds its own client via `build_simulation_client(...)` → `openresponses/client.py:39` → `resolve_llm_client(...)` **with no `max_retries`** → SDK default `2` → **3** HTTP attempts
4. `openresponses/target.py:255` then wraps it in `with_retry(max_attempts=4)`

→ up to **12 HTTP requests per target call** with two uncoordinated backoff curves. Same number, same shape as the defect this PR removes from the pipeline path — just relocated to the target path, on `make_agent_backend` (bare agent key, i.e. the common case).

Fix at `openresponses/client.py:39`, since that caller owns retry via `with_retry`:

```python
resolved = resolve_llm_client(..., max_retries=0)
```

Or drop `with_retry` at `target.py:255` and pass `max_retries=retry_attempts - 1` instead. Either works; keeping both layers doesn't, by this docstring's own rule.

Separately, a scoping call: `with_retry` also wraps seam-built clients at `simulation/agents/base.py:330,466`, `simulation/utils/structured_output.py:65,115`, `simulation/generators/first_message_generator.py:138`. With `MAX_RETRY_ATTEMPTS = 5` that's up to 5 × 3 = 15 attempts. Either fix them the same way, or narrow this sentence to say the no-wrapper rule applies only to clients built with an explicit `max_retries`.

src/evaluatorq/redteam/backends/registry.py
```diff
@@ -124,8 +137,7 @@ def _create_openresponses_backend(
 
     instructions = target_config.system_prompt if target_config else None
     timeout_ms = pipeline_config.target_agent_timeout_ms if pipeline_config else None
-    # retry_count is the number of *retries*; OrqResponsesTarget wants total attempts (initial + retries), hence the +1.
-    retry_attempts = pipeline_config.retry_count + 1 if pipeline_config else None
+    retry_attempts = pipeline_config.retry_attempts if pipeline_config else None
```
Baukebrenninkmeijer, line 140: The `retry_attempts` dedupe itself is exactly right — good change. Flagging the consumer: this 4-attempt `with_retry` budget lands on `OrqResponsesTarget`, whose client already carries the SDK's own `max_retries` (default 2, because `openresponses/client.py:39` doesn't pass one). See my comment on `common/llm_client.py:132` — the two layers multiply to 12.

src/evaluatorq/redteam/backends/orq.py
```diff
@@ -43,6 +43,32 @@ def _get_orq_server_url() -> str:
...
+    from orq_ai_sdk.utils.retries import BackoffStrategy, RetryConfig
+
+    backoff_window_ms = int(sum(min(500 * 2**i, 10_000) for i in range(retry_count)))
+    return RetryConfig(
+        strategy='backoff',
+        backoff=BackoffStrategy(
+            initial_interval=500,
+            max_interval=10_000,
+            exponent=2,
+            max_elapsed_time=backoff_window_ms,
```
Baukebrenninkmeijer: **This makes the orq retry inert for most real failures.** `max_elapsed_time` is total wall clock from before the first attempt, not a backoff budget — `orq_ai_sdk/utils/retries.py:230-231`:

```python
start = round(time.time() * 1000)
...
now = round(time.time() * 1000)
if now - start > max_elapsed_time:
    raise
```

So request duration counts against the window. At `retry_count=3` the window is 3500 ms, while `target_agent_timeout_ms` defaults to 240 000 ms — so any call that takes longer than 3.5 s to fail (router 503 behind a slow upstream, connection error after a TCP timeout) blows the budget on the **first** failure and gets **zero retries**. That's the common failure shape for LLM calls, and it's the path the commit message describes as "previously the last path with zero retry".

Size it to include attempt latency, e.g. `backoff_window_ms + retry_count * self._timeout_ms`, and add a test that fails a slow call and asserts the attempt count — otherwise this reads as fixed while still being zero in practice.

The docstring above is where the slip originates: "the window is sized to cover roughly that many backoff rounds" — it isn't a backoff-only budget.

src/evaluatorq/redteam/backends/orq.py
```diff
@@ -43,6 +43,32 @@ def _get_orq_server_url() -> str:
...
+        ),
+        retry_connection_errors=True,
+        status_codes_override=[str(code) for code in retry_on_codes] if retry_on_codes else None,
```
Baukebrenninkmeijer: This *narrows* retry coverage rather than configuring it. With the `retry_on_codes` default the override becomes exactly `429/500/502/503/504`, so 507/520/529 are no longer retried — while the docstring at line 47 says "429/5xx". Speakeasy accepts the pattern form, so `['429', '5XX']` matches the documented intent; otherwise fix line 47 to list the five codes.

src/evaluatorq/redteam/backends/orq.py
```diff
@@ -43,6 +43,32 @@ def _get_orq_server_url() -> str:
     return url.rstrip('/').removesuffix('/v3/router')
 
 
+def _orq_retry_config(retry_count: int, retry_on_codes: list[int] | None = None) -> Any:
```
Baukebrenninkmeijer: Nit: this function sits above the module's remaining imports (`from evaluatorq.common.thread_context import ...` at line 72). Ruff passes, but it reads as an accident — move it below the import block.

Baukebrenninkmeijer (commented):
Line-anchored version of the addressable items from my review update, so they can be resolved one by one. Ordered blocking first. Nothing new here beyond what's in that comment — just attached to the code.

KarinaKKarinaK:
## Review resolved (dca15dcb)

**Defect 1 (blocking): fixed at the construction seam.** `OrqResponsesTarget` now builds its own client with `max_retries=0`, threaded through `build_simulation_client(max_retries=...)`: `with_retry` is the only client-side layer on the agent-target path, so the 4 x 3 stack is gone. An injected client is untouched, and both are pinned by `test_orq_responses_target_client_carries_no_sdk_retries` (self-built at 0, injected keeps whatever the caller set).

**The wider class:** took your offered scoping option for the legacy simulation sites. The `resolve_llm_client` docstring now says exactly which callers the invariant binds: explicit `max_retries` means no wrapper on top, a caller that owns retry passes `0`, and the simulation paths still wrapping SDK-default clients are named as predating the rule. Migrating those five sites touches shared clients whose other call sites do not wrap, so it wants its own change rather than a rider here.

**Defect 2: the window now budgets attempts, not just backoff.** `_orq_retry_config` takes `timeout_ms` and sizes `max_elapsed_time` as `retry_count * timeout_ms + backoff rounds`; both construction sites pass their per-call timeout. `test_orq_retry_window_survives_a_slow_first_failure` pins that a first failure taking the full timeout leaves budget to retry, which fails under the old 3.5s window. The docstring now describes Speakeasy's wall-clock semantics rather than calling it a backoff budget.

**Codes nit:** kept the override as the literal configured `retry_on_codes` and fixed the docstring instead. Reason: the SDK's own per-operation default is exactly the same five codes (`responses.py:266`), so the override narrows nothing against the SDK, and honoring an explicit `retry_on_codes=[429]` exactly seems like the right config semantics; `['429', '5XX']` would ignore a narrowed config.

**Placement nit:** `_orq_retry_config` moved below the module's import block.

Checks: 3866 passed, 2 skipped; ruff check + format clean; basedpyright 0 errors repo-wide.

currentlycodinng (changes requested):
Both of Bauke's defects are fixed, and I measured them rather than reading them.

**Defect 1, the stacked retry layers.** `openresponses/client.py` now takes `max_retries` and `target.py:82` passes 0, with the reason written down at the call site. I checked that this is the branch the default path actually takes: `make_agent_backend` passes `llm_client=None`, so `OrqResponsesTarget` builds its own client and `_client_owned` is True. On the head commit the target client reports `max_retries=0` and `with_retry` runs `max_attempts=4`, so one layer and four attempts where Bauke measured twelve.

**Defect 2, the retry window.** `_orq_retry_config` now adds `retry_count * timeout_ms` to the backoff sum, and the docstring states the thing that made the old version wrong, that Speakeasy measures `max_elapsed_time` from before the first attempt so request duration spends the budget. That is the fix, and writing down why is what stops it regressing.

**One thing the fix introduced, and it is the reason I would not merge as-is.**

Collapsing to `max_retries=0` hands all retry on that path to `with_retry`, but the two layers did not cover the same set of errors. The SDK retried anything the transport raised (`_base_client.py:1650-1664`, the bare `except Exception` with `remaining_retries > 0`). `with_retry` retries `APIStatusError` by status, plus a five-name allowlist of httpx class names matched through `__cause__`. Everything outside that list used to get 3 attempts and now gets 1. Counted at the transport, one target call:

```
  error                     max_retries=0 (this PR)   max_retries=2 (before)
  ConnectError                                    4                       12
  ReadTimeout                                     4                       12
  ReadError                                       1                        3
  RemoteProtocolError                             1                        3
  WriteError                                      1                        3
  ProxyError                                      1                        3
```

`RemoteProtocolError` is the one I would worry about, since "server disconnected without sending a response" is the ordinary way a long router call dies mid-flight, and it is exactly the transient the retry exists for. It arrives as `APIConnectionError` with the httpx error as `__cause__`, same as `ConnectError`, and the only thing separating them is that one name is in `_RETRYABLE_NETWORK_ERRORS` and the other is not.

Since the SDK raises `APIConnectionError` and `APITimeoutError` only for transport failures, treating those two classes as retryable directly would cover the whole set and make the allowlist unnecessary, rather than growing it by four names and leaving the next httpx error class in the same position. Whichever way it goes, it belongs in `test_retry.py` next to `test_network_error_with_cause_is_retried`, which currently pins `ConnectError` and is the test that would have caught this.

Two smaller notes, neither blocking. `retry_attempts` as a property on `LLMConfig` is a real improvement over the `+ 1` that used to sit in the registry with a comment explaining it, and the `ge=0, le=10` bound on `retry_count` means the multiplication comment above it now describes a bounded worst case instead of an open one. And the `max_retries` docstring in `llm_client.py` scoping the invariant to new callers, with the legacy simulation paths named as tracked rather than quietly contradicting it, is the honest version.

CI is green on all thirteen legs. Locally on `90b712e8`: `ruff check src` and `ruff format --check src` clean, `basedpyright` 0 errors, `pytest -m 'not integration'` 3914 passed, 2 skipped.

---

What I checked: fetched `90b712e8`, traced `make_agent_backend -> resolve_backend('openresponses', llm_client=None) -> OrqResponsesTarget` to confirm the `max_retries=0` branch is the one the agent-target path takes, read the constructed client's `max_retries` and the `with_retry` attempt count off a live target, read the Speakeasy retry loop and the OpenAI base client's error handling to establish what each layer retries, then drove one target call per httpx error class through a `MockTransport` and counted transport hits under `max_retries=0` and `max_retries=2`. Ran the full local gate set.

KarinaKKarinaK:
## Addressed the retryable-error narrowing (dd31aaed)

Good catch, and thanks for measuring it per httpx class — that made the gap unambiguous.

Fixed the way you recommended: retry on the SDK class directly rather than growing the name allowlist. `with_retry` now treats `APIConnectionError` (and its `APITimeoutError` subclass) as retryable, since the OpenAI SDK raises those only for transport-level failures. That covers the whole set at once — `ConnectError`, `ReadTimeout`, `ReadError`, `RemoteProtocolError`, `WriteError`, `ProxyError`, DNS, timeout — so `RemoteProtocolError` (the server-disconnected-mid-response case you flagged) now retries like the rest. The httpx name list stays only as a defensive fallback for a raw httpx error raised outside the SDK, documented as not the primary matcher.

Tests are in `test_retry.py` next to `test_network_error_with_cause_is_retried`: a `RemoteProtocolError`-caused `APIConnectionError` and an `APITimeoutError` both retry now. I also confirmed the old allowlist returned `False` for exactly those two (the regression), so the tests fail against the previous logic:

```
RemoteProtocolError-caused   OLD=False  NEW=True
APITimeoutError              OLD=False  NEW=True
```

The two non-blocking notes (the `retry_attempts` property and the scoped `max_retries` docstring) I'm leaving as-is since you were happy with them.

Checks: `ruff check src` + `ruff format --check src` clean, `basedpyright` 0 errors, `pytest -m 'not integration'` 3916 passed, 2 skipped.

Baukebrenninkmeijer (approved):
## Approving — pulled `dd31aaed`. All findings addressed, each pinned by a test.

Verified against the new head, not taken on the commit messages.

### Blocking defect 1 — one retry layer on the agent-target path ✅

`openresponses/target.py:82-87` now builds its client with `max_retries=0`, with the 4 × 3 = 12 number named in the comment; `build_simulation_client` gained the param (`client.py:16`). The 12-request stack on `make_agent_backend` is gone.

The `resolve_llm_client` docstring is also now honestly scoped rather than asserting a repo-wide invariant it couldn't back: explicit `max_retries` means no wrapper on top, a caller that owns retry passes `0`, and the legacy simulation paths that still wrap `with_retry` around SDK-default clients are **named as predating the rule and tracked to migrate**. That's the right resolution — the alternative was a docstring that reads true and isn't.

### Defect 2 — the orq retry window is no longer inert ✅

`max_elapsed_time = backoff_window_ms + retry_count * timeout_ms` (`orq.py:98`), with `timeout_ms` threaded from both call sites (`:626` from `self._timeout_ms`, `:657`). The docstring now states the actual Speakeasy semantics — *total wall clock from before the first attempt* — and says why a backoff-only window retried nothing. `test_orq_retry_window_survives_a_slow_first_failure` pins exactly the case I described: a failure consuming the full per-call timeout still leaves budget for a retry.

### Nits ✅

`_orq_retry_config` moved below the import block (now line 66, after the imports at 46-50). `status_codes_override` resolved in the other direction than I suggested — the docstring now says "the configured `retry_on_codes` (default 429/500/502/503/504, matching the SDK's own per-operation default)" instead of claiming 5xx. Fine: it's accurate, and it no longer narrows anything relative to the SDK's own default.

### The unprompted fix in `dd31aaed` is the best change in the PR

Replacing the five-name httpx allowlist with `isinstance(err, APIConnectionError)` (`retry.py:57-60`) is a better fix than the test I asked for in round one. My finding was that the allowlist was version-fragile string matching gating the PR's whole motivation; the actual bug underneath was worse — `RemoteProtocolError` ("server disconnected without sending a response"), the ordinary way a long router call dies mid-flight, was **not in the list**, so the single most likely transient failure on this path was silently taking one attempt. Matching on the SDK class covers the whole transport set at once and can't fall behind httpx. Keeping the name list as an explicit defensive fallback for raw httpx errors raised outside the SDK, with a comment saying it is not the primary matcher, is the right shape.

`test_remote_protocol_error_is_retried` and `test_api_timeout_error_is_retried` pin it. Checked the interaction: `APITimeoutError` is retryable (SDK transport timeout) while `asyncio.TimeoutError` from a `wait_for` is not (`test_asyncio_timeout_is_never_retried`) — coherent, and the per-attempt `wait_for` still bounds worst-case latency on the target path.

### Local verification

```
uv run ruff check src          → All checks passed!
uv run ruff format --check src → 214 files already formatted
uv run basedpyright            → 0 errors, 0 warnings, 0 notes
uv run pytest -m 'not integration' -q → 3916 passed, 2 skipped
```

(The `fasthtml` collection error I hit first was a stale local env from the dashboard work on main, not this branch — clean after `uv sync --all-extras --all-groups`.)

### Where RES-832 lands

The ticket's deliverable is met, and met at the layer it asked for. Retry is configured once where clients are constructed in `evaluatorq.common`; `LLMConfig.client_retry` and the six hand-wrapped call sites are gone; the four sites the first pass missed are covered structurally rather than individually; the two backends the ticket's motivation named (`backends/openai.py` via the seam, `backends/orq.py` via `RetryConfig`) have retry for the first time; `retry_count + 1` lives in one property; and the retryable-error predicate no longer depends on enumerating another library's class names.

Worth a line in the ticket that the client-construction half landed under RES-897 and this PR closed the retry half, so the history reads honestly — but that's ticket hygiene, not code.

Nice work turning this around across three rounds.

currentlycodinng (approved):

PR #115, opened 2026-08-06
feat(jury): CyclicJudge round-robin judge assignment (RES-1023)

KarinaKKarinaK:
## What (RES-1023)

Adds CyclicJudge ([arXiv:2603.01865](https://arxiv.org/abs/2603.01865)) as a judge **assignment** strategy: `assignment='all' | 'cyclic'` on both `llm_jury` and `llm_jury_pairwise`. Default stays `'all'` (every judge on every item, unchanged behavior). `'cyclic'` gives each item exactly one judge, rotating through the panel.

## Why

The paper's variance decomposition (generalizability theory) shows round-robin assignment recovers the panel-mean score exactly in expectation while costing the same as single-judge evaluation, and achieves lower variance than both all-judges and random-single-judge at every budget. It is an allocation strategy, so it composes with (rather than competes against) vote-level aggregations.

## Methodology fidelity

- **Balanced design**: rotation is over the deduplicated panel, advanced synchronously at call entry, so shares stay exactly balanced even under concurrent scoring (pinned by a gather test).
- **One generation per item**: we cycle judges across items, which the paper states achieves the same offset cancellation as cycling across generations. Its caveat about shuffling the dataset first (so the cycle cannot line up with a latent ordering) is documented in both the docstrings and the docs.
- **Position bias**: in pairwise mode the assigned judge still runs both orderings under `swap`, so per-judge symmetrization is preserved.
- **Scope honesty**: docs state per-item verdicts are single-judge opinions; cyclic is for run-level scores. The paper's own limitation (cancels within-panel offsets, not bias shared by the whole panel) stays covered by panel diversity.

## Guardrails

- `min_successful_judges > 1` with cyclic raises (only one judge runs per item).
- Unknown `assignment` values raise.
- A mechanically failed item degrades to inconclusive (run-level redundancy exists), unless `replacement_judges` provides a stand-in; the lone-judge propagate-errors behavior is unchanged.

## Tests

8 new tests in `tests/unit/test_llm_jury_cyclic.py`: strict rotation order, exact balance under concurrency, dedup rotation, single-model degeneration, pairwise one-judge-per-pair with swap intact plus report rollup, and validation errors. Full unit+common suite: 704 passed. `ruff`, `basedpyright`, and `mkdocs build --strict` clean.


Baukebrenninkmeijer:
## Review — 8 agents (general quality, tests, silent failures, comment accuracy, minimality, plus 4 adversarial critics)

Mechanism is correct and the tests genuinely pin the balance guarantee. Findings below are ordered by how many independent reviewers raised them. Verified positives first, so they don't get re-litigated: **arXiv:2603.01865 is real** ("CyclicJudge: Mitigating Judge Bias Efficiently in LLM-based Evaluation", Zhu et al., v3 4 May 2026) and the paper does define round-robin judge→scenario assignment. The "increment before any await" claim is **true** — `next(cycle)` runs before the first `await` in both `scorer()` and `compare()`, and evaluation concurrency is single-threaded asyncio, so share balance holds. Pairwise correctly passes `deduped` and `swap` symmetrization survives.

---

### 1. Docs promise index-determinism the code can't deliver (2 reviewers)

`docs/llm-as-a-jury.md:113` — *"Datapoint 0 goes to judge 0, datapoint 1 to judge 1, and so on."*

The cursor advances in **call-arrival order**, not datapoint-index order. `ScorerParameter` (`types.py:131`) carries no index. `evaluate()` runs datapoints under `asyncio.Semaphore(parallelism)` (`evaluatorq.py:251`); once `parallelism` tasks are in flight, whichever job's inference finishes first reaches the scorer next. So at `parallelism > 1` the item→judge mapping is a function of provider latency jitter, and two identical runs assign different judges to different items. Only the *shares* are guaranteed, never the mapping.

The code comment hedges this correctly ("arrival order"); the public doc does not. Two fixes, pick one:

- **(a)** Thread a stable index into the scorer and do `panel[i % len(panel)]`. Stateless, deterministic, deletes the `itertools.count()`, the closure state, and the whole concurrency-safety argument. Strictly simpler than what's shipped.
- **(b)** Rewrite the doc: balance is guaranteed, mapping is not; index order only holds at `parallelism=1`.

### 2. Cursor never resets between runs (2 reviewers)

`llm_jury.py:460`, `llm_jury.py:585` — the cursor is created at *evaluator construction* time and lives forever. The module's own docstring (`:467`) encourages module-scope construction. So:

```python
jury = llm_jury(judges=["a","b","c"], assignment="cyclic")   # module scope
await evaluatorq(..., data=ten_items)    # ends mid-cycle
await evaluatorq(..., data=five_items)   # does NOT start at judge "a"
```

Second run's starting judge depends on the first run's length. Nothing documents this, no `reset()`, no test. A notebook re-running a cell or a CI job looping over folds gets silently different judge distributions. Fix (a) above kills this too.

### 3. `repetitions > 1` silently breaks the "single-judge cost" claim (2 reviewers)

`_validate_assignment` (`llm_jury.py:293`) guards `min_successful_judges` but not `repetitions`. `llm_jury(..., assignment="cyclic", repetitions=5)` calls the one assigned judge 5× (`common/jury.py:269`), costing 5× while the docstring (`:388-395`) and docs both promise "exactly one judge call". Untested, undocumented, unguarded. Either reject the combination, or document it as intra-judge self-consistency and drop the unqualified cost claim.

### 4. The failure path — the feature's main new risk — is entirely untested (3 reviewers)

`propagate_errors=(len(deduped) == 1 and not replacement_judges)` (`llm_jury.py:520`) is computed from the **full deduped panel**, while `run_panel` is always length 1 under cyclic. That asymmetry is the design, and it's the single most consequential behavior this PR introduces (silent degrade vs. hard failure) — and `test_llm_jury_cyclic.py` never raises from a judge. Three missing tests:

1. Multi-judge cyclic panel, one judge raises → item comes back inconclusive, no exception.
2. Same, with `replacement_judges` set → stand-in produces a decisive vote. (This path is *documented* at `:392` and has zero coverage.)
3. Single-model cyclic panel with a raising judge → still propagates loudly.

Related: a systematically broken judge (bad key for one provider, template that only that model chokes on) is diluted to exactly 1/N of the run and reported as scattered per-item "inconclusive". Under `assignment="all"` the same outage tanks *every* item's agreement rate — a loud aggregate signal. Under cyclic it's invisible in aggregate.

### 5. Judge identity is computed, then thrown away (3 reviewers)

`run_jury` knows the assigned judge (`JuryVote.model`, `contracts.py:535`), but `_to_evaluation_result` (`llm_jury.py:198-241`) never sets `raw_output`. The repo already has the convention for exactly this — `JURY_RAW_OUTPUT_KEY` (`contracts.py:522`), used by `redteam/frameworks/owasp/evaluatorq_bridge.py:363` and `redteam/adaptive/pipeline.py:753`. `llm_jury()` ignores it.

Pre-existing gap, but cyclic makes it materially worse: with one hidden-cursor judge per item there is now **no way to recover which judge scored what**. Can't audit the rotation, can't spot a stuck judge, can't rebuild per-judge distributions. Meanwhile `append_jury_summary` emits `[jury: 1/1 judges, raw agreement 100%]` on every cyclic item — indistinguishable from a plain single-judge evaluator, and reading as "full consensus" when no consensus was computed. Suggest `[jury: 1/1 (cyclic, panel=3), model=m2]` and stashing the `JuryResult` under `raw_output['jury']`.

Note the asymmetry: pairwise users *can* verify the claim (`PairwiseReport.per_judge` → `JudgeStats`). Pointwise users — the mode every doc example uses — cannot. The statistical justification for the whole feature is unverifiable in its primary API.

### 6. Two idioms for the same branch (2 reviewers)

`llm_jury.py:460` gates on `cycle is not None`; `PairwiseComparator` (`:585`, `:645`) unconditionally builds `self._cycle` and gates on `self._assignment == 'cyclic'`. Same decision, two shapes, same file, same PR — and the slicing expression is duplicated verbatim. One shared `_cyclic_panel(panel, cycle)` helper, one idiom.

---

### Solo flags (one reviewer each, lower confidence)

- **Duplicate-judge semantics diverge across modes.** `'all'` runs the raw `panel` (so `judges=["m1","m1","m2"]` gives m1 double weight); `'cyclic'` runs `deduped`. Pre-existing wart, but this PR is where the two interpretations now coexist behind one flag.
- **`llm_jury_pairwise` docstring (`:692-699`) omits the degrade-to-inconclusive caveat** that `llm_jury`'s (`:392`) states, despite identical mechanics.
- **Doc claim is a loose paraphrase.** "Beats both all-judges and random-single-judge at any fixed budget" conflates two axes. The paper: matches all-judges' bias cancellation at 1/k cost; beats random-single-judge on variance at equal cost.
- **Degradation comment reads as cyclic-specific but isn't** (`:517-519`) — `propagate_errors` is assignment-agnostic; a multi-judge `'all'` panel behaves identically.
- **RES-1023's "survey competing papers, document the comparison in the ticket" requirement** isn't visible in the ticket, PR, or docs. May be tracked elsewhere — worth confirming before closing.
- **`test_cyclic_balances_exactly_under_concurrency`** can't fail: `_fake_run_judge` never awaits, so no coroutine ever yields. It would still pass if someone moved `next(cycle)` after an await. Add an `await asyncio.sleep(0)` to the mock to give the test teeth.

### Not a concern

Diff hygiene is clean, no scope creep. Complexity is well-restrained — no pluggable `AssignmentStrategy` registry, no speculative `cyclic_seed`/`shuffle` knobs, shuffling correctly left to the caller. `itertools.count()` is right-sized, not reinvented.

---

**Blocking, in my read:** #1 (doc states something false) and #4 (the new risk path is untested). #2 and #3 are correctness bugs users will hit. Fix (a) in #1 collapses #1, #2, and #6 into one smaller change.

src/evaluatorq/llm_jury.py
```diff
@@ -466,9 +500,12 @@ async def judge_fn(judge_model: str) -> Prediction:
                 extra_kwargs=extra_kwargs,
             )
 
+        # The increment happens before any await, so concurrent scorer calls
+        # still hand out exactly balanced judge shares.
+        run_panel = [deduped[next(cycle) % len(deduped)]] if cycle is not None else panel
```
Baukebrenninkmeijer: With `run_panel` reduced to one judge, every conclusive item has exactly one vote — and `_agreement_rate` in `common/jury.py` is `max(counts.values()) / len(values)`, so a single categorical vote is always `1.0`. `_jury_stats` likewise gives `std=0.0`, and `append_jury_summary` renders `[jury: 1/1 judges, raw agreement 100%]`.

A cyclic run therefore reports perfect agreement and zero variance on every single item, indistinguishable from a real unanimous panel. That's the one number a reader is most likely to trust without checking how the run was configured, and it's the one number cyclic assignment cannot produce.

`build_report` in `pairwise.py` already has the right shape for this — it returns `None` for `mean_agreement` when nothing was decisive by ≥2 votes. Mirror it: null out `raw_agreement` and `stats` (or gate them on `len(run_panel) >= 2`) when the panel for that item is one judge.

src/evaluatorq/llm_jury.py
```diff
@@ -424,6 +450,14 @@ def llm_jury(
...
         judges=judges, model=model, min_successful_judges=min_successful_judges
     )
+    _validate_assignment(assignment, min_successful_judges=min_successful_judges)
+    # Round-robin assignment (CyclicJudge, arXiv:2603.01865): each datapoint is
+    # scored by exactly one judge, cycling through the deduplicated panel in
+    # arrival order. Every judge covers an equal share of the run, so panel-
+    # relative judge bias cancels in expectation at single-judge cost. Shuffle
+    # the dataset first if its order is meaningful, so the cycle cannot line up
+    # with a latent grouping (the paper's own caveat).
+    cycle = itertools.count() if assignment == 'cyclic' else None
```
Baukebrenninkmeijer: The cursor is created when the evaluator is *built*, and `PairwiseComparator._cycle` (line 585) is likewise per-instance. So the documented "datapoint 0 → judge 0" guarantee holds only for a freshly constructed evaluator.

Build the jury once at module scope and pass it to two experiments — or hit a retry — and the second run starts wherever the first stopped. Judge shares stay balanced only if the run length happens to be a multiple of the panel size; otherwise each run is skewed and nothing surfaces it.

Either document that one `llm_jury(...)` / `llm_jury_pairwise(...)` call belongs to exactly one run, or key the cursor off a run identifier.

Small thing while you're here: `itertools.count()` plus `% len(deduped)` at the two call sites is what `itertools.cycle(deduped)` already does — `next(cycle)` returns the model directly and no call site needs the length.

src/evaluatorq/llm_jury.py
```diff
@@ -424,6 +450,14 @@ def llm_jury(
     panel, deduped = _resolve_and_validate_panel(
         judges=judges, model=model, min_successful_judges=min_successful_judges
     )
+    _validate_assignment(assignment, min_successful_judges=min_successful_judges)
+    # Round-robin assignment (CyclicJudge, arXiv:2603.01865): each datapoint is
+    # scored by exactly one judge, cycling through the deduplicated panel in
+    # arrival order. Every judge covers an equal share of the run, so panel-
```
Baukebrenninkmeijer: The comment says "cycling through the deduplicated panel", and that is what the code does — but `assignment="all"` scores with the raw `panel`, where a judge listed twice genuinely gets two votes' worth of weight. So a caller who duplicated a judge deliberately to weight it loses that weighting the moment they flip to `"cyclic"`, silently.

Worth one sentence in the public docstring and the docs table, not just this internal comment.

tests/unit/test_llm_jury_cyclic.py
```diff
@@ -0,0 +1,165 @@
...
+        name="x", criteria="c", judges=["m1", "m2", "m3"], assignment="cyclic", client=MagicMock()
+    )
+    calls: list[str] = []
+    with patch.object(llm_jury_mod, "run_judge", side_effect=_fake_run_judge(calls)):
+        await asyncio.gather(*(ev["scorer"]({"data": _datapoint(), "output": "x"}) for _ in range(9)))
+    # Interleaving may reorder, but shares stay exactly balanced.
+    assert Counter(calls) == {"m1": 3, "m2": 3, "m3": 3}
+
+
+@pytest.mark.asyncio
```
Baukebrenninkmeijer, line 125: Every fake judge in this file succeeds, so the PR's own error-handling promise is untested: `llm_jury.py:517-519` and `docs/llm-as-a-jury.md:144-145` both state that a mechanically failed cyclic item degrades to inconclusive, or uses a `replacement_judges` stand-in.

A regression that keyed `propagate_errors` off the full deduped panel instead of `run_panel`, or that failed to promote a replacement on a cyclic miss, would pass everything here.

One test: make `run_judge` raise for one model in a 3-judge cyclic panel, then assert (a) no exception escapes, (b) that item is inconclusive with no replacements configured, (c) a configured replacement is promoted and votes.

docs/llm-as-a-jury.md
```diff
@@ -109,11 +109,43 @@ of `labels`; omit it and the verdict is still recorded but `passed` is `None`.
...
 | `structured_output` | `True` | Use the provider's structured-output API; falls back to a schema-injected `json_object` call for models that reject it. |
 
+### Cyclic assignment (CyclicJudge)
+
+`assignment="cyclic"` implements CyclicJudge
+([arXiv:2603.01865](https://arxiv.org/abs/2603.01865)): round-robin judge
+assignment. Datapoint 0 goes to judge 0, datapoint 1 to judge 1, and so on,
+wrapping around the panel. Every judge covers an equal share of the run, so
+systematic judge bias cancels in expectation across the dataset while each
+datapoint costs exactly one judge call. The paper shows this beats both
```
Baukebrenninkmeijer: "each datapoint costs exactly one judge call" — with `repetitions > 1` it's one *judge*, still N calls. `exactly one judge` alone is accurate and leaves call count to `repetitions`.

Baukebrenninkmeijer (approved):

Clean implementation and unusually good docs — the "use it for run-level scores, not per-item trust" framing and the shuffle-your-dataset caveat are exactly the right warnings to put next to this feature. Approving.

The one thing worth fixing before merge: under `assignment="cyclic"` every item has exactly one vote, so `raw_agreement` comes back `1.0` and `stats.std` `0.0` on *every* item — byte-identical to a genuinely unanimous 3-judge verdict. Anything downstream that filters or ranks on those fields will read a cyclic run as uniformly confident. Details inline.

Also: the round-robin cursor lives on the evaluator object, not the run, so reusing one `llm_jury(...)` result across two runs silently breaks the balanced-share property for the second.

currentlycodinng (approved):
This one is in good shape. Small surface, default unchanged, the guardrails are the right ones, and the docs are honest about what cyclic buys and what it costs (single-judge per-item verdicts, shuffle first, run-level scores only). I checked the citation: arXiv:2603.01865 is "CyclicJudge: Mitigating Judge Bias Efficiently in LLM-based Evaluation" (Zhu, Tieleman, Bukhtiyarov, Chen), it does propose round-robin judge assignment, and it does build on a variance decomposition over scenario/generation/judge/residual components. The description represents it accurately.

Advancing the cursor synchronously before any `await` is the correct way to get balance under concurrent scoring, and `test_cyclic_balances_exactly_under_concurrency` pins it properly with a real `asyncio.gather`.

Three small things, none blocking:

**The rotation cursor is per-evaluator, not per-run.** `cycle = itertools.count()` lives in the `llm_jury(...)` closure and `self._cycle` in the comparator's `__init__`, so reusing one evaluator across two `evaluate()` calls starts the second run wherever the first left off. Exact balance then only holds when each run's item count is a multiple of the panel size. That is fine and probably what you want, but the docs say "datapoint 0 goes to judge 0, datapoint 1 to judge 1", which only describes the first run. One sentence in the docs would cover it. Same caveat if the framework ever retries a datapoint: the cursor advances and the retry draws a different judge.

**`cyclic` + `repetitions > 1` is undocumented.** It is legal today and means one judge answers the same item N times, which reduces that judge's call noise but gives none of the panel diversity the mode is for. Worth a line saying repetitions still applies to the single assigned judge, so nobody reads `repetitions=3` as "three judges".

**`panel` vs `deduped` changes what a repeated judge entry means.** `assignment="all"` runs the raw `panel`, so `judges=["a", "a", "b"]` queries `a` twice per item; `cyclic` rotates over `deduped`, so `a` gets half the items, not two thirds. That is the right choice and `test_cyclic_rotates_deduplicated_panel` covers it, but it is a silent semantic difference between the two modes that a reader of the docs table would not predict.

Also: the branch is CONFLICTING with main and needs a rebase.

---

What I checked: checked out the branch, `uv sync --all-extras`, `pytest tests/unit tests/common` 704 passed, `ruff check src` clean, `basedpyright` 0 errors. Note CI has not run on this PR beyond the `label` job, so the local run is the only signal. Also verified the arXiv reference resolves to the paper the docs claim it does.

KarinaKKarinaK:
Addressed both reviews in cbd0ddcb (rebased onto main first, so the branch no longer conflicts; the pairwise cyclic test was updated to main's RES-1194 replacements namespace as part of the rebase).

**Bauke:**
1. **Fake unanimity on cyclic items (the pre-merge fix)**: per-item `stats` and `raw_agreement` are now nulled whenever assignment is cyclic, mirroring build_report's None mean_agreement. The jury summary renders `raw agreement n/a` instead of a 100% that is byte-identical to a real unanimous panel. Pinned by a test that also asserts assignment=all still reports the real rate.
2. **Per-evaluator cursor**: documented in both docstrings and both docs pages: the rotation cursor lives on the evaluator/comparator, a reused one (second run, framework retry) continues where it left off, so the judge-0 start and exact balance hold per freshly built evaluator. I kept the cursor per-evaluator rather than keying it off a run id, since the scorer has no run identity in scope; the docs now say to build a fresh jury per run.
3. **itertools.cycle**: swapped in at both call sites, no modulo or length needed.
4. **panel vs deduped**: now in the public docstrings and both docs pages: duplicate judge entries up-weight a judge under all but get an equal share under cyclic.
5. **Error-path tests**: added the exact test you described, twice: a failing judge in a 3-judge cyclic panel degrades that item to inconclusive with no exception escaping, and a configured replacement is promoted and votes (calls asserted as m1, m2, r1, m3).
6. **Docs 'one judge call'**: now 'exactly one judge', with repetitions multiplying calls to that one judge.

**Amina:**
1. **Cursor per-evaluator plus retry caveat**: covered above, including the retry sentence.
2. **cyclic + repetitions**: documented in the docstrings and the docs practical notes: repetitions=3 is three calls to one judge, never three judges.
3. **panel vs deduped**: covered above.
4. **Rebase**: done, force-pushed with lease.

Local CI run after the changes: 3862 passed, ruff clean, ruff format clean, basedpyright 0 errors, mkdocs build --strict and mermaid validation pass.

Baukebrenninkmeijer:
## Re-review of `cbd0ddcb` — 2 of 7 closed, 1 went backwards

Pulled and re-ran. `tests/unit/test_llm_jury_cyclic.py` is green (11 passed). The `itertools.count()+modulo` → `itertools.cycle()` swap is a genuine simplification. The `stats`/`raw_agreement` nulling is the right call and I verified it end-to-end: both fields are already `JuryStats | None` / `float | None` (`contracts.py:792-798`), `append_jury_summary` already guards (`common/jury.py:206` → `'n/a'`), and **numeric mode is unaffected** — the score comes from `deliberation.verdict` (`common/jury.py:458,504` → `llm_jury.py:214-221`), never from `jury.stats`. No downstream reader breaks. Good fix, and the new test pins it.

Status per prior finding:

| # | Finding | Status |
|---|---------|--------|
| 1 | Doc claims index-determinism | ⚠️ **Reinforced** |
| 2 | Cursor never resets between runs | ✅ Documented — fair resolution |
| 3 | `repetitions>1` vs cost claim | ✅ Documented — fair resolution |
| 4 | Error-path tests | 🟡 2 of 3; the important one is missing |
| 5 | Judge identity / `raw_output` | ⬜ Untouched |
| 6 | Two idioms for one branch | ⬜ Untouched |
| 7 | Concurrency test can't fail | ⬜ Untouched |

---

### #1 went the wrong way

`docs/llm-as-a-jury.md:122` is unchanged: *"Datapoint 0 goes to judge 0, datapoint 1 to judge 1, and so on."* And the fix now asserts it twice more:

- `docs/llm-as-a-jury.md:139` — "A freshly built jury sends datapoint 0 to judge 0"
- `llm_jury.py:391-393` — "the documented 'datapoint 0 goes to judge 0' start and exact balance hold per freshly built evaluator"

The original problem wasn't cursor *reuse* — it was that `next(cycle)` fires in **scorer call-arrival order**, and `evaluate()` runs datapoints concurrently under `asyncio.Semaphore(parallelism)` (`evaluatorq.py:251`). At `parallelism > 1`, whichever job's inference returns first takes the next judge. A fresh evaluator does not fix that; datapoint 0 gets judge 0 only if it happens to finish first.

So the new prose scopes the guarantee to "per freshly built evaluator" — a condition that is neither necessary nor sufficient. Balance is guaranteed. The mapping is not, at any parallelism above 1. There is still no mention of `parallelism` anywhere in the file.

Two ways out, same as before:

- **(a)** Thread a datapoint index into the scorer, `panel[i % len(panel)]`. Makes the doc true, and deletes the cursor, the per-evaluator-lifetime caveat you just had to write, the "before any await" argument, and finding #6 — one change closes four.
- **(b)** Say plainly: *"Judges are assigned round-robin in scorer-arrival order. At `parallelism=1` that is dataset order; above it, only the share balance is guaranteed, not which judge sees which item."*

Either is fine. What isn't fine is the current state, where the doc makes a reproducibility promise the runtime doesn't keep.

### #4 — the missing test is the one that matters

Both added tests are good (`test_cyclic_failed_item_degrades_to_inconclusive:144`, `test_cyclic_failed_judge_promotes_replacement:168`). The third is still absent: **single-model cyclic panel where the judge raises must propagate, not degrade.**

`test_cyclic_single_model_panel_degenerates_to_single_judge:121` only asserts routing (`calls == ["only","only","only"]`) with a never-failing fake. So `propagate_errors=(len(deduped) == 1 and not replacement_judges)` at `llm_jury.py:533` has no test on either side of the branch under cyclic. That expression is the whole silent-failure contract of this feature — a regression flipping `deduped` to `run_panel` (always length 1) would make *every* cyclic run swallow judge outages into "inconclusive", and nothing would catch it. Ten lines.

### #5 — still no way to know which judge scored an item

`_to_evaluation_result` (`llm_jury.py:229-234`) still builds `EvaluationResult` from `value`/`explanation`/`pass`/`token_usage` only. No `raw_output`, so `JURY_RAW_OUTPUT_KEY` (`contracts.py:720`) — the convention `evaluatorq_bridge.py:324` and `redteam/adaptive/pipeline.py` already use — stays unused here.

Nulling `raw_agreement` correctly stopped cyclic items from *lying* about consensus. But the user still can't audit the rotation, can't spot a judge that's systematically stuck, can't rebuild per-judge distributions. Pairwise users can (`PairwiseReport.per_judge`); pointwise users — every doc example — can't. The paper's variance claim remains unverifiable in the primary API. Not addressed and not mentioned in the commit message; if it's deliberately deferred, say so and I'll drop it.

### #6, #7 — unchanged

`llm_jury.py:518` gates on `cycle is not None`; `llm_jury.py:700` gates on `self._assignment == 'cyclic'`, with `self._cycle = itertools.cycle(panel)` built unconditionally at `:627`. Same decision, two shapes. (`itertools.cycle([])` isn't reachable today — `_resolve_and_validate_panel` rejects empty panels before construction, and there's only one call site — so it's latent, not live.)

`_fake_run_judge:26` still never awaits, so `asyncio.gather` never actually interleaves and `test_cyclic_balances_exactly_under_concurrency` would still pass with `next(cycle)` moved after an await. One `await asyncio.sleep(0)` gives it teeth.

---

**My read:** #1 blocks — the docs assert something the code doesn't do, and this commit made the assertion more confident rather than less. #4's missing test blocks, cheaply. #5, #6, #7 are yours to weigh; none are correctness bugs today.

Option (a) on #1 is still the move I'd take — it retires #1, #2, #6, and the "cursor lives on the evaluator" caveat in a single stroke, and leaves less code than what's there now.

ChielToto (approved):
#### What worked / tested
- CI fully green: lint, basedpyright, `mkdocs build --strict`, and pytest across Python 3.10-3.13 plus the SDK-compat matrix. Author reports 704 passing.
- 8 focused tests in `test_llm_jury_cyclic.py` cover the load-bearing claims: strict rotation order, exact balance under `asyncio.gather`, dedup rotation, single-model degeneration, failed-item to inconclusive, replacement promotion, the stats-nulling, and pairwise one-judge-per-pair with swap intact plus report rollup.
- Verified the "advance before any await" concurrency claim: in `scorer`, everything before `next(cycle)` (client resolve, `_build_replacements`) is synchronous, so each task consumes its cursor slot before suspending at `await run_jury`. Balance holds; it's actually deterministic, not just balanced.
- Verified `PairwiseComparator.__init__` is keyword-only (`*,`), so the defaulted `assignment` sitting before non-defaulted params is legal, not a syntax trap.
- Docs are honest about the real gotchas: per-item verdicts are single-judge opinions, the rotation cursor is evaluator state that persists across runs and framework retries, and the shuffle-first caveat.

#### Nits
- The scoring path nulls `stats`/`raw_agreement` for cyclic items, but the pairwise path does not. Looks intentional (pairwise "agreement" is within-judge swap reconciliation, not cross-judge), but a one-line comment saying so would save the next reader the same double-take.
- Cursor state persists across runs, and a framework retry of a datapoint advances the cycle, so heavy retrying can drift balance by a slot per retry. Documented, just worth flagging as a real-world edge.
- Pairwise has no dedup-rotation test and no single-model-cyclic failure test; the scoring path has both. Coverage is asymmetric, not missing anything critical.
- `PairwiseComparator` builds `self._cycle = itertools.cycle(panel)` unconditionally even for `assignment='all'` (harmless, never consumed), while the scoring path guards it with `None`. Minor inconsistency.

#### Rating
Rating: 5/5. Clean, faithful to the paper, default behavior unchanged, honest docs on the sharp edges, and the two claims I distrusted (concurrency balance, signature legality) both hold up. Nits are polish, not fixes.

Baukebrenninkmeijer:
## Must-fix before merge

Condensing the two reviews above into just the items that need action. Everything not listed here (`raw_output`/judge provenance, the duplicated branch idiom, the toothless concurrency test) is worth doing but I'm not blocking on it.

### 1. `docs/llm-as-a-jury.md:122` states something the code doesn't do

> "Datapoint 0 goes to judge 0, datapoint 1 to judge 1, and so on."

`next(cycle)` fires in scorer **call-arrival** order. `evaluate()` runs datapoints concurrently under `asyncio.Semaphore(parallelism)` (`evaluatorq.py:251`), so at `parallelism > 1` whichever job's inference returns first takes the next judge. Balance is guaranteed; the item→judge mapping is not. `cbd0ddcb` restated the claim twice more (`docs:139`, `llm_jury.py:391-393`) scoped to "per freshly built evaluator" — but a fresh evaluator doesn't make it true either. `parallelism` is still not mentioned anywhere in the file.

Pick one:

**(a) Make the doc true.** Thread a datapoint index into the scorer, `panel[i % len(panel)]`. Deletes the cursor, the per-evaluator-lifetime caveat, the "advance before any await" argument, and the two-idiom duplication. Net less code than what's there now.

**(b) Make the doc honest.** Replace with: *"Judges are assigned round-robin in scorer-arrival order. At `parallelism=1` that is dataset order; above it only the share balance is guaranteed, not which judge sees which item."* Then drop the "datapoint 0 goes to judge 0" phrasing from `docs:139` and `llm_jury.py:391-393` too.

### 2. Missing test: single-model cyclic panel must propagate, not degrade

`propagate_errors=(len(deduped) == 1 and not replacement_judges)` (`llm_jury.py:533`) is the silent-failure contract of this feature, and under cyclic neither branch is tested. `test_cyclic_single_model_panel_degenerates_to_single_judge:121` only asserts routing with a never-failing fake.

Add: single-model panel + `assignment="cyclic"` + a `run_judge` that raises → assert the exception **propagates**. Roughly ten lines, mirroring `test_cyclic_failed_item_degrades_to_inconclusive:144`. Without it, a regression swapping `deduped` for `run_panel` (always length 1) would make every cyclic run swallow judge outages into "inconclusive" silently.

---

That's it from me. The `stats`/`raw_agreement` nulling and the two new error-path tests in `cbd0ddcb` are solid — I verified numeric mode still scores correctly through `deliberation.verdict` and that no downstream reader breaks on the `None`s.

KarinaKKarinaK:
## Must-fix items resolved, option (a) taken (8dd8daf5)

**Must-fix 1: the doc is now true, not just honest.** Went with option (a). `evaluatorq()` now passes each datapoint's dataset index into the scorer (`ScorerParameter` gained an optional `row` key, forwarded by `process_evaluator`), and the cyclic scorer selects `deduped[row % len(deduped)]`. So datapoint i goes to judge i % n at any `parallelism`, across evaluator reuse, with no cursor state on the runner path. Pinned two ways: out-of-order sequential rows (3, 1, 4 map to m1, m2, m2 — an arrival cursor would give m1, m2, m3), and nine concurrent shuffled rows where each result's judge is read back and checked against row % 3, which asserts the full mapping rather than just the balance.

One deliberate scope note on (a): the arrival-order cursor survives as the fallback for a direct scorer call (no runner, no row) and for `PairwiseComparator.compare`, which is user-driven per pair and has no index to key on. The docs now state both plainly, mention `parallelism` explicitly, and the misleading per-fresh-evaluator caveat is gone. So the cursor deletion is partial rather than total, but the mapping claim is now true everywhere the framework drives iteration, and only-balance is documented everywhere it does not.

**Must-fix 2: the propagate contract has its raising-side test.** Single-model cyclic panel, no stand-ins, judge raises: asserts `RuntimeError` propagates out of the scorer. A regression swapping `deduped` for the always-length-1 run panel now fails this test.

**From the non-blocking list, two taken:**
- **#5 judge provenance:** every pointwise result now carries the full jury record in `raw_output['jury']` (same `JURY_RAW_OUTPUT_KEY` convention as the redteam bridge), so which judge scored each item, its verdict, and the rotation are auditable from results alone. The parallelism-independence test reads its assertions through this field, so the provenance path is load-bearing in CI.
- **#7:** the test fakes now `await asyncio.sleep(0)`, so the concurrency tests genuinely interleave.

`#6` (two idioms) is reduced rather than gone: the runner path is index-based and stateless; the cursor remains only where no index exists.

Also merged current main into the branch (no conflicts).

Checks: 3914 passed, 2 skipped on the merged tree; ruff check + format clean; basedpyright 0 errors; `mkdocs build --strict` clean.

Baukebrenninkmeijer:
## Re-review of `8dd8daf5` — both blockers cleared

Option (a), done properly. Ran the CI checks verbatim: `ruff check`/`format --check` clean, **basedpyright 0 errors** (whole repo — the `ScorerParameter` change touches shared surface, so that mattered), 3262 passed. The three local failures are missing optional extras in my env (`huggingface_hub`, `fasthtml`), not this branch.

### Must-fix 1 — fixed, and I verified the index is actually what it claims

The concern with keying on a row index was that it'd be batch-local and quietly reintroduce the same bug. It isn't:

- Non-streaming: `evaluatorq.py:352` — plain global `enumerate` over the full list.
- Streaming/paginated: `evaluatorq.py:245` declares `datapoint_index = 0` **outside** the `async for batch in fetch_dataset_batches(...)` loop and increments per datapoint at `:294-297`. Not reset per page; `fetch_data.py:58-132` pages sequentially by cursor. Genuinely monotonic across the whole dataset.
- Passed as an explicit parameter through `process_data_point` → `process_job` → `process_evaluator` → `scorer_param['row']`, never a captured loop variable, so `asyncio.create_task` scheduling can't late-bind it.
- `NotRequired` comes from `typing_extensions` (`types.py:6`), already used at `:158`. 3.10-safe.

The `next(cycle)` fallback also can't interleave with the row path and skew balance — `process_job:186-193` always passes `row_index` as an `int`, so a call site is consistently one mode or the other, never mixed. `PairwiseComparator.compare` is the one genuine arrival-order path and it's documented as such rather than glossed.

Docs now say the true thing, and say it with the right scope. Good.

### Must-fix 2 — fixed

`propagate_errors=(len(deduped) == 1 and not replacement_judges)` now has a test on the raising side. Both branches of the contract are pinned.

### Bonus items you picked up

`raw_output[JURY_RAW_OUTPUT_KEY]` (#5) and the `await asyncio.sleep(0)` in the fakes (#7) — the concurrency tests now genuinely interleave instead of passing vacuously, and the new row-mapping tests read judge identity back out of `raw_output`, which is a nice use of the field they just added.

---

### Two small things, neither blocking

**1. `raw_output` is unconditional, but documented as cyclic-only.** `llm_jury.py:238` sets it for every assignment mode; the doc addition sits inside the cyclic bullet list and the comment says "under cyclic assignment this is the only record of the item→judge mapping". A reader would reasonably conclude it's cyclic-gated.

It's inert today — nothing in this repo reads it (no report, dashboard, or OTel path), and `send_results.py:119-122` already pops `raw_output` before platform upload. But under `assignment="all"` with a 5-judge panel it's ~1-11 KB of pure redundancy per result, and it's a behavior change for any external caller who used `raw_output is None` as a signal.

Either gate it on `assignment == 'cyclic'`, or drop a line saying it's unconditional. Prefer whichever, just make the two agree.

**2. Worth documenting: same datapoint gets the same judge across jobs.** `processings.py:65` passes the same `row_index` to every job, so in a multi-job run datapoint *i* draws judge *i % len(panel)* under all of them. That's the right behavior — it removes judge identity as a confound when comparing job A vs job B — but it's a real benefit that nobody will discover from the current docs.

One caveat that belongs next to it: for the platform-fetch path, cross-run reproducibility of the mapping rests on the Orq API returning stable pagination order. In-memory datasets are exactly stable; dataset-ID runs are stable only insofar as the API is.

### Still open from earlier, still non-blocking

`llm_jury.py` gates on `cycle is not None` while `PairwiseComparator` gates on `self._assignment == 'cyclic'` (#6). Cosmetic.

---

LGTM once you've made the docs and the `raw_output` gating agree. Nice iteration — the row-keyed version is both more correct and less code than the cursor it replaced.

KarinaKKarinaK:
Both non-blocking items from your re-review are in, 67a3663b:

1. **raw_output gating now agrees with the docs.** Took the gate option: the jury record is set only under assignment='cyclic', where it is the sole record of the item to judge mapping. Under 'all' the panel itself is the record, so raw_output stays None and keeps its pre-cyclic meaning for any external caller using it as a signal. The docs bullet now says exactly that, and both sides are pinned by tests (cyclic carries the record, 'all' has raw_output None).

2. **Cross-job stability documented.** New docs bullet: a multi-job run gives the same datapoint the same judge under every job since the mapping is row-keyed, which removes judge identity as a confound when comparing jobs, with the caveat that dataset-ID runs inherit whatever pagination stability the platform API provides while in-memory datasets are exactly stable.

Also closed the cosmetic #6 while in there: the scorer now gates on assignment == 'cyclic' like PairwiseComparator, with an unconditional lazy cursor (which is also what basedpyright wanted, since the Optional cycle stopped narrowing under the unified gate).

ruff clean, basedpyright 0 errors, 3915 passed.

PR #116, opened 2026-08-06
fix(dashboard): count legacy red-team runs in the landing aggregate (RES-1202)

KarinaKKarinaK:
## Problem (RES-1202, split out of the RES-1165 review)

`_redteam_row()` preserves a legacy report's `resistance_rate` when `summary.evaluated_attacks` is absent, but `landing()` aggregates the donut and headline from `evaluated_attacks` / `vulnerabilities_found`. A legacy report added zero to both, so its resistance silently dropped out of the aggregate while its individual row still showed a rate.

## Decision: derive, not exclude

The ticket asked for a decision between deriving the counts from `results` or excluding legacy runs with a footnote. This PR derives: the red-team overview loop already walks `results` with a per-result classification (errored results were never evaluated, the rest split on the `vulnerable` flag), so the real counts exist on disk and the aggregate can simply be correct. Exclusion remains only for the degenerate case of a legacy summary with no `results` at all — a rate without an attack count has no weight in an attack-weighted aggregate, and inventing one would be worse than leaving it out.

## Changes

- `_redteam_result_counts(data)` helper in `dashboard/metrics.py`, mirroring the overview loop's classification.
- `landing()`: when `evaluated_attacks` is present the summary fields are used exactly as before (no behavior change for modern reports); when absent, the derived counts feed `resistant` / `vulnerable`.
- Regression tests: a legacy-shaped fixture (summary with only `resistance_rate`, results with a safe/safe/vulnerable/errored mix) asserts the donut counts and headline rate include the legacy run's real numbers next to a modern report, plus the no-results degenerate case (contributes nothing, rate stays `None`, run still listed).

## Verification

- `pytest tests/dashboard`: 570 passed
- `ruff check src`, `ruff format --check src`: clean
- `basedpyright`: 0 errors

## Second deliberate change: evaluation-crash results no longer count as resistant

Declared explicitly so it can be reverted independently if resistance rates move more than expected (raised in review): the shared `_redteam_counts` classifier is also used by `redteam_overview()`, and it follows `converters._is_evaluated` / `_is_vulnerable` rather than the old overview loop. A result whose generation succeeded but whose evaluation never produced a boolean verdict (judge crash, timeout, skip: `error=None`, no boolean `evaluation.passed`) was previously booked as **evaluated + resistant** on the `/redteam` overview; it is now **excluded from the counts entirely**. Resistance rates for reports already on disk can therefore change on the overview page. The new rule is the security-correct one: an un-evaluated attack must not pad the resistant bucket. Pinned by `TestOutcomeClassification` in `tests/dashboard/test_redteam_overview.py`.

The same rule extends one schema generation older: a result with no `evaluation` object, no `error`, and no `vulnerable` key at all now classifies as unknown rather than resisted.

Legacy runs' recent-runs rows also derive their score through the same classifier, so a run's row and the donut above it cannot show two different rates.

## Verification (current)

- `pytest -m 'not integration'`: 3884 passed, 4 skipped
- `ruff check src`, `ruff format --check src`: clean
- `basedpyright` (whole repo): 0 errors


KarinaKKarinaK:
### Screenshot validating the fix (with example reports)
Screenshot 2026-08-06 at 1 49 13 PM

Baukebrenninkmeijer:
## Automated review — 9 agents (6 aspect reviewers + 4 adversarial critics)

Ticket alignment is clean: RES-1202 asked for a decision between *derive from `results`* and *exclude with a footnote*, and this PR implements the derive branch exactly as the ticket's own comment specifies, including the documented no-results degenerate case. Tests pass (14/14 in `tests/dashboard/test_landing.py`), `ruff check`/`format --check` clean, `basedpyright` 0 errors. No acceptance criterion is unmet.

The disagreement is about **scope**: three independent critics landed on the same finding, that the fix corrects one of three summary-derived aggregates in the same loop.

### Important — decide before merge

**1. `by_severity` still silently drops legacy runs, and the fix makes that visible (`metrics.py:274-280`)**

`landing()` reads `summary.get('by_severity')` unconditionally, with no results-derived fallback. The PR's own legacy fixture (`test_landing.py:56-67`) proves the point — its summary is `{'resistance_rate': ...}` only, so it lacks `by_severity` for exactly the same reason it lacks `evaluated_attacks`.

Before this PR both numbers were 0 for a legacy run, which was wrong but internally consistent. After it, the donut counts a legacy run's vulnerabilities while the severity chart does not, so the two panels on the same page now contradict each other. `redteam_overview()` (`metrics.py:674-696`) already shows this is trivial: it derives `critical` from `attack.severity` on `results`, in the same loop `_redteam_result_counts` now walks — the severity data is already in hand.

Raised independently by the strategy, goal-alignment, and quality reviewers.

**2. `token_usage_total` / cost has the same shape (`metrics.py:270-273`)**

Same unconditional summary-only read. A legacy run is counted in `total_runs`/`redteam_runs` and now in the donut, but contributes nothing to `total_tokens`/`rt_tokens`/`rt_cost`/`tokens_by_kind`. Untested either way — `_legacy_redteam_payload` never sets `token_usage_total` and no assertion touches token/cost for the legacy fixture.

If the answer to 1 and 2 is "out of scope, separate ticket," that is a defensible call — but it should be an explicit one, since the ticket's framing was "the landing aggregate," not "the donut."

**3. The classification is now implemented three times (`metrics.py:327-345`)**

`_redteam_result_counts` reimplements the skip-errored / split-on-`vulnerable` classification that already exists inline in `redteam_overview()`'s loop (`metrics.py:677-695`), and `_redteam_row`'s `all_errored` calc is a third variant. The docstring says it "mirrors" the overview loop — that is a textual mirror kept in sync by hand, and it is already not identical in scope (the overview loop also tracks `critical`). This is the same structural cause that produced RES-1202: two code paths independently deciding what the real numbers are for one JSON blob. A shared `_redteam_classify(data)` called from both is the durable fix.

### Suggestions

- `metrics.py:258` — `'evaluated_attacks' in summary` is a presence check, not a validity check. A summary with `'evaluated_attacks': None` takes the modern branch, `_as_int(None)` coerces to 0, and the fallback never runs. `summary.get('evaluated_attacks') is None` covers both shapes; partial/hand-edited legacy files are exactly where this bites.
- `metrics.py:267` — `ev, vu` are the only two-letter locals in the file, sitting next to `resistant`/`vulnerable` accumulators. `legacy_evaluated, legacy_vulnerable`.
- `metrics.py:264-266` + `327-336` — the same rationale is written twice, once as an inline comment and once as a 9-line docstring on a 6-line helper. Keep one; the other will go stale.
- `test_landing.py:69-75` — `_legacy_result`'s `attack`/`agent` keys are never read by `landing()`. Drop to `{'vulnerable': ..., 'error': ...}`.

### Test gaps

- No assertion that `data.severity` / `data.total_tokens` / `data.total_cost` behave as intended for the legacy fixture — the two gaps above are invisible precisely because nothing checks them.
- `_redteam_result_counts` has no direct unit test; the `error` key absent (vs `None`), `vulnerable` key absent, and non-list `results` shapes are untested — notable for a function whose whole purpose is tolerating older report shapes.
- No test for a transitional report carrying `by_severity` but not `evaluated_attacks`.

### Not a finding

Collapsing to "always derive from `results`" was considered and rejected: it removes the O(1) summary fast path for every modern run and would diverge from authoritative summary counts on reports whose `results` list was pruned. The branch is the right shape.

---

Baukebrenninkmeijer:
### Follow-up: the derived classification disagrees with the authoritative one

The last reviewer landed after my first comment with the sharpest finding, and I verified it against source.

`_redteam_result_counts` classifies with `not res.get('error')` → evaluated, `res.get('vulnerable')` → vulnerable. The authoritative definitions in `src/evaluatorq/redteam/reports/converters.py:670-677` are:

```python
def _is_evaluated(r): return r.evaluation is not None and isinstance(r.evaluation.passed, bool)
def _is_vulnerable(r): return r.evaluation is not None and r.evaluation.passed is False
```

Two things make the derived version diverge:

- `RedTeamResult.error` is `row.error` (`converters.py:359`) — the **target/generation** error only. An evaluation-stage failure (evaluator crash, timeout, skip) leaves `error=None`.
- `vulnerable` is computed at `converters.py:343` as `evaluation is not None and evaluation.passed is False`, which is `False` both when the agent resisted **and** when nothing was ever evaluated.

So a result whose generation succeeded but whose evaluation failed has `error=None, vulnerable=False` and the fallback silently books it as **evaluated + resistant**. `evaluated_attacks` from `compute_report_summary` would have excluded it. The derived counts therefore cannot reproduce what a modern summary would hold for the same data, and the bias is toward a rosier resistance rate — the wrong failure direction for a security metric, with no log line marking it.

Same for malformed/truncated records: a missing `vulnerable` key is indistinguishable from an explicit `False` and lands in the resistant bucket.

**Caveat on blame:** this is not a regression this PR introduces — `redteam_overview()`'s loop at `metrics.py:677-695` already classifies exactly this way, and the helper faithfully mirrors it as the PR says. But that mirroring is precisely what propagates the divergence into the landing aggregate, where it now feeds the headline number. Deriving `evaluation.passed` instead (`res.get('evaluation')` is a dict with a bool `passed`) matches the authoritative rule and is the same amount of code. Doing it in a shared helper would fix the overview loop at the same time — which folds into decision #3 in my earlier comment.

Also worth a `logger.debug` in the fallback branch (`metrics.py:263-269`) recording that a report was counted by derivation rather than its stored summary. Right now a legacy report's row can show one rate while the donut is built from a different rule, with no breadcrumb — and `CLAUDE.md` calls for loguru everywhere.

---

Baukebrenninkmeijer:
### Review applied — ddf675d4

**Decisions**

- **#1 severity — fixed.** Legacy runs now contribute to the severity bars, derived per-result from `attack.severity` on vulnerable results.
- **#2 tokens/cost — fixed.** Derived from both `execution.token_usage` and `evaluation.token_usage`, the same two holders `converters._aggregate_token_usage` walks. A legacy run no longer counts as a run whose spend is invisible.
- **#3 merge — done.** `redteam_overview()` and the landing fallback now share one `_redteam_counts(data)` classifier, so they cannot drift. Dropped the dead `run_vuln` accumulator that fell out of it.

**Classification fix** (the follow-up finding, folded into #3): `_result_outcome` now mirrors `converters._is_evaluated` / `_is_vulnerable` — only a boolean `evaluation.passed` counts as an evaluation. A result whose generation succeeded but whose judge crashed (`error=None, vulnerable=False`) is no longer booked as resistant. Reports too old to serialize `evaluation` at all still fall back to the flags they carry, so the oldest reports keep counting.

**Recommendations applied:** `summary.get('evaluated_attacks') is None` instead of the `in` check; `ev`/`vu` renamed; duplicated rationale collapsed into one comment; unused fixture keys stripped; `logger.debug` added on the derivation path per the loguru convention in `CLAUDE.md`.

**Verification:** `3594 passed, 2 skipped` (`pytest -m 'not integration'`), `ruff check src` and `ruff format --check src` clean, `basedpyright` 0 errors on `metrics.py`.

Three tests added beyond the original two: severity + token derivation, and the evaluation-stage-failure case that motivated the classification fix.

---

src/evaluatorq/dashboard/metrics.py
```diff
@@ -640,39 +750,23 @@ def redteam_overview(roots: list[Path] | None = None, *, page: int = 1, per_page
...
-            else:
-                evaluated += 1
-                if is_vuln:
-                    vulnerable += 1
-                    run_vuln += 1
-                    if severity == 'critical':
-                        critical += 1
-                else:
-                    resistant += 1
+        counts = _redteam_counts(data)
```
Baukebrenninkmeijer, line 953: This is where the scope grows past RES-1202. The old overview loop was:

```python
is_error = bool(res.get('error'))
if is_error: run_errors += 1
else:
    evaluated += 1
    if bool(res.get('vulnerable')): vulnerable += 1
    else: resistant += 1
```

A modern result with `evaluation={'passed': None}` (judge crashed or timed out) and no top-level `error` was counted **evaluated + resistant**. Under `_result_outcome` it returns `None` and drops out of `evaluated`/`resistant`/`vulnerable` altogether.

So `/redteam` overview numbers move for reports already on disk — the PR body's "no behavior change for modern reports" is scoped to `landing()` only, and the helper's docstring calls itself a mirror of this loop when it deliberately isn't.

I think the new rule is correct and worth keeping. Two things to make it safe:
1. A `redteam_overview()` test with a modern result carrying `evaluation.passed=None` and no error, pinning the intended counts — right now only `test_landing.py` gained fixtures, so a regression in `_result_outcome` breaks the overview page silently.
2. Say it in the PR body as a second deliberate change, so it can be reverted independently if the resistance rates move more than expected.

src/evaluatorq/dashboard/metrics.py
```diff
@@ -314,6 +331,99 @@ def landing(roots: list[Path] | None = None) -> Landing:
...
+    Reports too old to serialize ``evaluation`` at all fall back to the flags
+    they do carry.
+    """
+    evaluation = res.get('evaluation')
+    if isinstance(evaluation, dict):
+        passed = evaluation.get('passed')
+        return passed if isinstance(passed, bool) else None
+    if 'evaluation' in res or res.get('error'):
+        return None
+    return not res.get('vulnerable')
```
Baukebrenninkmeijer: Oldest-schema fallback: a result with no `evaluation` key, no `error`, and no `vulnerable` key at all comes back `True` — resisted. That's the same optimistic bias the docstring above warns against, one schema generation older.

`'vulnerable' in res` distinguishes "evaluated, not vulnerable" from "we don't know", and `None` is the honest answer for the second. No fixture covers a legacy result with `vulnerable` absent, so nothing would catch this shape arriving.

src/evaluatorq/dashboard/metrics.py
```diff
@@ -254,20 +256,35 @@ def landing(roots: list[Path] | None = None) -> Landing:
...
             summary = data.get('summary')
-            if isinstance(summary, dict):
+            summary = summary if isinstance(summary, dict) else {}
+            if summary.get('evaluated_attacks') is None:
+                # Legacy report: its summary predates ``evaluated_attacks``, and
+                # with it ``by_severity`` and ``token_usage_total``. Every
+                # aggregate below is therefore derived from the results list —
+                # otherwise the run adds zero to the donut, the severity bars and
+                # the cost totals while its own row still shows a rate (RES-1202).
+                counts = _redteam_counts(data)
```
Baukebrenninkmeijer, line 361: Consistency point worth a decision, not necessarily a change: `landing()` now derives a legacy run's resistant/vulnerable over an *evaluated-only* denominator, but `_redteam_row()` (untouched) still shows that same run's stored `summary.resistance_rate` in the recent-runs list.

If a legacy `resistance_rate` was computed over total results rather than evaluated ones, the same run shows one percentage in the row and contributes a different one to the donut above it — on the same screen.

Either route `_redteam_row`'s legacy score through `_redteam_counts` too, or note the intentional difference where the row is built.

src/evaluatorq/dashboard/metrics.py
```diff
@@ -314,6 +331,99 @@ def landing(roots: list[Path] | None = None) -> Landing:
...
+    evaluation = res.get('evaluation')
+    if isinstance(evaluation, dict):
+        passed = evaluation.get('passed')
+        return passed if isinstance(passed, bool) else None
+    if 'evaluation' in res or res.get('error'):
+        return None
+    return not res.get('vulnerable')
+
+
+def _result_usage(res: dict[str, object]) -> tuple[int, float]:
```
Baukebrenninkmeijer, line 515: `_result_usage` returns `(0, 0.0)` for any legacy result whose usage isn't under the expected nested shape, so an unrecognised schema contributes $0 to the cost panel rather than announcing itself. The debug log at line 267 reports `counts.tokens` after the fact but can't distinguish "cheap run" from "shape we couldn't read".

One `logger.debug` when a report has results but derives zero tokens across all of them would make that visible without adding noise to normal runs.

tests/dashboard/test_sim_transcript.py
```diff
@@ -239,19 +239,17 @@ def test_transcript_content_type_is_html(self, client: TestClient, roots: list[P
         r = client.get(f'/r/{rid}/sim/transcript?idx=0')
         assert 'text/html' in r.headers.get('content-type', '')
 
-    def test_transcript_contains_persona(
-        self, client: TestClient, roots: list[Path]
-    ) -> None:
+    def test_transcript_contains_persona(self, client: TestClient, roots: list[Path]) -> None:
```
Baukebrenninkmeijer: Nit, scope only: this file and `test_pairwise_surface.py` carry ~70 lines of pure reformatting (double→single quotes, re-joined signatures) unrelated to the legacy-counts fix. It makes the diff harder to read for a reviewer looking for the behaviour change, and it'll collide with anyone else touching these files.

If a formatter setting changed, that's its own commit.

Baukebrenninkmeijer (dismissed):

The landing fix itself is right: deriving from `results` beats a footnote, the degenerate no-results case is handled sanely, and the legacy fixture test is the one that matters. Approving.

But the PR does more than the description says. `redteam_overview()` was pointed at the same new classifier (line 753), and that classifier is *not* a faithful mirror of the loop it replaced — the old loop booked "generation succeeded, evaluation crashed" as **resistant**; `_result_outcome` now drops it from the counts entirely. So the `/redteam` overview page's resistance rates change for modern reports that are already on disk, while the body says "no behavior change for modern reports". Two lenses landed on this independently.

The new rule is the *better* one — a security metric shouldn't count an un-evaluated attack as resisted, and it matches `converters._is_evaluated`. It just needs to be stated as a deliberate second change with a test, not shipped inside a landing-aggregate fix. @Baukebrenninkmeijer, your call whether that rides along or splits out.

Baukebrenninkmeijer (changes requested):

Correcting my own verdict: approving while asking @Baukebrenninkmeijer to decide the scope question was the wrong call — switching to `REQUEST_CHANGES`. My earlier approval is dismissed; the inline comments on it still stand.

The landing fix is right and I'm not asking for changes to it. What blocks:

1. **Undeclared behaviour change to `/redteam`** (`metrics.py:753`) — needs Bauke's call on whether it rides along or splits out. `redteam_overview()` now uses `_result_outcome`, which does *not* match the loop it replaced: a result whose generation succeeded but whose evaluation crashed was counted **resistant** before and is now dropped entirely. Resistance rates move for reports already on disk, while the PR body says "no behavior change for modern reports" and the helper's docstring calls itself a mirror of that loop. The new rule is the better one — it just isn't what the PR claims to do.
2. **No test covers the changed path.** Only `test_landing.py` gained fixtures, so a regression in `_result_outcome` silently breaks the overview page's numbers.
3. **`metrics.py:364`** — a result missing the `vulnerable` key entirely counts as resisted, the same optimistic bias the docstring warns against one schema generation earlier.
4. **`metrics.py:367`** — an unreadable usage shape contributes `$0` to the cost panel with nothing logged.

Once 1 is decided and 2 exists, the rest are small.

currentlycodinng (approved):
Deriving rather than excluding is the right call, and folding the two classifications into one `_redteam_counts` is a good move. `_result_outcome`'s reasoning about not booking an unevaluated result as resistant is exactly right for a security metric. Three things.

**The overview refactor is not behaviour-preserving, and the description does not say so.**

The description scopes "no behavior change" to modern reports in `landing()`, which holds. But `redteam_overview` also swapped from the inline `is_error` / `is_vulnerable` loop to `_result_outcome`, and those two classify differently for modern reports. I ran the same fixtures against `origin/main` and against `ddf675d`:

```
A: target errored, judge still returned a verdict
   {'vulnerable': True, 'error': 'timeout', 'evaluation': {'passed': False}, severity critical}
   main    break_rate=None  critical_findings=0  avg_robustness=None
   branch  break_rate=1.0   critical_findings=1  avg_robustness=0.0

B: evaluation dict present, no boolean verdict
   {'vulnerable': False, 'error': None, 'evaluation': {}}
   main    break_rate=0.0   critical_findings=0  avg_robustness=1.0
   branch  break_rate=None  critical_findings=0  avg_robustness=None
```

Both new answers are better than the old ones, so I am not asking you to revert. A is a result that was judged and found vulnerable, and main was throwing it away because the target also errored; B is main padding the resistant bucket with a result that has no verdict, which is the exact optimistic bias `_result_outcome`'s docstring argues against. But this shifts ASR, critical findings and robustness on the red-team overview page for reports already on disk, and nothing in the PR says it will. Two asks: mention it in the description, and pin both cases with a test. Neither is covered right now, which is why the suite is green on main and on the branch for the same input.

**Two of the four files are unrelated formatting churn.** `tests/dashboard/test_pairwise_surface.py` and `tests/dashboard/test_sim_transcript.py` are pure `ruff format` reflows (quote style, line joins) with no connection to RES-1202. That is most of the +297/-79 and it makes the real change harder to find. It is also very likely why this is conflicting. I would pull them into their own commit or PR.

**The branch is CONFLICTING with main** and needs a rebase before it can go in.

**Nit:** a result can now be both errored and evaluated (case A), but `_RedteamCounts.errors` still counts every truthy `error`, so `_lifecycle_status(all_errored=...)` can mark a run "all errored" even though every result got a verdict. Small, and pre-existing in spirit, but the new classifier makes it reachable in a way it was not before.

Approving on the substance since the derived counts are correct and the overview changes are improvements. Please land the description note and the two regression tests with it.

---

What I checked: checked out `ddf675d`, `uv sync --all-extras`, `pytest tests/dashboard` 572 passed, `ruff check src` and `ruff format --check src` clean, `basedpyright` 0 errors. Note that CI has not actually run on this PR beyond the `label` job, so that is the only signal there is. The main-vs-branch comparison above came from running both, not from reading the diff.

KarinaKKarinaK:
## Review applied + conflicts resolved (b6291ccc)

Merged origin/main (resolving the metrics.py conflicts against the cost-accounting rework from RES-1238/RES-1194-era main) and addressed everything in the CHANGES_REQUESTED review.

### The 4 blocking items

1. **Undeclared `/redteam` behavior change: declared and kept.** The evaluation-crash rule is now stated in the PR body as a second deliberate change with its own revert boundary. It stays because it is the security-correct rule and matches `converters._is_evaluated`.
2. **Test coverage for the changed path: added.** `TestOutcomeClassification` in `test_redteam_overview.py` pins the overview counts for a run containing `evaluation.passed=None` with no error (dropped from evaluated, break rate and robustness computed over the remaining two), plus a direct unit test of `_result_outcome` across every schema generation.
3. **`vulnerable`-key-absent counts as resisted: fixed.** The oldest-schema fallback now returns unknown when the key is absent (`'vulnerable' in res` distinguishes the verdict from a truncated record), with a landing test for the truncated-record shape.
4. **Silent $0 from unreadable usage shapes: logged.** `_redteam_counts` emits one `logger.debug` when a report has results but zero derivable tokens across all of them.

### Inline comments

- **Row vs donut consistency (metrics.py:266):** took the route-through option rather than a comment. A legacy run's row score now derives through `_redteam_counts`, so the row and the donut cannot disagree; `test_landing_legacy_row_score_matches_derived_donut` pins a stored total-denominator rate (0.5) being replaced by the derived evaluated-only rate (2/3).
- **Reformatting noise:** the ~70 lines of pure reformatting in `test_sim_transcript.py` and `test_pairwise_surface.py` are reverted to main's formatting; the PR diff now only carries behavior.

### Merge notes

Two semantic conflicts worth flagging beyond the textual ones: main changed `_cost_usd` to return `None` for unrecorded cost, so `_result_usage` now guards it, and the legacy derivation path feeds `costed_runs`/cost totals only when a cost was actually derived (the input/output cost split is not recoverable from legacy results and stays out of those totals).

### Checks

`uv run pytest -m 'not integration'`: 3884 passed, 4 skipped. `ruff check src`, `ruff format --check src` clean. `basedpyright` (whole repo): 0 errors.

Baukebrenninkmeijer:
### UX review of the derivation — conclusions and fixes (8524ed0a)

The classifier change is right, but it moves numbers users have already seen. Four confusion vectors, all now addressed.

**1. Rates shift with no run re-executed — documented.**
Attacks with no judge verdict were previously counted as resisted and are now excluded, so runs recorded before this change read lower. To a user that looks like the agent got worse, not like the dashboard got honest. `docs/dashboard.md` gains a "How attack resistance is counted" section and an explicit note that rates can move for existing runs. Nothing in the UI can explain a number changing without a re-run, so this has to live in the release notes and the docs.

**2. Dashboard vs. the run's own report — recalculate, but show it.**
Worth stating the tradeoff since it was the open question. Two inconsistencies were available and we could only avoid one:

- *Don't recalculate:* the row matches the exported report, but a run's row disagrees with the aggregate it feeds.
- *Recalculate:* row and donut agree, but the row contradicts the saved HTML report, the CLI output at run time, and whatever number someone already pasted into a ticket.

We took recalculate — but the second inconsistency is the checkable one (1:1, and the artifact is the record), so it can't be silent. A re-derived rate now renders with a `*` and the Score tooltip gives the value the report records: *"Resistance rate: attacks the target withstood · 60 of 100 attacks evaluated · recalculated by the dashboard — the report for this run records 0.71"*. Rows where the derivation agrees are left unmarked, so the marker carries signal instead of decorating every legacy run.

**3. A run could lose its score — fixed.**
`_redteam_row` overrode the stored rate unconditionally, so a legacy report with a recorded rate but no results to re-derive from went from `0.50` to blank:

```
legacy report, 40 attacks, empty results, summary.resistance_rate = 0.5
before: headline "40 attacks" | score: None
after:  headline "40 attacks" | score: 0.50
```

Deriving nothing is not the same as measuring zero. Three tests pin this and the marker rules.

**4. Unstated denominator — tooltip added.**
The headline is `total_results` (every attack attempted); the score is over evaluated-only. `100 attacks · 0.95` where the 0.95 is over 60 is two correct numbers sitting next to each other meaning different things. The tooltip now names both.

**Verification:** `3893 passed, 2 skipped`, `ruff check src` + `ruff format --check src` clean, `basedpyright` 0 errors repo-wide, `mkdocs build --strict` clean.

Still open, not addressed: `_redteam_counts` defaults a missing `attack.severity` to `low`, which buckets unknown-severity vulnerabilities into the lowest band — the same optimistic-default smell this PR spent its time removing, one field over. Small, and arguably its own change.

---

KarinaKKarinaK:
## Remaining item closed (5d5a812c)

The severity work in 1f8062c6 covered a *missing* `attack.severity` but not a present, unrecognised one: a value like `sev1` or `medium-high` still created a bucket the display comprehension silently drops, so the bars could sum to less than the donut again. Both producers now fold anything off the report's scale into the `unknown` bucket: `_redteam_counts` for derived counts and `_summary_severity` for stored summaries, which also case-normalises keys so a stored `HIGH` renders instead of vanishing. Two tests pin it (derived path and stored-summary path).

Also swapped the em dash in the recalculated-score tooltip for a semicolon per UI copy style.

Checks: 3897 passed, 4 skipped; ruff check + format, basedpyright repo-wide all clean.

With that, every item from all three review rounds is addressed. @Baukebrenninkmeijer the blocking CHANGES_REQUESTED review is yours to dismiss when you agree.

Baukebrenninkmeijer:
### Unrecorded severity now has its own bucket (1f8062c6)

Closing the loose end flagged in the last comment: `_redteam_counts` was resolving a missing `attack.severity` to `low`.

**Not `None`, though.** `Landing.severity` is assembled by filtering over `SEVERITY_ORDER`, so an unmapped severity would be dropped from the list outright — the vulnerability would still count in the donut but disappear from the severity bars, leaving that panel summing to less than the number sitting next to it. Same silent-loss shape as the bug this PR started on, just relocated one panel over.

So it gets a named bucket instead:

```python
UNKNOWN_SEVERITY = 'unknown'   # sorts after the real scale, rendered only when non-zero
```

The bars now sum to the donut, and the bucket says what it is rather than quietly claiming a grade it never had. `severity` is a required field on the contract (`redteam/contracts.py:421`), so this only fires on a truncated or hand-edited report — defensive, not a path real runs take.

**Fell out while wiring it:** the severity bar palette was a positional list of four colours, and `Landing.severity` drops zero-count buckets. A run whose only findings were `low` therefore rendered that row **red**, because `low` had shifted into position 0. The palette is now keyed by severity name; anything off the scale, `unknown` included, stays neutral. Pinned with a test.

**Verification:** `3897 passed, 2 skipped`, `ruff check src` + `ruff format --check src` clean, `basedpyright` 0 errors repo-wide. Four tests added.

That clears everything raised across this review — no open findings from my side.

---

Baukebrenninkmeijer (approved):
Approving. Everything raised across this review is addressed, and `5d5a812c` closed a hole I left open.

**What this branch ended up being.** It started as "legacy runs are missing from the landing donut" and turned into a consistency pass over how the dashboard counts a red-team run. Worth naming the through-line, because every fix here was the same bug wearing different clothes: **an unreadable or absent field resolving to the most reassuring value.** Unevaluated attacks counted as resisted. A missing severity counted as low. An off-scale severity counted as nothing at all. On a security surface that bias only ever points one way.

**The substantive changes:**

- One classifier (`_redteam_counts`) behind both the overview and the landing aggregate, keyed on `evaluation.passed` per `converters._is_evaluated`. A judge that crashed no longer reads as a target that held.
- Legacy runs contribute their severities and token usage, not just their counts — a run in the run total whose spend was invisible everywhere else was its own inconsistency.
- Off-scale and missing severities fold into a named `unknown` bucket that renders, so the bars sum to the donut instead of quietly falling short. `5d5a812c` extended this to present-but-unrecognised values and to the stored-summary path, and the case-normalisation there (`'HIGH'` no longer vanishes) is a real fix in its own right.
- A re-derived rate is marked and the tooltip gives the value the report records, so a row can be reconciled against that run's exported report rather than silently contradicting it.
- The Score tooltip names the denominator, since the headline counts every attack and the rate counts only the graded ones.

**One consequence to land deliberately, not discover in a support thread:** resistance rates will read lower for runs recorded before this. Nothing about those runs changed — only what counts as a measured attack. It's documented in `docs/dashboard.md`, but it belongs in the release notes too.

Verified locally at `5d5a812c`: `3899 passed, 2 skipped`, `ruff check src` clean, `basedpyright` 0 errors repo-wide. CI green across all four Python versions.

No open findings.

PR #130, opened 2026-08-09
fix(cost): honest cost reporting across spans, reports and dashboard

Baukebrenninkmeijer:
## What

Adds provider-reported cost to `Usage`, records it on OTel LLM spans, and surfaces it in red-team reports, simulation reports and the dashboard — without ever rendering an unknown cost as `$0.00`.

The final commit is a fix pass applied after an adversarial review of the branch. It resolves four defects that the full test suite, ruff and basedpyright all passed over:

- **`record_llm_response` could raise on a successful call.** `Usage.extract` validates token counts with `ge=0` and now sits on the success path of every LLM call. A provider or proxy reporting a negative count raised a `ValidationError` out of an otherwise-fine response. Counts are clamped at 0, mirroring the clamp already applied to cost.
- **The simulation report's cost-coverage label was unreachable.** The section never emitted `calls`/`priced_calls`, so `cost_coverage()` always saw 0 and silently dropped the `(N of M calls)` qualifier — a partial total read as authoritative. The red-team report already plumbed these; simulation now matches.
- **Red-team per-agent cost columns** rendered bare figures from the same partial data the overall row already qualified.
- **The dashboard showed lower-bound spend totals unlabelled.** `Landing`, `SimOverview` and `RedTeamOverview` now carry `priced_calls`/`cost_calls` and render the same qualifier as the reports.

Plus: `orq-ai-sdk` is no longer a core dependency (every `orq_ai_sdk` import is lazy behind an explicit "not installed" error, so the pin belongs only in the extras); `_bars` accepts a formatter callable so cost bars reuse `fmt_cost` instead of a parallel `'${:.4f}'` string that had drifted from it; and the openai-extras test uses `model_validate` instead of `.construct()` — construct skips validation entirely, so it would have passed under the exact `extra='ignore'` regression it exists to catch.

## Design notes

- `None` cost means "the provider did not report it" and is distinct from a real `0.0`. It renders as an em dash, never `$0.00`.
- A cost summed over calls where only some reported one is a **lower bound**. Every surface that shows such a figure appends `(N of M calls)`.
- Reports saved before `priced_calls` existed carry `priced_calls=0`, which is treated as "not tracked" rather than "nothing was priced", so old reports stay unannotated instead of claiming `0 of N`.

## Testing

- `uv run pytest -m 'not integration'` — 3825 passed, 2 skipped.
- ruff, `ruff format --check`, basedpyright all clean.
- Each structural fix was reverted in isolation and the suite re-run: the new tests fail without their fix and pass with it.

## Known gaps

Deliberately left out, all minor and none reachable from real provider data: per-run cost cells in the dashboard run lists (aggregate coverage covers the reported defect), `priced_calls > calls` is constructible, `with_calls(-1)` bypasses `ge=0`, `model_dump(exclude_none=True)` still emits `cost_usd: None`, and a cache-write-only payload is dropped by `extract`'s signal check.


currentlycodinng (approved):
The core idea is right and the fix pass caught real things. `_usage_first_int` skipping rather than clamping is the correct call, and I confirmed the guard also covers the nested detail reads since `_usage_detail_int` routes through it. Two things below, one of which I think is worth another commit.

**The dashboard aggregate still shows an unqualified lower bound when legacy and new reports mix.**

`_cost_calls` returns `(0, 0)` for a report with no `priced_calls` key, but that report's `cost_usd` is still summed into `total_cost`. So on a landing page with one pre-coverage report and one fully-priced new one, the totals are `priced=1, calls=1`, `cost_coverage` returns `''`, and the tile renders the combined figure bare. Probed on the branch:

```
legacy report: $10.00 over 10 calls, no priced_calls key
new report:    $0.50  over  1 call,  priced_calls=1

total_cost   = 10.5
priced/calls = (1, 1)
coverage label rendered: False
```

10 of those 11 calls have unknown pricing coverage and the page says nothing. That is the same "a partial total reads as authoritative" failure this PR exists to close, just relocated to the aggregate surfaces, and `test_legacy_report_does_not_inflate_the_coverage_denominator` locks it in as intended.

I follow the reasoning that "1 of 11 calls" would be a fabricated warning. The problem is that an empty label currently means both "every call was priced" and "we have no idea", which are opposite claims. A third counter for calls with unknown coverage would let you say something honest, e.g. `(3 of 10 calls priced, 10 calls unknown)`, or at minimum suppress the confident-looking bare figure when any contributing report predates the field. Your call on the wording, but I do not think "no label" is the right rendering for "unknown".

**The `orq-ai-sdk` removal needs a CHANGELOG entry.** I checked every `orq_ai_sdk` import site and the claim holds: `deployment.py:27`, `fetch_data.py:18` and `backends/orq.py:24` are all `TYPE_CHECKING` or `try`-guarded, the runtime imports are function-local, and both `_orq_cls is None` sites raise a message pointing at `evaluatorq[orq]`. So the code is fine. But this changes what a bare `pip install evaluatorq` gives you: anyone already calling `deployment()`, `fetch_data`, or an ORQ redteam target without the extra goes from working to `ImportError`. The 1.3.0 changelog documents the two moves in the opposite direction (loguru and openai becoming core deps) under "Notable defaults", so this one belongs there too, and arguably under Breaking Changes. Right now nothing in the release notes will tell a user why their install broke, and the `fix(cost):` title does not hint at it.

**Nit:** `_usage_first_int`'s `logger.warning` fires per bad key per call. A provider that consistently reports something non-finite will log on every single LLM call; debug level or a once-per-process guard would be kinder.

Approving since neither of these is a correctness bug in the new code, but I would like to see the coverage-unknown case addressed before this ships rather than as a follow-up.

---

What I checked beyond reading: checked out `d20a0e3`, `uv sync --all-extras`, ran the five touched test files (121 passed), `ruff check src` and `ruff format --check src` clean, `basedpyright` 0 errors, and CI is green on 3.10 through 3.13 plus all three SDK compatibility matrices. The aggregate finding above came from a scratch probe against `metrics.landing` / `view.landing_body` on the branch, not from reading.

Baukebrenninkmeijer:
Thanks — all three addressed in c8bacc8d.

**Coverage-unknown aggregate.** You're right that an empty label was carrying two opposite meanings. `_cost_calls` now returns a third counter: a report that contributes cost but predates `priced_calls` counts its calls as *unknown coverage* rather than being dropped (or folded into the denominator, which would fabricate the "1 of 11 calls" warning in the other direction). The tiles render it. Your exact probe on the branch now gives:

```
total_cost 10.5   priced/calls/unknown (1, 1, 10)
label: (1 of 1 calls priced, 10 calls of unknown coverage)
```

A legacy report with no cost at all still contributes nothing — it adds nothing to the total, so it has no coverage to be unknown about. Three new tests in `TestDashboardCostCoverage` cover mixed, legacy-only, and costless-legacy.

**`orq-ai-sdk` removal.** Added to the 1.3.0 Breaking Changes section, naming the paths that go from working to an error under a bare `pip install evaluatorq` (`deployment()`, `fetch_data()`, ORQ redteam/simulation targets) and the extras that restore them.

**Nit.** All three `Usage.extract` hostile-value logs dropped to `debug`, with a comment on why — a consistently-bad provider logged once per LLM call.

Unrelated but in the same push: the ORQ agent backend now emits one LLM span per `agents.responses.create` mirroring the chat-completions/responses shape, so its aggregate `gen_ai.usage.*` span write could be removed without losing the numbers. Usage lives on per-call spans only; the sink aggregates.

PR #133, opened 2026-08-10
fix: distinguish "not evaluated" from "0% resistant" in red-team reports

Baukebrenninkmeijer:
## Why

A red-team run whose judge calls all failed reported `resistance_rate: 0.0` with `vulnerabilities_found: 0` and exited `0` — indistinguishable from a fully compromised target, and silent about why. Two changes make that state impossible to reach.

## Nullable rates

All eight rate aggregates (`ReportSummary`, `VulnerabilitySummary`, `CategorySummary`, `DimensionSummary` and its subclasses, `ReportSnapshot`) are now `float | None`, where `None` means "nothing in this slice could be scored". A `0.0` resistance rate reads as *fully compromised* and a `0.0` vulnerability rate reads as *fully safe* — neither is true when no verdict exists.

Every renderer branches three ways (vulnerable / resistant / not evaluated): both dashboards, HTML + Markdown export, terminal display, executive summary, hooks. Sorts put unevaluated slices last; risk rankings exclude them.

Also fixes a latent `pct` shadowing bug — `common/reports/__init__.py` re-exported it from both `html_helpers` and `md_helpers`, and the later import won, so the None-safe implementation was dead for every package-level importer.

## Coverage gate

`EvaluatorConfig.min_evaluation_coverage` (default `0.8`, CLI `--min-evaluation-coverage`) is a **run-level** floor. It is distinct from `min_successful_judges`, the **per-attack** jury quorum that is what *creates* unevaluated attacks in the first place.

`ReportSummary.no_verdict` and `.coverage_below_minimum` are the single definition of each condition — previously the same check was spelled four different ways. The CLI exits non-zero on either and names the dominant cause.

> [!WARNING]
> **Behaviour change.** A run finishing at 79% evaluation coverage now exits `1` where it previously exited `0` with a warning. Pass `--min-evaluation-coverage 0` (or set `min_evaluation_coverage=None` in Python) for the old warn-only behaviour. Deliberately *not* marked `BREAKING CHANGE:` — this ships as a patch; see the docs for the migration note.

## Judge-error surfacing

`RedTeamResult.evaluation_error` carries a structured `RunError` for a judge that could not return a verdict. Deliberately separate from `error`, which means the attack itself never ran — the attack *did* run here and the transcript exists, so collapsing the two would make an unscored result look like an execution failure. It rolls into the same error tally, keyed `evaluation/<code>`.

The panel-of-judges path records the cause too (previously only the single-judge fast path did), and a scorer exception that only reached `EvaluatorScore.error` is lifted rather than discarded.

**Dropped the fail-loud re-raise** on API connection/status judge errors. Nothing honoured it: `run_evaluations` catches every exception into a bare error string, so the run continued anyway *and* the cause was lost — which is precisely the reported bug. The coverage gate is what fails the run now.

## Verification

- `uv run basedpyright` — 0 errors
- `uv run pytest -m 'not integration'` — 3782 passed, 2 skipped
- `uv run --group docs mkdocs build --strict` — passing

Docs updated in the same PR: `docs/guides/red-teaming.md`, `docs/cli-reference/redteam.md`, `docs/faq.md`, `docs/llm-as-a-jury.md`, `docs/custom-evaluators-and-frameworks.md`, README, and the 20 red-team examples.


currentlycodinng (changes requested):
The diagnosis is right and the shape of the fix is the right one. Making the rates nullable rather than picking a safer default is the only version of this that cannot lie, and separating `evaluation_error` from `error` matters more than it looks, because "the attack never ran" and "the attack ran and we could not score it" need different responses from whoever reads the report.

I want to call out the part that is easy to miss: `src/evaluatorq/redteam/ui` is excluded from basedpyright in `pyproject.toml:125`, so every one of the tri-state changes in that 137-line file had to be right by hand with no type checker behind them. I read all eight arithmetic sites and they are all guarded, three of them by `rankable_*` comprehensions that filter None before sorting, and the risk ranking surfaces unevaluated vulnerabilities separately rather than dropping them silently. I also rendered a 0-of-3-evaluated report through `export_markdown`, `export_html`, `build_redteam_facts` and `print_report_summary`: nothing raises, and no surface prints a percentage for a rate that does not exist. Locally `ruff check src`, `ruff format --check src`, `basedpyright` (0 errors) and `pytest -m 'not integration'` (3782 passed) all match what you reported.

Two things still report a lie, both of them the same lie the PR is named after.

## The coverage gate never fires on a static run

`min_evaluation_coverage` is only ever stamped onto the summary at `runner.py:2744`, inside `_run_dynamic_or_hybrid`. `_run_static` builds its report through `static_results_to_report`, which calls `compute_report_summary(unified)` with no floor (`converters.py:451`), and nothing after the dispatch in `red_team()` stamps it either. `merge_reports` inherits the floor from its sub-reports, but for a static run every sub-report has None, so there is nothing to inherit.

Since `coverage_below_minimum` returns False whenever the floor is None, the gate is a no-op for `--mode static`. I ran it:

```
--- static pipeline (static_results_to_report)
  default floor in EvaluatorConfig : 0.8
  evaluated / total               : 2 / 4
  evaluation_coverage             : 0.5
  summary.min_evaluation_coverage : None
  no_verdict                      : False
  coverage_below_minimum          : False    <-- CLI exit gate
--- same results, floor stamped (what _run_dynamic_or_hybrid does)
  coverage_below_minimum          : True
```

So `eq redteam run --mode static --min-evaluation-coverage 0.8` accepts the flag and exits 0 at 50% coverage. `no_verdict` is unaffected, so the total-blackout case still fails correctly on both paths; it is only the partial-coverage gate that is dynamic-only. The docs added here state the behaviour unconditionally, and `docs/faq.md` tells Python-API callers to check `report.summary.coverage_below_minimum` themselves, which has the same hole.

Nothing catches it because the new CLI tests construct `ReportSummary` objects by hand and set `min_evaluation_coverage` inline, so they exercise the CLI's branch but never the question of who stamps the floor. That also explains `test_exits_0_when_partial_coverage`: it asserts exit 0 at 1-of-5 evaluated, and the docstring calls whether partial failures should fail CI "an open policy question". This PR answers that question, and the only reason that test passes is the unstamped floor. As written it pins the bug.

Stamping inside `compute_report_summary`'s callers, or in `red_team()` after the dispatch so both legs go through one place, would fix the gate and the test together.

## "Passed" still counts unevaluated attacks as passed

`display.py:132`:

```python
passed_count = vuln_summary.total_attacks - vuln_summary.vulnerabilities_found
```

This PR changed the two lines on either side of it and left the arithmetic. On a run where nothing could be scored, the terminal summary prints:

```
│ Vulnerability                       │ Domain   │ Tested │ Passed │  ASR │
│ Agent Goal Hijacking (ASI01)        │ Agent    │      3 │      3 │  n/a │
```

Three attacks, no verdicts, three passed. The ASR column correctly says `n/a`, and the Passed column next to it says the target resisted all three. It is worse than a neutral rendering, because `display.py:140` styles the cell `'green' if passed_count == vuln_summary.total_attacks`, so a fully unevaluated vulnerability renders green.

The other two places that compute this get it right, `ui/dashboard.py:923` and `dashboard/metrics.py:284`, and both use `evaluated_attacks - vulnerabilities_found`. `display.py` cannot do the same because `VulnerabilitySummary` has no evaluated count; neither does `CategorySummary` or `DimensionSummary`. So this needs an `evaluated_attacks` field on the per-slice summaries rather than a local fix, which is also what would let the tables show the "2 of 5 scored" denominator the summary block already shows at the top.

## Smaller things

**`run_evaluations` does not exist.** It is cited in `evaluator.py:320`, `converters.py:107` and `converters.py:525`, and in the PR description, as the reason the fail-loud re-raise could be dropped. The function is `process_evaluator` in `src/evaluatorq/processings.py:202`, and the claim is correct: line 269 catches every exception into `EvaluatorScore(score=EvaluationResult(value=''), error=str(error))`, so the re-raise could never abort a run and the cause survived only as a string. Since those comments are the recorded justification for removing a safety behaviour, they should name a function someone can grep for.

I checked the replacement behaviour rather than assume it. With a client that raises `APIConnectionError` on every call:

```
single judge   passed=None  evaluation_error='api_connection' stage='evaluation'
panel of 3     passed=None  evaluation_error='api_connection' stage='evaluation'
               explanation: No judge produced a usable verdict; last error: Connection error.
```

Neither path raises and both name the cause, which is the behaviour the reported bug was missing.

**`RedTeamResult.vulnerable` quietly stopped being required.** It went from `vulnerable: bool` to `bool | None = None`, so a caller that forgets the field now gets "not evaluated" instead of a `ValidationError`. Given the whole point is that None must never be read as passing, a required tri-state (`bool | None` with no default) keeps the new semantics and the old guarantee.

**`static_sample_to_result:423`** sets `evaluation_error=... if evaluation else None`, so a sample with no evaluation at all gets `vulnerable=None` with no error attached, which contradicts the field's own "Always accompanies vulnerable=None".

**The branch is conflicted.** `README.md`, `docs/cli-reference/redteam.md`, `docs/faq.md` and `src/evaluatorq/common/reports/md_helpers.py` all conflict with main. The md_helpers one is worth resolving carefully: main added `fmt_cost`/`cost_coverage` right above `pct`, and the conflict sits on the definition this PR deliberately replaced with `pct = _html_pct`. Taking main's side there silently restores the float-only shadow, which is the exact latent bug this PR fixes. `test_md_helpers_pct_is_the_same_object_as_html_helpers_pct` would catch it, which is a good reason that test exists.

---

What I checked: fetched `81a0664`, ran `uv sync --all-extras`, `ruff check src`, `ruff format --check src`, `basedpyright` and the full non-integration suite (3782 passed); probed the coverage gate through `static_results_to_report` and compared against a stamped summary; rendered a 0-coverage and a 1-of-3 report through all four reporting surfaces; drove `OWASPEvaluator` single-judge and 3-judge panel paths against a client that always raises to confirm neither re-raises and both record a structured cause; read `process_evaluator` to confirm the swallow claim; read all eight rate-arithmetic sites in the basedpyright-excluded `redteam/ui`; and checked the merge against current main.

Baukebrenninkmeijer:
Thanks — the static gate one was the real find. Both blockers fixed in df41aa60, plus three of the four smaller ones.

**The coverage gate never fires on a static run.** Confirmed and fixed, and it was slightly worse than you described: the whole `runner.py:2741-2768` block is dynamic-only, so static lost the `NO VERDICT` and below-floor pipeline *warnings* too, not just the stamp. Extracted it to `_apply_coverage_policy(report, pipeline_config)` and call it from both legs. In static it runs before the report is persisted, so the saved JSON records the policy as well — stamping post-dispatch in `red_team()` would have fixed the exit code and left the artifact unstamped.

On the test: `test_exits_0_when_partial_coverage` is now `test_exits_0_when_partial_coverage_and_no_floor_configured`, asserts the floor really is None, and the "open policy question" docstring is gone — this PR answers it. New `TestApplyCoveragePolicy` covers stamping, the None case, a 0.0 floor (which must not collapse into "absent"), both warning branches, and a static run end-to-end asserting `min_evaluation_coverage == 0.8` and `coverage_below_minimum is True`.

**"Passed" counts unevaluated attacks as passed.** Confirmed and fixed. One correction: `CategorySummary` does already carry `evaluated_attacks` (`contracts.py:1448`) — it's `VulnerabilitySummary` and `DimensionSummary` that lacked it, and only the vulnerability table has a Passed column. Added the field to `VulnerabilitySummary`, populated at the single construction site (`converters.py:940`; `merge_reports` recomputes rather than merging by hand), and took your suggestion on the denominator — the Tested cell now reads `0 of 3` when coverage is partial. Regression test renders the table and asserts the cells.

**`run_evaluations` does not exist.** Correct, and fair — those three comments are the recorded justification for deleting a safety behaviour, so they should name something greppable. All three now say `evaluatorq.processings.process_evaluator`, and the claim itself holds: line 269 catches into `EvaluatorScore(score=EvaluationResult(value=''), error=str(error))`.

**`static_sample_to_result`.** Fixed. It now synthesizes a `no_evaluation` `RunError` when there's no evaluation block — but only when `row.error` is empty. If the attack failed to execute, `error` is the right field, and duplicating it into `evaluation_error` would claim the judge failed on a transcript that never existed. Two tests, one per branch.

**`RedTeamResult.vulnerable` should be a required tri-state.** Declining this one, and I don't think it's a close call. A no-default field breaks loading every report JSON written before this PR, and report round-tripping is what the dashboard and `eq redteam from-run` are built on. The guarantee you want — that None is never read as passing — is enforced at the read sites rather than at construction; that's what the rest of this PR is. Happy to reopen if you think the artifact compatibility argument is wrong.

**The conflict.** Already resolved in e4dc3bf5 before your review landed, and resolved the way you'd want: `md_helpers.py` kept both main's new `fmt_cost`/`cost_coverage` and this PR's `pct = _html_pct`, with `test_md_helpers_pct_is_the_same_object_as_html_helpers_pct` green. Your instinct about taking main's side there was right — it would have restored the float-only shadow.

One caveat worth stating plainly, since it's a behaviour change nobody asked for: because the floor is a fraction, small runs are sensitive to it. A 3-attack smoke test now fails CI on a single transient target 500 (2/3 = 67% < 80%). I checked that every unevaluated path in the static scorer is a genuine error path — missing category, no evaluator registered, target error, judge failure, all returning `_error_result` with `pass=None` (`evaluatorq_bridge.py:170,238-247`) — so a healthy run is always at 1.0 and never trips. But "one network blip" and "the judge is misconfigured" now look identical to CI. `--min-evaluation-coverage 0` is the escape hatch.

Gates: `ruff check src`, `ruff format --check src`, `basedpyright` (0 errors), `pytest -m 'not integration'` (3988 passed), `mkdocs build --strict` all clean. Docs needed no edit — they already stated the gate unconditionally, which step 1 made true rather than aspirational.

PR #134, opened 2026-08-10
fix(evaluatorq): route recommendation calls through structured output (RES-822)

KarinaKKarinaK:
## Summary

Implements RES-822. The redteam and simulation recommendation report modules were the last two LLM call sites still on bare `response_format={'type':'json_object'}` + `json.loads`, with no schema and no fence tolerance. On any provider that fences the response (Anthropic/Gemini via the Orq router, the exact case the existing fence stripper was written for) both silently fell into their `except Exception` branch and logged `Failed to generate recommendations for ...`, dropping the section from the report.

Both now go through the house structured-output-first pattern.

## Changes

- **`src/evaluatorq/common/structured_output.py`** and **`common/extract_json.py`** (moved, preserving history): `generate_structured` and the fence-tolerant extractor now live in `common` so redteam can reach them without a simulation dependency and without forking a third copy (an explicit acceptance criterion). `generate_structured` delegates to the canonical `common.tracing.with_llm_span`.
- **`simulation/utils/structured_output.py`** / **`extract_json.py`**: thin re-export shims, so the 6+ existing simulation callers (persona/scenario/traces generators + their tests) are untouched.
- **`redteam/reports/recommendations.py`** and **`simulation/reports/recommendations.py`**: each gains a small Pydantic response model and calls `generate_structured`; the fallback `raw` is parsed via `extract_json_from_response` + `model_validate_json`. The `except Exception -> warning` path is kept, so a malformed payload degrades to no-recommendations instead of crashing.

### Helper generalization (kept it one copy, not a fork)

`generate_structured` gained two things so both surfaces migrate onto it faithfully:

- **Optional `temperature`** (sent only when not `None`) — the simulation path deliberately lets the provider default stand for reasoning models.
- **`extra_kwargs` passthrough** merged into both `parse()` and the fallback `create()` via a single dict splat — carries the redteam path's `extra_body` (router retry hints), the evaluator's reasoning-model `temperature=1.0` escape hatch, and user `llm_kwargs`, without ever tripping a "multiple values for keyword" error.

One correctness point the migration surfaced: **trace headers and run metadata are now applied LAST**, so the active span's `traceparent` still propagates (merged over, not replaced by, a caller's own headers); a naive "extra_kwargs wins" merge would have let a stale caller `traceparent` shadow the active span. Metadata precedence follows `apply_pipeline_metadata`'s contract: run metadata fills in defaults and a caller-supplied metadata key wins on conflict (the earlier "run metadata stays authoritative" wording was backwards).

## Testing

- `tests/redteam/test_recommendations.py` and `tests/simulation/reports/test_recommendations.py`: happy path (structured `parse`), fenced ` ```json ` fallback (the silent-drop bug), and malformed-payload path for both surfaces. The RES-817 temperature / `extra_body` / run-metadata / `llm_kwargs`-precedence assertions are preserved, ported onto the `parse()` call path.
- Full non-integration suite: **3940 passed, 2 skipped**. `ruff check src`, `ruff format --check src`, `basedpyright` (whole repo), and `mkdocs build --strict` clean.

## Out of scope (per ticket)

- `adaptive/*` (already migrated), the model-family kwarg sanitizer.

Ticket: RES-822


src/evaluatorq/common/structured_output.py
```diff
@@ -0,0 +1,142 @@
...
+        input_messages=messages,
+        attributes={'orq.llm.purpose': label},
+    ) as span:
+        trace_headers = await get_trace_context_headers()
+
+        def _params(response_kwargs: dict[str, Any]) -> dict[str, Any]:
+            base: dict[str, Any] = {
+                'model': model,
+                'messages': typed_messages,
+                'max_tokens': max_tokens,
```
Baukebrenninkmeijer: `max_tokens` here is what changes the wire format for both migrated call sites. Before this PR:

- `redteam/reports/recommendations.py` → `completion_params(..., max_completion_tokens=1500, ...)`
- `simulation/reports/recommendations.py` → `create(..., max_completion_tokens=800, ...)`

Every other chat call in the repo uses `max_completion_tokens` too (`common/llm_call.py:124`, `common/judge.py:249`, `redteam/adaptive/*`, and `LLMCallConfig.completion_params` at `contracts.py:188`). OpenAI rejects `max_tokens` outright for the o-series and gpt-5 families, and the base dict is shared by both the `parse()` and the fallback `create()`, so a direct (non-router) client with `openai/gpt-5.4-mini` 400s twice and the caller's `except Exception` turns it into "Failed to generate recommendations for …" — the exact silent drop RES-822 removes.

If the Orq router normalises the key, this is a no-op and worth a one-line comment saying so; if it doesn't, `max_completion_tokens` in the base dict fixes all `generate_structured` callers at once.

src/evaluatorq/redteam/reports/recommendations.py
```diff
@@ -186,40 +199,36 @@ async def generate_focus_area_recommendations(
...
-                    response_format={'type': 'json_object'},
-                    extra_body=cfg.retry_extra_body(llm_client),
-                ),
+            # extra_kwargs carries the same three things completion_params used
+            # to merge, in the same precedence: the router retry body, then the
+            # evaluator's own extra_kwargs (which is where a reasoning model's
+            # temperature=1.0 escape hatch lives), then user llm_kwargs on top.
+            # generate_structured splats these LAST over its base params, so an
+            # override wins without a "multiple values for keyword" error.
+            extra_kwargs: dict[str, Any] = {
```
Baukebrenninkmeijer: Building `extra_kwargs` by hand here drops the `_RESERVED_COMPLETION_KEYS` guard that `completion_params` applied on the old path (`contracts.py:143` + `:180`), which raised `ValueError` when `extra_kwargs` carried `model`, `messages`, `response_format` or `extra_body`.

Two concrete consequences now that `generate_structured` does `base.update(extra_kwargs)` *after* setting `response_format`:

- `EvaluatorConfig(extra_kwargs={'extra_body': {...}})` silently replaces `cfg.retry_extra_body(llm_client)`, so the router retry hints vanish instead of erroring.
- `extra_kwargs={'response_format': ...}` silently replaces `_FocusAreaLLMResponse` — defeating the schema this PR adds, with no error and no log.

Both were loud failures before. Reinstating the check inside `_params()` would cover every `generate_structured` caller, not just this one.

src/evaluatorq/redteam/reports/recommendations.py
```diff
@@ -30,6 +31,18 @@
...
 
+class _FocusAreaLLMResponse(BaseModel):
+    """Schema the analysis LLM fills for one focus area (RES-822).
+
+    Structured-output-first: ``generate_structured`` enforces this via
+    ``parse()`` and falls back to ``json_object`` for models that reject it,
+    where a fenced payload is recovered with ``extract_json_from_response``.
+    """
+
+    recommendations: list[str] = Field(default_factory=list)
```
Baukebrenninkmeijer: The fallback path is stricter than the code it replaces, in the one place that can least afford it.

Old: `recs = [str(r) for r in recs if r]` — a stray `1` or `null` in the array was coerced or filtered, and the rest of the section survived.
New: `model_validate_json` against `list[str]`. Checked against pydantic v2 in this repo's env:

```
{"recommendations": [1, "ok"]}   -> ValidationError
{"recommendations": [null]}      -> ValidationError
{"patterns_observed": 5}         -> ValidationError
```

That `ValidationError` is caught by the outer `except Exception` and drops the whole focus area. It only bites on the `json_object` fallback — i.e. the weaker models most likely to emit a stray non-string. `_SuggestionsLLMResponse.suggestions` at `simulation/reports/recommendations.py:42` has the same shape.

A `BeforeValidator` that stringifies and drops falsy items restores the old tolerance while keeping the schema for `parse()`. If the tightening is deliberate, a test asserting "one bad item drops the section" would pin it.

src/evaluatorq/common/structured_output.py
```diff
@@ -0,0 +1,142 @@
...
+                base['temperature'] = temperature
+            # extra_kwargs first: a caller's provider options (extra_body, a
+            # reasoning-model temperature override, user llm_kwargs) win over the
+            # base fields without a "multiple values for keyword" error.
+            if extra_kwargs:
+                base.update(extra_kwargs)
+            # Trace headers and run metadata are applied LAST so they stay
+            # authoritative: the active span's traceparent must propagate even
+            # when a caller passed its own headers (merge, don't replace, so
+            # other caller headers survive), and run metadata is not something a
```
Baukebrenninkmeijer: The trace-header half of this comment is right; the run-metadata half is backwards. `apply_pipeline_metadata` (`common/llm_call.py:112`) does:

```python
params['metadata'] = {**md, **(params.get('metadata') or {})}
```

Caller-supplied `metadata` — already merged into `base` from `extra_kwargs` on the line above — is spread last and wins on key conflict; its own docstring says so. So a caller passing `metadata={'evaluatorq_run_id': ...}` *can* shadow the real run id, which is what this comment (and the PR description) says cannot happen. Position in the merge doesn't change that, because the override happens inside the helper.

Comment-only fix unless caller-wins is itself the bug, in which case it belongs in `apply_pipeline_metadata`, not here.

Baukebrenninkmeijer (commented):

The migration itself is clean: both call sites move onto one helper, the shims keep the six existing simulation callers untouched, and the fenced/malformed fallback paths are actually tested rather than asserted in the description. Spec-wise this matches RES-822 — no divergence found against the ticket.

Four things worth a look before merge, none of them blocking on their own:

1. **Token-limit parameter name changed on both call sites.** Both previously sent `max_completion_tokens` (redteam 1500, simulation 800); the shared helper sends `max_tokens`. On a plain OpenAI client with a reasoning model (`DEFAULT_MODEL` is `openai/gpt-5.4-mini`) that is a 400 — and it lands in the same `except Exception -> warning` that drops the section, i.e. the symptom this PR removes. Not flagged as blocking because the shared helper already behaved this way for the existing simulation callers, so if it does break, it breaks more than this PR — but these two call sites are a regression from correct to incorrect.
2. The `_RESERVED_COMPLETION_KEYS` guard is no longer in the path (details inline).
3. The fallback parse is stricter than the code it replaces (details inline).
4. One comment describes the metadata precedence backwards (details inline).

Lenses run: code, errors, tests, types, comments, simplify, spec, minimal. Nits about a redundant `fallback = True` local, the tripled explanation of the merge order, and stdlib-vs-loguru in the moved helper were dropped as not worth your time.

KarinaKKarinaK:
All four review points are addressed and the branch is up to date with main:

1. The helper now sends max_completion_tokens, same as every other chat call in the repo.
2. The structural-keys guard is back inside generate_structured: extra_kwargs with model, messages or response_format raises. A user extra_body now merges into the router retry body instead of replacing it.
3. Fallback parsing is tolerant again: stray numbers get stringified and nulls dropped, so one bad item cannot drop a whole section. While testing this I also found and fixed a precedence bug in extract_json_from_response where an unfenced object with an array inside had the inner array extracted.
4. The metadata comment now describes the real precedence: a caller metadata key wins, per apply_pipeline_metadata's contract.

Each fix has a test pinning it. Full suite green.

currentlycodinng (approved):
Verified on head `07c0b0e`: `pytest tests/redteam/test_recommendations.py tests/simulation/reports/test_recommendations.py tests/simulation/test_structured_output.py tests/simulation/test_extract_json.py` → **49 passed**, CI green.

I also exercised `generate_structured` live through the orq router rather than trusting the unit mocks, since the whole premise is provider fencing behavior:

```
openai/gpt-4o-mini:          parse() ok -> 2 recs
anthropic/claude-sonnet-4-5: parse() ok -> 2 recs
google/gemini-2.5-flash:     parse() ok -> 2 recs
openai/o4-mini:              parse() ok -> 2 recs   (reasoning, at the redteam max_tokens=1500)
openai/gpt-5-mini:           parse() ok -> 2 recs   (reasoning, same budget)
```

Anthropic and Gemini — the two that used to fence and drop the section — now go through `parse()` cleanly, and the reasoning models don't exhaust the 1500-token budget on this prompt shape. So the fix does what it says on the real transport.

The earlier findings all look properly addressed: `max_completion_tokens` is back, `_STRUCTURAL_KEYS` rejects a schema-swapping `extra_kwargs`, `extra_body` merges into the router retry body instead of replacing it, and `coerce_str_list` / `coerce_str` restore the pre-RES-822 tolerance on the fallback path so a stray `1` or `null` no longer costs the whole focus area. The `BeforeValidator` annotations don't disturb the strict JSON schema the SDK derives, which the live runs confirm.

Two small things, neither blocking:

**The truncation error names a remedy that doesn't apply here.** `structured_output.py` tells the operator to "raise the budget via `EVALUATORQ_LLM_MAX_TOKENS`", but that variable is only read at `simulation/agents/base.py:75` for `DEFAULT_MAX_TOKENS`. Both migrated call sites pass a hardcoded budget (`recommendations.py:229` → 1500, simulation → 800), so setting it changes nothing on this path. The wording came along with the move, but it now lives in `common` and is reachable from red-team, where the variable never applied at all. Either make these two budgets read the env var, or name the actual knob.

**"Fail loudly" doesn't survive the call site.** The `LengthFinishReasonError` branch deliberately raises rather than falling back, but `redteam/reports/recommendations.py:251` catches `Exception` → warning → `continue`, so a truncated structured output still degrades to a dropped recommendations section — the symptom this PR set out to remove, now with a more informative log line. That may well be the right trade (a truncated result is unusable either way), but the helper's docstring and the call site currently disagree about which behavior is intended.

Cosmetic: the `fallback = True` local in the 400-fallback branch exists only to be handed to `set_attribute` — pass `True` directly.

src/evaluatorq/common/structured_output.py
```diff
@@ -0,0 +1,164 @@
...
+from evaluatorq.common.tracing import get_trace_context_headers, record_llm_response, with_llm_span
+
+logger = logging.getLogger(__name__)
+
+T = TypeVar('T', bound=BaseModel)
+
+# Structural request fields a caller's extra_kwargs may not replace: they are
+# owned by this helper, and letting extra_kwargs swap them out would silently
+# break the call it rides on (e.g. replacing the response_format schema this
+# helper exists to enforce). Same contract as LLMCallConfig.completion_params.
```
Baukebrenninkmeijer: `_RESERVED_COMPLETION_KEYS` at `src/evaluatorq/contracts.py:144` reserves four keys: `model`, `messages`, `response_format` and `extra_body`. `_STRUCTURAL_KEYS` reserves three. It leaves out `extra_body` on purpose, and the next two lines say so.

The two contracts are not the same. A reader who trusts this sentence and stops reading gets the wrong rule. Please delete the "Same contract" sentence, or write "like `LLMCallConfig.completion_params`, except that `extra_body` stays open".

src/evaluatorq/common/extract_json.py
```diff
@@ -0,0 +1,136 @@
...
+
+from __future__ import annotations
+
+import json
+import re
+
+_JSON_BLOCK_PATTERN = re.compile(r'```(?:json)?\s*\n?([\s\S]*?)\n?```', re.IGNORECASE)
+
+
+def coerce_str_list(value: object) -> object:
```
Baukebrenninkmeijer, line 11: These two validators also run on the structured `parse()` path, not only on the fallback.

`generate_structured` passes the same model as `response_format` to `client.chat.completions.parse`, so the SDK validates `message.parsed` through the same `BeforeValidator`. A field docstring in `redteam/reports/recommendations.py:151` says the tolerance keeps "the fallback" tolerant, which reads as fallback-only.

The behaviour is fine. The scope in the docstrings is not. Please say that the coercion applies to both paths.

Baukebrenninkmeijer (commented):

**4/5** — the four earlier findings are fixed and each fix has a test; three documentation points remain.

Re-reviewed `fed3f3c8...07c0b0e`. CI is green on all 13 checks. I ran `pytest tests/redteam/test_recommendations.py tests/simulation/reports/test_recommendations.py tests/simulation/test_structured_output.py tests/simulation/test_extract_json.py` on the head commit: **49 passed**. I read `apply_pipeline_metadata` at `src/evaluatorq/common/llm_call.py:112` to confirm the corrected precedence comment. I also read every `generate_structured` call site: none sends `model`, `messages` or `response_format` in `extra_kwargs`, so the new guard breaks no existing caller.

| # | Item | Resolution |
|---|------|-----------|
| 1 | The helper sent `max_tokens` | ✅ |
| 2 | `extra_kwargs` could replace `response_format` | ✅ |
| 3 | The fallback parse dropped a section on one bad item | ✅ |
| 4 | The metadata precedence comment was backwards | ✅ |

### P2

- **The comment says the two key sets are the same contract. They are not.** (`src/evaluatorq/common/structured_output.py:33`)
- **The coercing validators also run on the `parse()` path.** (`src/evaluatorq/common/extract_json.py:11`)
- **The pull request description states the old precedence.** It says "pipeline run metadata stays authoritative over `llm_kwargs`". The code and the corrected comment now say the caller wins. Please update the description.

KarinaKKarinaK:
Addressed the remaining P2 points and merged main in (branch is now level with base).

From @currentlycodinng:
- Truncation error no longer names EVALUATORQ_LLM_MAX_TOKENS. It names the real knob: the max_tokens budget passed to the call. Neither migrated call site reads that env var, so the old wording was misleading.
- Fail-loudly vs the call site: documented the real behavior in the helper docstring. It raises on length truncation (a same-budget json_object retry would truncate again), but both report call sites still wrap the call in a broad except and skip that one section, now with a clear log line instead of a silent drop. Docstring and call sites agree now.
- Dropped the throwaway fallback local; the span attribute value is passed directly.

From @Baukebrenninkmeijer:
- Reserved-keys comment fixed: the set is intentionally narrower than LLMCallConfig.completion_params (extra_body is not reserved here). No longer claims same contract.
- coerce_str_list / coerce_str docstrings now say they run on the parse() path too (no-op on clean structured output), not only the fallback.
- PR description updated: the precedence line now matches the code and apply_pipeline_metadata's contract (caller metadata key wins).

Test updated to pin the new wording. ruff clean, basedpyright 0 errors, full suite green (4113 passed). CI re-running on the new commits.

tests/simulation/test_structured_output.py
```diff
@@ -25,7 +25,7 @@ async def test_generate_structured_raises_when_parse_hits_length_limit() -> None
     client.chat.completions.parse = AsyncMock(side_effect=parse_error)
     client.chat.completions.create = AsyncMock()
 
-    with pytest.raises(RuntimeError, match="EVALUATORQ_LLM_MAX_TOKENS"):
+    with pytest.raises(RuntimeError, match="max_tokens"):
```
Baukebrenninkmeijer: The match string `max_tokens` is short and generic. Any `RuntimeError` that names a token budget passes this test. Match on `truncated` as well, so the test still identifies the truncation branch.

Baukebrenninkmeijer (approved):

**4 → 4.5/5** — the three documentation findings are fixed; one test nit remains.

Re-reviewed `07c0b0e...0e8c6d2`. All 13 checks pass on the head commit. I read `_RESERVED_COMPLETION_KEYS` and `LLMCallConfig.completion_params` at `src/evaluatorq/contracts.py:165`: that guard reserves `extra_body`, and the helper's key set is narrower, as the new comment now states. I also confirmed that the new truncation message names the `max_tokens` argument, which both call sites pass directly.

| # | Item | Resolution |
|---|------|-----------|
| 1 | The comment said the two key sets are the same contract | ✅ |
| 2 | The coercing validators also run on the `parse()` path | ✅ |
| 3 | The pull request description stated the old precedence | ✅ |

### P2

- **The new match string is generic.** (`tests/simulation/test_structured_output.py:28`)

KarinaKKarinaK:
Tightened the test match to 'Raise the max_tokens budget' so it pins the actionable remedy rather than the incidental 'max_tokens'. Pushed in 10eb6070. Thanks for the reviews.

PR #136, opened 2026-08-11
feat(simulation): add apply_suggestions to fold sim recommendations into the agent (RES-1142)

KarinaKKarinaK:
## What

Adds an `apply_suggestions` method to agent simulation (RES-1142). It takes the remediation suggestions a run already produces via `generate_recommendations` and folds them into the agent's `instructions` with an LLM, then optionally writes the revised instructions back to the orq agent.

## Behavior

- **Preview by default.** It fetches the agent and returns `original_instructions` + `new_instructions` without touching the platform. Nothing is written unless the caller passes `apply=True`.
- **Opt-in write is versioned.** When `apply=True`, the revised instructions go out via `agents.update(..., version_increment='minor')`, so the change is a new agent version and is revertable, not an in-place overwrite.
- **One merge pass.** Suggestions across all recommendations are flattened, deduped (order-preserving), and merged in a single LLM call so the prompt keeps its structure instead of getting a raw bullet list appended.
- **No-ops are safe.** No suggestions, or an empty LLM response, returns `applied=False` and writes nothing.

## Shape

```python
from evaluatorq.simulation.reports import apply_suggestions

result = await apply_suggestions(
    run.recommendations, agent_key="support-bot",
    orq_client=orq_client, llm_client=llm_client, model=model,
    apply=False,  # default: preview only
)
result.new_instructions   # proposed merge
result.applied            # False unless apply=True
result.new_version        # agent version after a write, else None
```

Lives in `simulation/reports/apply.py`, exported from `simulation.reports` alongside `generate_recommendations`. The orq SDK surface (`agents.retrieve` / `agents.update` with `instructions` + `version_increment`) was verified against the installed sdk before wiring.

## Tests

`tests/simulation/reports/test_apply_suggestions.py`: dedup/cap, preview does not write, `apply=True` writes a new version with the right kwargs, and both no-op paths. Full simulation suite (799) passes; `ruff check src`, `ruff format --check src`, and `basedpyright` (whole repo) are clean.

## Follow-up

RES-1143 mirrors this for red teaming; a CLI/`--apply` surface can come later if wanted. This PR keeps the write path library-only and opt-in.

### DIFF
--- a/src/evaluatorq/simulation/reports/apply.py
+++ b/src/evaluatorq/simulation/reports/apply.py
@@ -4,15 +4,20 @@ Takes the suggestions produced by ``reports.recommendations`` and folds them
 into the ORQ agent's ``instructions`` with an LLM, then optionally writes the
 revised instructions back as a new agent version.
 
+The flow follows the reviewed design: aggregate the raw suggestions into a
+single revised prompt (the "step in between"), expose a diff of the change for
+the user to approve, and only on approval (``apply=True``) send the update. To
+avoid re-applying the same fix, the caller passes the suggestions already
+applied to the agent (tracked on ``SimulationRun.applied_suggestions``); those
+are skipped, and the newly applied ones come back on the result to append.
+
 Preview by default: nothing is written to the platform unless ``apply=True``.
-The caller inspects ``original_instructions`` / ``new_instructions`` first and
-opts in to the write, so a run report never mutates a live agent as a side
-effect.
 """
 
 from __future__ import annotations
 
 import asyncio
+import difflib
 import functools
 from typing import TYPE_CHECKING, Any
 
@@ -24,6 +29,8 @@ from evaluatorq.simulation.utils.extract_json import extract_json_from_response
 from evaluatorq.simulation.utils.structured_output import generate_structured
 
 if TYPE_CHECKING:
+    from collections.abc import Sequence
+
     from openai import AsyncOpenAI
 
     from evaluatorq.simulation.types import SimulationRecommendation
@@ -39,20 +46,27 @@ class ApplySuggestionsResult(BaseModel):
     """Outcome of folding suggestions into an agent's instructions.
 
     ``applied`` is False in preview mode (the default): the caller gets the
-    proposed ``new_instructions`` without any platform write. ``new_version``
-    is the agent's version string after a successful write, else None.
+    proposed ``new_instructions`` plus a ``diff`` to show, without any platform
+    write. ``suggestions`` are the ones merged this call (already-applied
+    suggestions excluded); append them to ``SimulationRun.applied_suggestions``
+    after a successful write so they are not applied again. ``new_version`` is
+    the agent's version string after a successful write, else None.
     """
 
     agent_key: str
     suggestions: list[str] = Field(default_factory=list)
     original_instructions: str = ''
     new_instructions: str = ''
+    diff: str = ''
     applied: bool = False
     new_version: str | None = None
 
 
 _MAX_SUGGESTIONS = 20
-_MAX_INSTRUCTIONS_TOKENS = 2000
+# Instructions can be long; the merge returns the WHOLE revised prompt, not a
+# diff, so keep the budget generous. generate_structured raises loudly rather
+# than writing a truncated prompt if this is still hit.
+_MAX_INSTRUCTIONS_TOKENS = 4096
 
 _SYSTEM_PROMPT = """\
 You revise the system instructions of a conversational AI agent. You are given \
@@ -75,9 +89,14 @@ Respond ONLY with valid JSON. No markdown, no explanation outside the JSON."""
 def _collect_suggestions(
     recommendations: list[SimulationRecommendation],
     max_suggestions: int,
+    already_applied: Sequence[str] = (),
 ) -> list[str]:
-    """Flatten recommendations into a deduplicated, order-preserving suggestion list."""
-    seen: set[str] = set()
+    """Flatten recommendations into a deduplicated, order-preserving suggestion list.
+
+    Suggestions in ``already_applied`` are skipped so a previously applied fix is
+    not merged in again.
+    """
+    seen: set[str] = {s.strip() for s in already_applied}
     out: list[str] = []
     for rec in recommendations:
         for raw in rec.suggestions:
@@ -86,6 +105,18 @@ def _collect_suggestions(
                 seen.add(suggestion)
                 out.append(suggestion)
     return out[:max_suggestions]
+
+
+def _unified_diff(original: str, revised: str) -> str:
+    """Line-level unified diff of the instructions change, for the approval view."""
+    return ''.join(
+        difflib.unified_diff(
+            original.splitlines(keepends=True),
+            revised.splitlines(keepends=True),
+            fromfile='instructions (current)',
+            tofile='instructions (proposed)',
+        )
+    )
 
 
 def _build_user_prompt(current_instructions: str, suggestions: list[str]) -> str:
@@ -106,6 +137,7 @@ async def apply_suggestions(
     apply: bool = False,
     temperature: float = 0.0,
     max_suggestions: int = _MAX_SUGGESTIONS,
+    already_applied: Sequence[str] = (),
 ) -> ApplySuggestionsResult:
     """Fold simulation suggestions into an agent's instructions.
 
@@ -118,18 +150,23 @@ async def apply_suggestions(
         llm_client: AsyncOpenAI-compatible client for the merge call.
         model: Model identifier for the merge call.
         apply: When False (default) the agent is only read and the proposed
-            instructions are returned. When True the revised instructions are
-            written back as a new minor agent version.
+            instructions plus a diff are returned. When True the revised
+            instructions are written back as a new minor agent version.
         temperature: Sampling temperature for the merge call.
         max_suggestions: Cap on how many unique suggestions are merged.
+        already_applied: Suggestion strings previously applied to this agent
+            (typically ``SimulationRun.applied_suggestions``); they are skipped
+            so a fix is not applied twice.
 
     Returns:
-        An ``ApplySuggestionsResult``. When there are no suggestions, or the LLM
-        yields no usable text, nothing is written and ``applied`` is False.
+        An ``ApplySuggestionsResult``. When there are no new suggestions, or the
+        LLM yields no usable text, nothing is written and ``applied`` is False.
+        On a successful write, append ``result.suggestions`` to the run's
+        ``applied_suggestions`` so they are not re-applied.
     """
-    suggestions = _collect_suggestions(recommendations, max_suggestions)
+    suggestions = _collect_suggestions(recommendations, max_suggestions, already_applied)
     if not suggestions:
-        logger.info(f'No suggestions to apply for agent {agent_key!r}')
+        logger.info(f'No new suggestions to apply for agent {agent_key!r}')
         return ApplySuggestionsResult(agent_key=agent_key)
 
     agent = await asyncio.to_thread(orq_client.agents.retrieve, agent_key=agent_key)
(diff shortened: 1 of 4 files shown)


KarinaKKarinaK:
Updated to match the design from the ticket (aggregate -> diff -> approve -> update, plus applied-tracking):

- **Applied-tracking.** New `SimulationRun.applied_suggestions` holds the suggestion strings already sent to the agent. `apply_suggestions` takes `already_applied` and skips them, so a fix is never applied twice; the newly applied ones come back on `result.suggestions` for the caller to append. This is what lets the dashboard render applied ones differently.
- **Diff.** `ApplySuggestionsResult.diff` is a unified diff of the instructions change, for the approve view. Preview still writes nothing unless `apply=True`.
- **Truncation.** Raised the merge budget; `generate_structured` already raises loudly rather than writing a truncated prompt.

Two product calls I want to confirm rather than bake in silently:

1. **Field: `instructions` vs `system_prompt`.** I write the agent's `instructions` (the required behavioral field). On the orq agent model `system_prompt` is an optional template override ("if omitted, the default template is used"), so `instructions` looks right, but you said "system prompt" so flagging it. If the fixes should land in `system_prompt` instead, one-line change.
2. **In-place vs versioned.** On approve I send `agents.update(..., version_increment='minor')`, so it cuts a new revertable version rather than overwriting live. You said "just do an update request", so confirm whether you want the version bump or a plain in-place update.

Note: the approve button and diff rendering are the dashboard side; this PR is the library layer it calls. Same shape will mirror onto red teaming for RES-1143.

src/evaluatorq/simulation/reports/apply.py
```diff
@@ -0,0 +1,227 @@
...
+        suggestions=suggestions,
+        original_instructions=original,
+        new_instructions=new_instructions,
+        diff=_unified_diff(original, new_instructions),
+        applied=False,
+    )
+    if not apply:
+        return result
+
+    updated = await asyncio.to_thread(
```
Baukebrenninkmeijer, line 215: **A no-op merge still writes a new agent version.**

There's an early return when the LLM yields empty instructions, but none when it yields the *original* instructions back. If the model decides nothing needs changing — plausible when every suggestion is already reflected in the prompt — `new_instructions == original`, `diff` is the empty string, and `apply=True` still calls `agents.update` with `version_increment='minor'` and a `version_description` claiming N suggestions were applied.

The result is a version bump with an identical prompt and a misleading description, and the approval view above it would have shown an empty diff.

One line next to the existing empty-output guard:

```python
if new_instructions == original:
    logger.info(f'Revised instructions match the original for agent {agent_key!r}; nothing to write')
    return result
```

placed after `result` is built, so the caller still gets the (empty) diff to render.

src/evaluatorq/simulation/reports/apply.py
```diff
@@ -0,0 +1,227 @@
...
+    diff: str = ''
+    applied: bool = False
+    new_version: str | None = None
+
+
+_MAX_SUGGESTIONS = 20
+# Instructions can be long; the merge returns the WHOLE revised prompt, not a
+# diff, so keep the budget generous. generate_structured raises loudly rather
+# than writing a truncated prompt if this is still hit.
+_MAX_INSTRUCTIONS_TOKENS = 4096
```
Baukebrenninkmeijer, line 69: **The truncation error's remediation doesn't work from this call site.**

When the merge exceeds this budget, `generate_structured` raises with `Raise the budget via EVALUATORQ_LLM_MAX_TOKENS and retry` (structured_output.py:111). But `apply.py` never reads that variable — it always passes the hardcoded 4096 — so a user who follows that advice hits the identical error on retry.

The comment above is right that this should fail loudly rather than truncate; the issue is only that the escape hatch it points at is wired shut. `simulation/agents/base.py:75` already does the derivation:

```python
DEFAULT_MAX_TOKENS = _env_int('EVALUATORQ_LLM_MAX_TOKENS', 8192)
```

Same treatment here (keeping 4096 as the default) makes the error message honest. Worth noting a full rewritten system prompt is exactly the kind of output that runs long, so this is likely to be hit in practice.

src/evaluatorq/simulation/reports/apply.py
```diff
@@ -0,0 +1,227 @@
...
+        messages=messages,
+        response_format=_RevisedInstructions,
+        temperature=temperature,
+        max_tokens=_MAX_INSTRUCTIONS_TOKENS,
+        label='apply_suggestions',
+    )
+    if parsed is None:
+        # Fallback path: the model ignored structured output and returned a
+        # json_object body, possibly fenced.
+        parsed = _RevisedInstructions.model_validate_json(extract_json_from_response(raw))
```
Baukebrenninkmeijer, line 191: **The json_object fallback parses unguarded, and no test covers it.**

`extract_json_from_response` documents its last resort as returning the trimmed content as-is, so a model that replies with prose (or an empty body — `choices` can be empty in `generate_structured`'s fallback branch) reaches `model_validate_json` with a non-JSON string and raises `ValidationError` straight out of `apply_suggestions`.

What makes this worth a comment isn't that it raises — it's that it's inconsistent with the line right below. Empty instructions get a warning and a graceful `applied=False` result; malformed instructions get an uncaught pydantic traceback. Same failure category, two different contracts.

Either is defensible — pick one. Failing loudly is consistent with the `LengthFinishReasonError` design, in which case a line in the docstring saying so is enough. Returning gracefully matches the neighbour and `generate_recommendations` (recommendations.py:196 wraps its whole LLM-and-parse block).

Either way the branch has no test: `_stub_merge` always returns a non-None parsed model, so nothing exercises `parsed is None`. A stub returning `(None, '```json\n{"instructions": "NEW"}\n```')` would cover the fenced-response path too.

src/evaluatorq/simulation/reports/apply.py
```diff
@@ -0,0 +1,227 @@
...
+            fromfile='instructions (current)',
+            tofile='instructions (proposed)',
+        )
+    )
+
+
+def _build_user_prompt(current_instructions: str, suggestions: list[str]) -> str:
+    suggestion_lines = '\n'.join(f'- {xml_escape(s)}' for s in suggestions)
+    return (
+        f'<current_instructions>\n{xml_escape(current_instructions)}\n</current_instructions>\n\n'
```
Baukebrenninkmeijer, line 125: The `xml_escape` calls here are the prompt-injection defense the module docstring advertises, and untrusted text really does flow in — agent instructions and LLM-generated suggestions both.

No test asserts the escaping happens, so removing either call (or adding a third interpolation that forgets it) is a silent regression. Capturing `messages` off the `generate_structured` AsyncMock's `call_args` and asserting a suggestion containing `</suggestions>` comes back escaped is a few lines and pins the behaviour.

src/evaluatorq/simulation/types.py
```diff
@@ -457,6 +457,10 @@ class SimulationRun(BaseModel):
     recommendations: list[SimulationRecommendation] | None = None
     """LLM-generated remediation suggestions for remediable failures
     (see ``reports.recommendations``). None when never generated."""
+    applied_suggestions: list[str] = Field(default_factory=list)
```
Baukebrenninkmeijer, line 460: **Judgement call for @Baukebrenninkmeijer — the applied-tracking half.**

This field is inert in this PR: nothing in `src/` writes to it, and nothing reads it. The docstring says it's "written back onto the run so the dashboard renders them differently", but `simulation/ui/dashboard.py` and `reports/export_html.py` are untouched, and `apply_suggestions` leaves the append to the caller — of which there are none yet.

The skip logic itself is real and tested via the `already_applied` parameter, so "don't accidentally reapply" from Bauke's Linear note is genuinely covered. It's the "so it shows differently" half that isn't, and the field only pays for itself once someone writes to it.

Bauke — is the dashboard rendering in scope for RES-1142's DoD, or a deliberate follow-up? The PR body's "a CLI/`--apply` surface can come later" reads like the latter, in which case this is fine as-is and the docstring is just describing intent rather than current behaviour.

src/evaluatorq/simulation/reports/apply.py
```diff
@@ -0,0 +1,227 @@
...
+        applied=False,
+    )
+    if not apply:
+        return result
+
+    updated = await asyncio.to_thread(
+        functools.partial(
+            orq_client.agents.update,
+            agent_key=agent_key,
+            instructions=new_instructions,
```
Baukebrenninkmeijer, line 219: **Judgement call for @Baukebrenninkmeijer — the two questions Karina flagged as open.**

From RES-1143: "Same two open questions as on PR #136 carry over (write `instructions` vs `system_prompt`, and versioned vs in-place update)". Neither has an answer yet, and both surfaces are waiting on the same call.

Bauke's Linear reply said "we just do an update request to the agent… we already pull the live agent from orq when opening the dashboard, so we have the system prompt already" — which reads as in-place, and names *system prompt*. The code writes `instructions` with `version_increment='minor'`.

Both choices look right to me and Karina verified the field name against the installed SDK before wiring it, so this isn't a defect — but the minor bump does mean every approval click creates a version, which is a product decision rather than an implementation one. Worth a yes or no before #137 mirrors it.

src/evaluatorq/simulation/reports/apply.py
```diff
@@ -0,0 +1,227 @@
...
+    from evaluatorq.simulation.types import SimulationRecommendation
+
+
+class _RevisedInstructions(BaseModel):
+    """Schema the LLM fills with the full rewritten agent instructions."""
+
+    instructions: str = ''
+
+
+class ApplySuggestionsResult(BaseModel):
```
Baukebrenninkmeijer, line 45: **Judgement call for @Baukebrenninkmeijer — duplication with #137.**

This module and `redteam/reports/apply.py` in #137 are near-identical: `_unified_diff` is byte-for-byte the same, `ApplySuggestionsResult` and `ApplyRecommendationsResult` have the same seven fields differing only in one name, and the collect/dedup/cap/skip logic and the retrieve→merge→diff→optional-write flow match step for step. The real differences are the item name, the XML tag names, and how the merge call is made (`generate_structured` here vs. the red-team pipeline LLM config there).

That's a deliberate mirror rather than an accident — Karina said up front that #137 would mirror this once it landed, and shipping them as twins first is a reasonable way to see the shape before committing to an abstraction. So this is a question, not a finding: now that both are visible, is the shared flow worth pulling into something like `common/apply_instructions.py` with the two becoming thin wrappers over their own prompt and merge function, or do you want them kept separate so the two surfaces can diverge?

Cheaper to answer now, while both are in flight, than after they've drifted.

Baukebrenninkmeijer (commented):

Clean, well-documented module, preview-by-default is the right shape, and the tests cover the paths that matter. CI is green. Nothing here blocks the merge.

Two functional things worth fixing before it lands, and three calls that are Bauke's rather than yours — I've left them inline so they're in one place.

**Needs Bauke's input**
1. `instructions` vs `system_prompt`, and versioned vs in-place — you flagged both as open on RES-1143 and they're still unanswered. Bauke's Linear reply said "we just do an update request to the agent... we have the system prompt already"; the code writes `instructions` with `version_increment='minor'`. Your SDK check justifies the field name and the version bump is the safer default, but neither has a yes from him yet.
2. The applied-tracking half. `already_applied` skipping works, but `SimulationRun.applied_suggestions` is inert in this PR — nothing writes it and nothing reads it. Bauke's design note asked for both "doesn't reapply" and "shows differently"; only the first is here.
3. Shared shape with #137. The two modules are near-identical — `_unified_diff` is byte-for-byte the same and the result models differ only in field naming. Worth deciding now, while both are in flight, whether that gets extracted or stays duplicated.

**Worth fixing**
- A no-op merge still bumps a version (apply.py:215).
- The truncation error tells users to set an env var this call site never reads (apply.py:69).

KarinaKKarinaK:
Consolidated into #137 per review: the two apply implementations were ~90% identical, so both surfaces now share one engine in common/apply.py (commit 40460232 on that branch) with one harmonized ApplyRecommendationsResult. This branch's apply_suggestions survives as a thin wrapper with its exact public signature, and simulation gains the edits-mode fast path that landed there. Closing this PR; review continues on #137.

KarinaKKarinaK:
Closed in favor of #137, which now carries all of this: the shared apply engine in common/apply.py, the harmonized result object, and the sim dashboard UI. Nothing from this branch is lost.

PR #138, opened 2026-08-11
perf(dashboard): stop re-reading every report on each run-list load

Baukebrenninkmeijer:
## Problem

The run-list pages were slow in proportion to the size of the run store, not the page. Both `sim_overview` and `redteam_overview` read **every** report in full, then sliced 8 rows off the end. The KPI band genuinely sums over all runs, so paging can't avoid the visit — but nothing forces it to be a full read every time.

## Fix

Three layers, cheapest first:

1. **Migration-on-read.** A legacy report with no manifest gets a `.manifests/` sidecar written the first time it's scanned, so later scans build its row from the manifest alone. No migration command to run — a runs dir heals itself as it's browsed. `eq runs` gets it for free (shares `list_manifests`).
2. **Per-run stats distillers.** `_sim_run_stats` / `_redteam_run_stats` reduce a report to ~20 scalars, mtime-keyed like `read_json` so a rewrite is a miss. The report JSON is dropped once the scalars exist.
3. **Whole-store aggregate** cached against `library.fingerprint()` — file count plus newest mtime across reports **and** `.manifests/`. Manifests are swept because an in-flight run rewrites its own on every stage flush; count alone would freeze a running run's row.

Also raises the per-run cache to 4096. At 512 a 1000-run store thrashed (**1832 misses / 168 hits**), so "warm" was still re-parsing everything.

## Numbers

1000 sim runs x 20 cases x 176 KB reports:

| | before | after |
|---|---|---|
| repeat load | 448 ms | **8 ms** |
| page 2 | 448 ms | **8 ms** |
| after a new run lands | 448 ms | 116 ms (only the new report is parsed) |
| cold (first ever load, writes sidecars) | 448 ms | 1071 ms |

Fingerprint sweep itself is ~5 ms / 1000 files.

## Notes

- `library._default_roots` -> `default_roots`: metrics needs it for the cache key, and `app.py` was already importing the private name.
- Backfill skips pairwise (no `ManifestSurface` member) and errored reports, and never overwrites an existing sidecar.
- Not covered: `landing()` and search still full-scan per load; a fresh process still pays cold. Same treatment applies when either becomes the complaint.

## Tests

New: manifest backfill (incl. the broken-report skip), stats cache invalidating on rewrite, a new run appearing without restart, fingerprint reacting to both an advancing in-flight manifest and a landing report.

`ruff check src`, `ruff format --check src`, `basedpyright` (bare), `pytest -m 'not integration'` — 4060 passed.

### DIFF
--- a/src/evaluatorq/dashboard/app.py
+++ b/src/evaluatorq/dashboard/app.py
@@ -138,11 +138,11 @@ def _settings_config(roots: list[Path] | None) -> list[tuple[str, str | list[str
     masked suffix (never the full value)."""
     import os
 
-    from evaluatorq.dashboard.library import _default_roots
+    from evaluatorq.dashboard.library import default_roots
     from evaluatorq.dashboard.orq_workspace import classify_host, resolve_base_url, resolve_slug
     from evaluatorq.simulation.types import DEFAULT_MODEL
 
-    scan_roots = roots if roots is not None else _default_roots()
+    scan_roots = roots if roots is not None else default_roots()
     store_paths = [str(p) for p in scan_roots] or ['—']
     config: list[tuple[str, str | list[str]]] = [('Run stores', store_paths)]
     config.append(('Default sim model', DEFAULT_MODEL))
--- a/src/evaluatorq/dashboard/library.py
+++ b/src/evaluatorq/dashboard/library.py
@@ -13,6 +13,8 @@ from typing import TYPE_CHECKING, TypeVar
 
 from loguru import logger
 
+from evaluatorq.common.run_manifest import MANIFESTS_DIR_NAME
+
 if TYPE_CHECKING:
     from collections.abc import Callable, Iterator
 
@@ -132,7 +134,7 @@ class ReportCard:
     stage: str | None = None
 
 
-def _default_roots() -> list[Path]:
+def default_roots() -> list[Path]:
     from evaluatorq.pairwise_run import get_pairwise_runs_dir
     from evaluatorq.redteam.runner import get_runs_dir
     from evaluatorq.simulation.utils.run_store import get_sim_runs_dir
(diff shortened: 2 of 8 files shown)


Baukebrenninkmeijer:
## Review pass — 7 lenses (Codex gpt-5.6-luna, high)

Ran code quality, tests, silent failures, comments, minimality, plus a two-axis standards/spec review over `origin/main...HEAD`. Summary of what survived verification.

### Fixed in this branch — a4646b5f

`_rt_run_name()` compared the trailing parenthetical against `report.tested_agents`, which holds bare keys (`my-key`), but the runner interpolates `PreparedTarget.target` — the full target string (`agent:my-key`, see `runner.py:1373`). The common case therefore left `(agent:my-key)` sitting in the hero title, which is exactly what the split was meant to remove. My original test used a bare key and passed. Both sides now compare with the `kind:` prefix dropped; tests cover prefixed / bare / `N targets`, plus per-pill model association on a multi-agent run.

Also corrected three inaccurate comment claims (model is optional in the pill sub; the agent row now really does use `.report-hero-sub`'s `4px 0 16px` margins rather than claiming to while using 6px).

### Filed — RES-1276 (not fixed here; this is the perf commit's area)

https://linear.app/orqai/issue/RES-1276

1. **Manifest backfill degrades `eq redteam runs` / `eq sim runs`.** `_backfill_manifest()` writes `summary={'total_results': N}` — enough for the dashboard card (`library.py:214`), not enough for the CLI table, which reads `pipeline`, `tested_agents`, `total_attacks`, `vulnerability_rate` off the same summary (`redteam/cli.py:831-844`). Legacy runs used to fall through to a full read and populate all four. After the dashboard is opened once the sidecar exists, the manifest branch wins, and those columns go permanently empty — on a surface this PR never touches, with nothing in the CLI to signal it. This is the one I'd want fixed before merge; the PR body describes the migration as "`eq runs` gets it for free".

2. **`fingerprint()` sweeps files the aggregate never reads.** It globs every root `*.json` including the `01_`/`02_`/`03_` artifacts `_iter_report_files()` deliberately excludes — so it over-invalidates, and the "~5 ms per 1000 files" figure measures a different file set than the aggregate reads.

### Raised and closed — no action

- **`fingerprint()` staleness on `(count, newest_mtime)`.** Five of seven agents flagged this as a live bug. Reading the code, it's mostly theoretical: any ordinary write sets mtime to now, which advances `newest` unconditionally. It only bites on mtime-preserving restores (`cp -p`, rsync `-t`) or a backwards clock step. Worth a comment naming that ceiling, not a redesign. (Agreement across agents isn't evidence — they read the same code and made the same inference.)
- **"Delete the whole-store aggregate cache."** The argument was that removing the layer removes the staleness bug; per the above that bug is largely theoretical, so the trade reduces to losing the warm-page shortcut for an edge case. Keep it.
- **Missing docs for `default_roots()` / the hero layout.** CLAUDE.md's "docs ship in the same PR" rule targets public surface (`__all__`, CLI flags, env vars, registry members). Neither qualifies.

No scope creep found on either commit.

PR #139, opened 2026-08-11
docs: fix API reference rendering; export ORQAgentTarget and framework targets

Baukebrenninkmeijer:
## Problem

Code examples on the API reference pages rendered as plain, uncolored text.

The docstrings behind those pages used RST literal blocks — `Example::` followed by an indented block. mkdocstrings emits those with no language, so Pygments falls back to the plain-text lexer. Signatures and already-fenced blocks were highlighted correctly, which is why this was easy to miss.

## Fix

- Converted the 14 RST literal blocks to fenced code blocks: ` ```python `, plus ` ```text ` for the AI SDK stream sample and the Bradley-Terry formula.
- Un-indented the `Tracing & PII:` body in `simulation/__init__.py` — the whole env-var section was being parsed as an indented code block, not merely rendered unhighlighted.
- Swapped ` ```dotenv ` for ` ```ini ` in `docs/configuration.md`; Pygments has no `dotenv` lexer, so those blocks rendered plain for the same reason.
- Second commit: replaced `config = DynamicRunConfig(targets=[target])` in four integration Usage examples. `DynamicRunConfig` does not exist anywhere in the repo — grep returns only those four docstrings. Pre-existing, but the first commit promotes those lines into highlighted Python that reads as verified code, so they're fixed here rather than left to mislead.

The highlighting fix itself changes no behavior — docstrings and one docs page only. Two later commits do change behavior, and are described below: the `evaluatorq.integrations` recursion fix, and the `evaluatorq()` parallelism default.

## Two things a reviewer will want to know

**The de-indented body under `Example:` / `Usage:` is deliberate.** `mkdocs.yml` sets `docstring_style: google`, and `Example:` is a Google section keyword. griffe only opens a structural section when the body is *indented*, so these still parse as a single text section and the fences pass through to the Markdown renderer. Indenting them "to make it a proper Google section" would break it twice over — a 4-space-indented fence inside a `cleandoc`-ed docstring becomes a literal code block. Please don't "fix" this later.

**`pymdownx.highlight: default_lang: python` was considered and rejected.** It would have been a one-line config change and it does apply to indented blocks, but it's global and blind: it would also lex the ASCII span-hierarchy trees in `docs/tracing.md`, the Bradley-Terry formula, and the Vercel wire-format sample as Python. `guess_lang` is worse still — unreliable on short snippets and advised against by Material.

## Verification

Built the site and counted `<pre>` blocks with no Pygments spans:

- `reference/**.html`: **8 plain blocks → 0**
- Site-wide, the remaining plain blocks are deliberate: mermaid diagrams, terminal output dumps, and the span-tree diagrams in `tracing.md`.

Also audited every fence language used across `docs/`, `src/` and `examples/` against Pygments; `dotenv` was the only one with no lexer. And AST-parsed every ` ```python ` block in the diff — the six that don't parse are all deliberate elisions (`simulate(target=target, ...)` in the six `integrations/*/target.py` files).

Checks run verbatim as CI does: `ruff check src`, `ruff format --check src`, `basedpyright` (0 errors), `pytest -m 'not integration'` (4129 passed), `mkdocs build --strict`.

## Also fixed here: three public functions were missing from the docs entirely

Chasing the highlighting bug turned up a worse one on the same pages.

`evaluatorq/__init__.py` re-exports three functions from a sub-module of the same name — `from .evaluatorq import evaluatorq`, and likewise `deployment` and `llm_jury`. At runtime the import rebinds the name and the function wins. griffe resolves statically, where it does not:

```
evaluatorq.evaluatorq          -> Kind.MODULE     (function shadowed)
evaluatorq.deployment          -> Kind.MODULE     (function shadowed)
evaluatorq.llm_jury            -> Kind.MODULE     (function shadowed)
evaluatorq.llm_jury_pairwise   -> Kind.FUNCTION   (no clash, renders fine)
```

mkdocstrings then emitted nothing for them — silently, even though `gen_pages.py` pins all three in the generated `members:` list, and even though `mkdocs build --strict` stayed green. `evaluatorq()`, the library's primary entry point, had no API reference entry anywhere on the site.

`gen_pages.py` now detects the clash by asking where the runtime object is actually defined, and emits those symbols as a separate `::: pkg.mod.symbol` block addressed by their defining module. Covered by `tests/unit/test_docs_api_pages.py`.

Second gap, same investigation: `CrewAITarget` and `PydanticAITarget` were in no documented package's `__all__`, so `from evaluatorq.simulation import CrewAITarget` raised ImportError and neither class was documented (`reference/evaluatorq/integrations/` renders only the package docstring — mkdocstrings does not recurse into sub-modules). Both are now exported alongside the other four targets, with `_OPTIONAL_TARGET_EXTRAS` entries so a missing extra gives the install hint. **This commit is `feat:`** — it adds public API surface, so this PR now computes a minor release rather than no release.

Verified after both fixes — every one of these now has an anchor on a reference page:

| Symbol | Before | After |
|---|---|---|
| `evaluatorq()` | absent | `reference/evaluatorq/` |
| `deployment()` | absent | `reference/evaluatorq/` |
| `llm_jury()` | absent | `reference/evaluatorq/` |
| `CrewAITarget` | absent | `reference/evaluatorq/integrations/` |
| `PydanticAITarget` | absent | `reference/evaluatorq/integrations/` |

(Both landed on `simulation/` first; a later commit in this PR moved them — see the de-duplication note in follow-up round 2.)

## Also fixed here: `from evaluatorq.integrations import X` raised RecursionError

Not a docs bug — a live runtime one, found because `docs/gen_pages.py` carries a workaround for it.

The lazy loader answered every name with `from . import <name>` inside `__getattr__`. That form is resolved by `_handle_fromlist`, which calls getattr on the package — re-entering `__getattr__`, which runs `from . import <name>` again:

```
>>> from evaluatorq.integrations import langgraph_integration
RecursionError: maximum recursion depth exceeded
```

All six lazy sub-modules, on Python 3.13. Only the fully-qualified form (`from evaluatorq.integrations.langgraph_integration import LangGraphTarget`) worked — so the usage the package's own docstring advertises was broken.

`importlib.import_module` goes through the import machinery instead of the getattr path and binds the sub-module on the package, so later lookups never re-enter `__getattr__`. Six near-identical branches collapse to one `__all__` membership check (24 lines deleted, 9 added). Laziness verified unchanged: importing the package still pulls in no optional dependency.

The regression test fails on 6 of 7 sub-modules against the old implementation and passes on all 7 here. (`langchain_integration` was the exception at this commit — imported eagerly at module top, so it never reached `__getattr__`. A later commit in this PR made it lazy too, for a separate reason: the eager import made `import evaluatorq.integrations` require langchain.)

Also reverted the `_OPTIONAL_TARGET_EXTRAS` entries added for `CrewAITarget` / `PydanticAITarget` a commit earlier. Both targets are duck-typed and import neither `crewai` nor `pydantic_ai` at runtime — verified by blocking each library and importing the target anyway — so the missing-extra branch can never fire for them. The entries, and the tests I had written for them, asserted unreachable behavior. (The pre-existing `LangGraphTarget: langgraph` entry is dead for the same reason; left alone as out of scope.)

## Known follow-up, deliberately not in this PR

**1. Sphinx cross-reference roles render as literal text** — tracked as [RES-1278](https://linear.app/orqai/issue/RES-1278/docs-sphinx-cross-reference-roles-render-as-literal-text-on-the-api). Pages show `:class:` and `:func:` prefixes verbatim beside a code span that links nowhere: 68 occurrences rendered today, 228 role usages across 57 files in `src/`. Needs a decision between stripping the prefixes and converting to autorefs syntax, so it does not belong in a highlighting PR.

---

## Follow-up round (post-review)

Four more commits after an adversarial review pass:

- **`docs: drop Sphinx roles that render as literal text`** — 227 `:class:`/`:func:`/`:attr:`/`:meth:`/`:mod:`/`:data:`/`:ref:` roles across 57 files rendered verbatim on the reference pages, one line from the code blocks this PR colorized. Stripped the prefix, kept the backticks. Autorefs would need 227 dotted paths resolved by hand with `--strict` failing on any miss, for links that never worked. **0 role occurrences left in the built site.**
- **`fix(redteam): export ORQAgentTarget`** — same defect class as the `CrewAITarget`/`PydanticAITarget` fix, missed on the first pass. Plus a test walking all of `evaluatorq.redteam.__all__`, since F822 can't catch it in a module with `__getattr__`.
- **`docs: keep gen_pages importable without regenerating the site`** — the new test's import ran all three generators at module scope. Note `runpy.run_path` names the file `<run_path>`, not `__main__`, so a plain `__main__` guard would have silently emptied the site.

**On the module-shadowing fix:** the review argued for renaming `evaluatorq.py`/`deployment.py`/`llm_jury.py` at source instead. Checked — at runtime all three names stay bound to their functions (the `from .x import x` rebind wins, and later `import_module` calls hit `sys.modules` without re-binding), and basedpyright resolves all three to functions. Only griffe's static resolution is affected, so the fix stays in `gen_pages.py`. Recorded in the docstring so it isn't re-litigated.

---

## Behavior change: `evaluatorq()` now defaults to `parallelism=10`

`feat: default evaluatorq() parallelism to 10` — `src/evaluatorq/evaluatorq.py:107` and `src/evaluatorq/types.py:248`.

The old default was `1`, i.e. every evaluation ran strictly sequentially unless the caller knew to opt out. That is the wrong default for a library whose work is almost entirely provider-bound I/O: the common case paid a latency penalty to protect the uncommon one.

**This is a real behavior change for existing callers who omit `parallelism`.** Ten jobs now run concurrently, which matters in two cases:

- a provider that rate-limits at low concurrency — pass `parallelism=1` (or any lower value) to restore the old behavior;
- jobs that mutate shared, non-thread-safe state — these were previously serialized by accident, not by design.

Both are called out in the docstring. Kept deliberately in this PR rather than split out.

## Follow-up round 2 (post-review)

A second adversarial pass (six independent review lenses) found the gates added earlier on this branch each had a way to pass while the defect they exist to catch was present. Two commits:

**`docs: close the escape hatches in the rendering gates`**

- `docs/hooks.py` exempted any unhighlighted block whose first word was `graph`/`flowchart`/etc., so an unfenced Python sample starting `graph = {}` bought itself an exemption from the very check. Matches a mermaid *header line* now.
- The Sphinx-role test used a fixed nine-role allowlist — `:command:` or `:type:` would render literally and stay green. Matched generically; the tree is still clean under the wider pattern.
- The fencing test recognised only `Example:`. Adding `Usage:` turned up three genuinely unfenced blocks (`cli.py`, `simulation/cli.py`, `common/reports/palette.py`), now fenced.
- `test_gen_pages_run_gate` asserted only that filenames were generated — dropping the `::: evaluatorq.evaluatorq.evaluatorq` blocks left every page present and `--strict` green while the entry points vanished. Asserts on the page body now, mutation-verified.
- The simulation export test checked `hasattr` only; a lazy mapping handing back the wrong class would pass. Asserts identity against the canonical integration module.
- Two subprocess tests ran without a timeout.

**`fix(integrations): import every sub-module lazily`**

- `from . import langchain_integration` ran at package import, and `langchain_integration.convert` imports `langchain_core` at module level — so `import evaluatorq.integrations`, the package whose whole job is making integration dependencies optional, raised `ModuleNotFoundError` unless langchain was installed. All seven sub-modules go through `__getattr__` now.
- `CrewAITarget` and `PydanticAITarget` were rendered **in full on two pages**. `evaluatorq/integrations/` documents each sub-module's `__all__`, and `evaluatorq/simulation/` kept them too: a lazy re-export is in neither `vars()` nor `sys.modules`, so `_safe_getattr` returned `None`, the owner read as `""`, and the "document each symbol once" rule never fired. A name is now dropped only when a sub-module page is *known* to render it — deliberately narrow, because dropping on ownership alone took `ReplayError` and `OrqResponsesTarget` off the site entirely.
- Uppercase exports were filtered off every page, so `DEFAULT_MODEL` and `OWASP_LLM_TOP_10` had no reference entry at all. Sorted to the bottom of the page instead; the concern was prompt blobs *opening* a page, not their existing.
- The new test generates in a subprocess: `_safe_getattr` reads `vars()` first, so once another test touches `evaluatorq.simulation.TokenUsage` the member lists shift. mkdocs generates in a fresh interpreter, so that is what to assert against.

### Known gap, pinned rather than fixed

Eight symbols are documented on no reference page at all: `AgentContext`, `JuryResult`, `JuryStats`, `JuryVote`, `KnowledgeBaseInfo`, `MemoryStoreInfo`, `ReasoningOutputItem`, `ToolInfo` — all re-exported by `evaluatorq.redteam` from `evaluatorq.contracts`, which `evaluatorq.__all__` does not list, so the owner's page renders nothing for them and the re-exporting page drops them. Same defect class as the `evaluatorq()`/`deployment()` gap above.

Restoring them makes autorefs prefer the redteam page for every `AgentTarget`/`AgentContext` cross-reference and emit anchors mkdocstrings never wrote (it anchors re-exports under the defining module), failing `--strict` in two separate places. That needs its own fix. The set is pinned in `test_gen_pages_run_gate.py` so it cannot grow silently.


currentlycodinng (approved):
Verified on head `62f94e5`: `pytest tests/unit/test_docs_api_pages.py tests/redteam/test_public_exports.py tests/simulation/test_framework_targets_exposed.py` → **113 passed, 1 skipped**, and `mkdocs build` completes clean. The diagnosis is right (mkdocstrings emits RST literal blocks with no language, Pygments falls back to plain text), and the reasoning for *not* setting `default_lang: python` is the correct call — it would have lexed the span-hierarchy trees and the Bradley-Terry formula as Python.

Two things:

**The `reference/**` count is 8 → 1, not 8 → 0.** I built the site and scanned every `<pre>` for Pygments spans (the output is minified, so unquoted `class=kn` attributes — worth noting if your counter looked for `class="`). One plain block survives, in `reference/evaluatorq/index.html`:

```
# Using keyword arguments (recommended):
await evaluatorq("name", data=[...], jobs=[...], parallelism=5)
```

Source is `src/evaluatorq/evaluatorq.py:124`. It's a bare indented block after "Can be called with either a params dict/object or keyword arguments:" — no `Example::` marker, so the RST-literal-block sweep didn't catch it. It's Python, it's on a reference page, and it's exactly the symptom this PR fixes. Slightly ironic that it's on the `evaluatorq()` page, which only exists because of the `_submodule_shadowed` work in this same PR.

For reference, the other 10 site-wide plain blocks are all the deliberate ones you named: 4 mermaid diagrams, the terminal-output dump on the landing page, and the span trees.

**"No behavior changes — docstrings and one docs page only" isn't accurate.** This PR also rewrites `integrations/__init__.py`'s `__getattr__` from six explicit `from . import X` branches to `importlib.import_module` gated on `__all__`, adds shadowed-symbol handling to `docs/gen_pages.py`, and gitignores the generated docs tree. The `__getattr__` change is correct — every name in `__all__` is a real sub-module, so an unknown name still raises `AttributeError` and a missing optional dep still raises `ImportError`, same as before, and `test_public_exports` covers it. But a reviewer who reads "docstrings and one docs page only" will skip the one runtime file in the diff. Worth a line in the description.

The `docs/reference/`, `docs/examples/` etc. gitignore entries are safe — none of those paths are tracked today, so nothing goes stale behind the ignore.

PR #140, opened 2026-08-11
fix(dashboard): backfill full manifest summary; scope fingerprint to reports

Baukebrenninkmeijer:
Follow-up to #138 review findings ([RES-1276](https://linear.app/orqai/issue/RES-1276/dashboard-manifest-backfill-degrades-eq-runs-rows-fingerprint-sweeps)).

## 1. Manifest backfill degraded `eq runs` rows

`_backfill_manifest()` wrote a sidecar whose summary was `{'total_results': N}` — enough for the dashboard card, not for the CLI tables, which read the same summary. Once the dashboard had been opened, `eq redteam runs` lost pipeline / tested_agents / vulnerability_rate for that run, and `eq sim runs` lost mode / target kind / scorer averages. Silent, and only for runs the dashboard had seen.

- `RedTeamReport.manifest_summary()` and `SimulationRun.manifest_summary()` are now the single definition of the summary shape. The live runners and the dashboard backfill both call it, so a backfilled row is field-identical to one the run wrote itself.
- `summary_is_complete()` (in `common/run_manifest.py`) treats a summary carrying only `total_results` as incomplete. `list_run_records()` demotes such a record to the legacy full-report row (so the CLI never renders half-blank columns), and `scan()` re-reads the report and rewrites the sidecar — thin sidecars already on disk heal on the next dashboard visit.

## 2. `fingerprint()` swept files no aggregate reads

It globbed every root `*.json`, including the `01_`/`02_`/`03_` stage artifacts `_iter_report_files()` deliberately excludes. Rewriting an artifact invalidated the whole-store cache for nothing. It now sweeps through `_iter_report_files()`, so one predicate decides what counts as a report. The "~5 ms per 1000 files" figure in #138 measured that wrong file set — claim dropped rather than re-measured.

## Tests

- `eq redteam runs --json` stats row is unchanged across a dashboard `scan()` that backfills the sidecar (the acceptance test for the regression).
- Backfilled sidecar summary equals `RedTeamReport.manifest_summary()`.
- A thin sidecar is rewritten by `scan()` and demoted to a full-read row by `list_run_records()`.
- Rewriting a `01_` artifact does not change the fingerprint.

Existing manifest-first fixtures were updated from `{'total_results': N}` to a realistic writer summary — a thin summary is now, by design, not manifest-servable.

`ruff check` / `ruff format --check` on `src`, bare `basedpyright`, and `pytest -m 'not integration'` (4072 passed) all clean.

### DIFF
--- a/src/evaluatorq/common/run_manifest.py
+++ b/src/evaluatorq/common/run_manifest.py
@@ -28,7 +28,7 @@ from typing import TYPE_CHECKING, Any
 from loguru import logger
 
 from evaluatorq.contracts import (
-    RUN_SUMMARY_TOTAL_KEY,
+    RUN_SUMMARY_VERSION,
     ManifestStatus,
     RunManifest,
     RunSummary,
@@ -136,9 +136,12 @@ class ManifestWriter:
         if report_path is not None:
             self.manifest.report_path = str(report_path)
         # Compact headline stats so a run-list row can be built from the manifest
-        # alone — no full-report read needed for completed runs.
+        # alone — no full-report read needed for completed runs. Stamped with the
+        # shape's version here, the one place a summary is stored, so no writer
+        # can persist an unversioned (or stale-shaped) one.
         if summary is not None:
             self.manifest.summary = summary
+            self.manifest.summary_version = RUN_SUMMARY_VERSION
         self.flush()
 
     def cancel(self) -> None:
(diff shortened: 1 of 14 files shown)


currentlycodinng (approved):
Verified locally on head `a9dc559`: `uv sync --all-extras` + `pytest tests/dashboard tests/common -m 'not integration'` → **973 passed**. CI green across all four Python versions.

The fix is the right shape. Collapsing the summary into `manifest_summary()` on the report models means the two writers (`redteam/runner.py:1133`, `simulation/api.py:1339`) and the dashboard backfill can't drift — I grepped for any other `.complete(` / `summary={` call site and those two are the only ones, so the "single source of truth" claim actually holds. Routing `scan()` through `list_run_records()` so the dashboard and the CLI tables share one listing is a better fix than patching the summary shape in place, and dropping the "~5 ms per 1000 files" number rather than re-measuring it against the corrected file set is the honest call.

Two observations, neither blocking:

**A completed manifest with a thin summary whose report file is gone now disappears from every listing.** `list_run_records` `continue`s past it to leave the report uncovered, and the legacy glob then finds nothing to carry it:

```
report exists? False
list_run_records -> []
scan cards -> 0
```

Before this PR it listed from the manifest (with the blank columns this PR is fixing). Arguably an improvement — a card whose report is deleted only 404s on click — but it's a silent listing change that isn't in the docstring or the tests, and thin sidecars on disk are exactly the population this PR is about. Worth one line in `summary_is_complete`'s docstring at minimum.

**A backfill that keeps failing re-pays its cost on every scan.** `_backfill_manifest` now does a full `ADAPTERS[...].load(path)` in addition to the `_card` read. On success that's the documented one-off migration price. On a schema mismatch it returns after the `logger.warning`, no sidecar is written, and the next scan does the same full validate and logs the same warning again — permanently slower scans plus a warning per report per scan for any runs dir holding old-schema reports. A marker (even an empty-summary sidecar with a `backfill_failed` status) would stop the retry loop.

Nit: `scan()`'s docstring still describes the algorithm it used to implement itself ("Each `.manifests/*.json` sidecar becomes a card...", "Report files already covered by a manifest's `report_path` are de-duplicated out") — that logic now lives in `list_run_records`, and the docstring should point there rather than restate it.

PR #141, opened 2026-08-12
feat(simulation,redteam): harmonize recommendations config across surfaces (RES-1286)

KarinaKKarinaK:
## What

Makes recommendation generation configurable on **both** surfaces through one flag, `recommendations=` (RES-1286). Previously the limits were module constants in `simulation/reports/recommendations.py` and literals in `redteam/reports/recommendations.py`; there was no way to tune them without editing the package.

The flag reads the same way everywhere:

| value | meaning |
|---|---|
| `True` | on, with defaults |
| `False` | skip the LLM call |
| a config instance | on, tuned |

`red_team(generate_recommendations=...)` still works and warns; `recommendations=` is the spelling going forward.

## How

**One shared base, two surface configs.** `common/recommendations.py` holds `RecommendationConfigBase` (`max_suggestions`, `max_tokens` — the knobs that mean the same thing on both surfaces) plus `resolve_recommendations()`, the single place `True`/`False`/instance turns into "a config or `None`". Each surface subclasses it with what only it has, rather than one wide model where half the fields are inert:

- `RedTeamRecommendationConfig` (`evaluatorq.redteam`) — `max_areas`, `max_attacks`, `max_attack_chars`, `max_explanation_chars`, and overrides `max_suggestions=5` / `max_tokens=1500`.
- `SimulationRecommendationConfig` (`evaluatorq.simulation.reports`) — the three judge-metric thresholds, `max_results`, `max_transcript_chars`, `max_message_chars`.

The names are distinct on purpose: two public classes both called `RecommendationConfig` fail late and confusingly when a caller imports the wrong one.

**Every previously hardcoded limit is now a field.** The analysis calls' `max_tokens` (800 / 1500), the red-team prompt's 500- and 300-char truncations, simulation's 400-char per-message budget, and the `max_results=10` cap. Red teaming also *enforces* its suggestion cap on the reply instead of only asking for "3-5" in the prompt. `extra='forbid'` on both configs, so a misspelled knob raises instead of silently falling back to the default.

**Reachable from the SDK, not just the CLI.** `simulate()` and `generate_and_simulate()` take `recommendations=` too. It defaults to `False` there because the returned `SimulationResult` list has nowhere to carry suggestions — they land on the run, so they surface via `save=True` / `report=` or the dashboard. That mirrors how `executive_summary=` already works. Simulation generation moved into `_simulate_core` (replacing the CLI-only `post_run` hook) so the analysis spans nest under the simulation root and carry the run metadata.

Defaults are unchanged throughout: behaviour is identical unless a caller opts in.

## Tests

`tests/common/test_recommendations_flag.py` covers the flag resolving identically on both surfaces, unknown-field rejection, the deprecated `generate_recommendations=` alias (warning + still functional), and the public `simulate()` / `generate_and_simulate()` forwarding. Per-surface tests assert the new knobs actually reach the LLM call: token budget, suggestion cap (requested *and* enforced), and prompt truncation budgets.

`ruff check src`, `ruff format --check src`, `basedpyright` (whole repo) and `pytest -m 'not integration'` are clean; `mkdocs build --strict` passes.

## Docs

`docs/faq.md` documents both configs, the full default table, and the `simulate(save=True, recommendations=...)` path.

Ticket: RES-1286.


src/evaluatorq/common/env_config.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+
+from __future__ import annotations
+
+import os
+
+from loguru import logger
+
+
+def env_int(name: str, default: int) -> int:
```
Baukebrenninkmeijer: These two readers warn on a value they cannot parse, but they accept any value they can parse.

Set `EVALUATORQ_RECOMMENDATION_MAX_SUGGESTIONS=0`. `_MAX_SUGGESTIONS` becomes `0`. The system prompt at `src/evaluatorq/simulation/reports/recommendations.py:104` then asks the model for "a list of 1-0 concise, actionable strings". The LLM call still runs and is still billed. Then `raw[:0]` at line 182 is always empty, `if not suggestions: continue` at line 183 drops the result, and no log line explains the empty report. A negative value drops items off the end of the list instead.

`EVALUATORQ_RECOMMENDATION_MAX_TRANSCRIPT_CHARS=0` has the same shape: `_truncate` returns the string `'...'`, and the model is asked to judge that.

The module docstring says this mirrors the `EVALUATORQ_SPAN_MAX_TEXT_CHARS` pattern. That reader, at `src/evaluatorq/tracing/setup.py:51`, rejects a non-positive value with a warning and returns the default. Please add the same check here, or add a `minimum` argument for the two count knobs. The threshold knobs have the same gap outside the range 0 to 1.

CLAUDE.md
```diff
@@ -267,6 +267,12 @@ src/evaluatorq/
 - `OPENAI_API_KEY` — for direct OpenAI backend or pipeline LLM calls
 - `ORQ_WORKSPACE` (or `ORQ_WORKSPACE_SLUG`) — workspace slug for dashboard→Orq trace deep-links; buttons hidden when unset
 - `ORQ_UI_BASE_URL` — optional Orq UI base for deep-links (defaults to `ORQ_BASE_URL` or `https://my.orq.ai`)
+- Simulation recommendation limits (all optional, defaults preserve prior behaviour):
+  - `EVALUATORQ_RECOMMENDATION_MAX_SUGGESTIONS`: max suggestions per result (default 3)
```
Baukebrenninkmeijer: `docs/configuration.md` holds the environment-variable table that ships on the documentation site. It already carries rows for the other simulation-only knobs, such as `EVALUATORQ_LLM_TIMEOUT_S` and `EVALUATORQ_REASONING_EFFORT`, each marked **Simulation only**.

The five new variables are absent there. Please add five rows to that table as well.

src/evaluatorq/common/env_config.py
```diff
@@ -0,0 +1,37 @@
...
+    """Return the ``int`` value of env var ``name``, or ``default`` if unset/invalid."""
+    raw = os.getenv(name)
+    if raw is None or raw.strip() == '':
+        return default
```
Baukebrenninkmeijer: This module is the fourth environment reader in the package, and the third contract:

- `src/evaluatorq/simulation/agents/base.py:47` raises `ValueError` on a bad value.
- `src/evaluatorq/tracing/setup.py:51` warns, rejects non-positive values, and returns the default.
- `src/evaluatorq/common/tracing.py:41` caches the read with `functools.lru_cache`.
- This module warns and returns the default.

Moving the other three is out of scope for RES-1286. Please pick one of the existing contracts for the new module, and open a follow-up ticket to retire the private copies. A reader who meets two of these cannot tell which rule applies.

tests/common/test_env_config.py
```diff
@@ -0,0 +1,49 @@
...
+    monkeypatch.setenv('EVALUATORQ_RECOMMENDATION_MAX_SUGGESTIONS', '5')
+    monkeypatch.setenv('EVALUATORQ_RECOMMENDATION_MAX_TRANSCRIPT_CHARS', '1000')
+    monkeypatch.setenv('EVALUATORQ_RECOMMENDATION_FACTUAL_ACCURACY_BELOW', '0.3')
+    from evaluatorq.simulation.reports import recommendations
+
+    reloaded = importlib.reload(recommendations)
+    try:
+        assert reloaded._MAX_SUGGESTIONS == 5
+        assert reloaded._MAX_TRANSCRIPT_CHARS == 1000
+        assert reloaded.FACTUAL_ACCURACY_BELOW == 0.3
```
Baukebrenninkmeijer: The test asserts `FACTUAL_ACCURACY_BELOW` only. `HALLUCINATION_RISK_ABOVE` and `TONE_APPROPRIATENESS_BELOW` are never set and never asserted.

The three `env_float` calls at `src/evaluatorq/simulation/reports/recommendations.py:37-39` are near-identical lines. A copy-paste error that points `TONE_APPROPRIATENESS_BELOW` at the `HALLUCINATION_RISK_ABOVE` variable passes this suite. Two more `assert` lines close the gap.

Baukebrenninkmeijer (commented):

**3.5/5** — the change does what the ticket asks and the defaults are safe, but the new readers accept values that turn the feature off without a warning, and the user-facing environment table is not updated.

CI is green on all 13 checks. I ran `pytest tests/common/test_env_config.py tests/simulation/reports/test_recommendations.py` in one process, in both orders: **22 passed** each time. The `importlib.reload` in the new test does not leak into the other tests. I read `docs/configuration.md` and confirmed that the file already states that environment variables are read at import, so reading the five constants at import time follows the house rule.

### P1

1. **`env_int` and `env_float` accept values that silently disable the feature.** (`src/evaluatorq/common/env_config.py:16`) — `EVALUATORQ_RECOMMENDATION_MAX_SUGGESTIONS=0` sends the prompt "a list of 1-0 concise, actionable strings", pays for the LLM call, then drops every result. No log line says why.
2. **`docs/configuration.md` does not list the five new variables.** — That table is the published reference for users. `CLAUDE.md` is the instruction file for agents. A user who reads the documentation site will not find these knobs.

### P2

- **The repository now holds four environment readers.** (`src/evaluatorq/common/env_config.py:16`)
- **The test asserts one threshold out of three.** (`tests/common/test_env_config.py:45`)

currentlycodinng (commented):
Small and clean, and I confirmed the safety claim: defaults are byte-identical, the 33 tests touching `env_config` and the recommendation path pass, and the reload test leaves nothing behind. Not approving yet because of two gaps, both cheap to close.

## The pattern this mirrors does two things this does not

The PR says it mirrors `EVALUATORQ_SPAN_MAX_TEXT_CHARS`. That reader, in `common/tracing.py`, is a `@functools.lru_cache`'d function that reads on first use and clamps its result:

```python
    # Any non-positive value (including -1) disables truncation.
    return value if value > 0 else None
```

Both of those matter here, and dropping them is where the two issues come from.

**Nonsense values pass straight through and fail silently.** `env_int` and `env_float` validate parseability only, so any parseable value is accepted no matter what it means. I set three and reloaded the module:

```
MAX_SUGGESTIONS = 0            -> raw[:0]         -> [] suggestions, after the LLM call is paid for
MAX_TRANSCRIPT_CHARS = -20     -> _truncate(t,-20) -> 100 chars becomes 83: 'xxxxxxxx...'
FACTUAL_ACCURACY_BELOW = inf   -> 0.99 < inf      -> every result marked remediable
```

The transcript one is the nastiest. `_truncate` does `text[:max_chars] + '...'`, so a negative budget silently drops characters off the *end* of the transcript, which is exactly where a simulation failure shows up, and the output still looks correctly truncated. `nan` on any of the three thresholds parses fine too and makes every comparison `False`, quietly disabling that trigger. A `min_value` argument on the two helpers, or a clamp at the five call sites, closes all of this. Since the whole point of the change is to hand these knobs to users, the knobs should refuse a setting that cannot mean anything.

**The vars are read at import, so setting them from Python does not work.** Because the five constants are evaluated at module import rather than on first use, `os.environ['EVALUATORQ_RECOMMENDATION_MAX_SUGGESTIONS'] = '5'` after `import evaluatorq` has no effect. Your own test has to reach for `importlib.reload` to observe the override, which is the tell. Shell-level configuration works fine, so this is not a bug, but the lazy `lru_cache` shape from `tracing.py` would make the Python path work too and would let the test drop the reload dance.

## Users cannot find these

`docs/configuration.md` is the user-facing env var table, and it already lists `EVALUATORQ_SPAN_MAX_TEXT_CHARS` alongside `EVALUATORQ_LLM_MAX_TOKENS` and the rest of the seven. The five new vars went only into `CLAUDE.md`, which is maintainer notes. A ticket about letting teams tune without code changes should land its five rows in the table the README points people at.

## Otherwise correct

The three trigger predicates read the module globals inside lambdas rather than binding them at definition, and nothing outside the module imports any of the five names, so reload genuinely re-binds everything and the test is not proving less than it looks like it is. Empty-string handling (`raw.strip() == ''`) is right, and the warning-then-default behaviour on a malformed value is the correct choice for this kind of knob.

KarinaKKarinaK:
Both reviews addressed in a07be487.

Nonsense values are now refused instead of silently accepted: env_int gained min_value, env_float gained min_value/max_value plus a non-finite check, and out-of-range values fall back to the default with a warning, same as malformed ones. The five call sites declare their ranges: the three thresholds are 0 to 1, suggestions minimum 1 (a cap of 0 paid for the LLM call and dropped everything), transcript budget minimum 100 (a negative value truncated from the wrong end, exactly as you showed). inf and nan are rejected for the thresholds.

The five variables are now in docs/configuration.md next to EVALUATORQ_SPAN_MAX_TEXT_CHARS, so users can actually find them; CLAUDE.md keeps its copy for agents.

Tests: all three thresholds asserted (not just one), plus the exact nonsense cases from the review (0 suggestions, negative transcript chars, nan threshold) proving fallback. 25 pass in the two touched files, full suite 4,091 green.

On the lazy lru_cache read: left as is for now since shell-level configuration is the documented path and the import-time read follows the existing house rule in docs/configuration.md; happy to switch the readers to the tracing.py shape in a follow-up if you want the Python-level override path to work too.

src/evaluatorq/common/env_config.py
KarinaKKarinaK: Fixed in a07be487. env_int now takes min_value and env_float takes min_value/max_value, plus a non-finite check, so 0 suggestions, negative budgets, and nan/inf thresholds all fall back to the default with a warning.

CLAUDE.md
KarinaKKarinaK: Fixed in a07be487. Added rows for all five variables to the table in docs/configuration.md.

tests/common/test_env_config.py
KarinaKKarinaK: Fixed in a07be487. The test now sets and asserts all three thresholds, not just FACTUAL_ACCURACY_BELOW.

src/evaluatorq/common/env_config.py
KarinaKKarinaK: Good call. Opened RES-1297 to consolidate the private env readers on common/env_config.

src/evaluatorq/common/env_config.py
```diff
@@ -0,0 +1,61 @@
...
+
+from __future__ import annotations
+
+import math
+import os
+
+from loguru import logger
+
+
+def env_int(name: str, default: int, *, min_value: int | None = None) -> int:
```
Baukebrenninkmeijer: `env_int` takes only `min_value`. `env_float` takes `min_value` and `max_value`.

Set `EVALUATORQ_RECOMMENDATION_MAX_TRANSCRIPT_CHARS=1000000`. The value is above the minimum `100`, so the guard accepts it. Every recommendation prompt then carries the full transcript, and the cost of each call increases with no warning.

Add an optional `max_value` parameter to `env_int`. Pass an upper bound at the two `env_int` call sites.

src/evaluatorq/simulation/reports/recommendations.py
```diff
@@ -45,14 +46,20 @@ class _SuggestionsLLMResponse(BaseModel):
...
-# Fixed thresholds; make them parameters if teams need tuning.
-FACTUAL_ACCURACY_BELOW = 0.5
-HALLUCINATION_RISK_ABOVE = 0.5
-TONE_APPROPRIATENESS_BELOW = 0.5
-
-_MAX_TRANSCRIPT_CHARS = 3000
-_MAX_SUGGESTIONS = 3
+# these marks the result as remediable. Defaults preserve prior behaviour; override any of them via
+# the EVALUATORQ_RECOMMENDATION_* env vars so teams can tune without code changes.
+FACTUAL_ACCURACY_BELOW = env_float(
```
Baukebrenninkmeijer: The range `[0.0, 1.0]` includes the two values that make a trigger unsatisfiable.

Set `EVALUATORQ_RECOMMENDATION_FACTUAL_ACCURACY_BELOW=0.0`. The predicate at line 101 is `v < FACTUAL_ACCURACY_BELOW`, and a judge score is never below `0.0`. The trigger never fires, and no log line says why. `EVALUATORQ_RECOMMENDATION_HALLUCINATION_RISK_ABOVE=1.0` has the same effect.

This is the same shape as the last review's item 1, and it is much smaller: the user asks for the value. Log a warning when a threshold can never be satisfied, or write in `docs/configuration.md` that `0.0` and `1.0` disable the trigger.

Baukebrenninkmeijer (commented):

**3.5 → 4/5** — both items from the last review are closed; two small gaps in the new guards remain.

CI is green on all 13 checks. I read the new guards in `env_config.py` and checked all five call sites in `recommendations.py`. `docs/configuration.md` now lists the five variables, and the wording of each row agrees with the guard it describes. The new tests reject `0`, `-20`, `nan`, `inf`, `1.5` and `-0.1`, and they assert all three thresholds.

| # | Item | Resolution |
|---|------|-----------|
| 1 | P1 `env_int` and `env_float` accept values that disable the feature | ✅ |
| 2 | P1 `docs/configuration.md` does not list the five new variables | ✅ |
| 3 | P2 the test asserts one threshold out of three | ✅ |

### P2

- **`env_int` has no `max_value` parameter, but `env_float` has one.** (`src/evaluatorq/common/env_config.py:18`) — a very large value passes the minimum check.
- **The values `0.0` and `1.0` pass the range check and stop a trigger.** (`src/evaluatorq/simulation/reports/recommendations.py:51`) — the fall-back is silent for a setting that can never fire.

src/evaluatorq/common/env_config.py
Baukebrenninkmeijer: Obsolete after the redesign: `env_config.py` is deleted along with the whole `EVALUATORQ_RECOMMENDATION_*` reader (bf3e88ee). The transcript budget is now `SimulationRecommendationConfig.max_transcript_chars`, a per-call argument rather than an ambient env var, so a huge value is something a caller writes deliberately at the call site instead of something the process inherits. I left it without an upper bound on purpose — any ceiling I picked would be arbitrary, and `max_results` × `max_transcript_chars` is the real cost bound, both visible in the same constructor.

src/evaluatorq/simulation/reports/recommendations.py
Baukebrenninkmeijer: Still applied, in e59b41ec: the endpoints are a legitimate "ignore this metric" switch (and now an explicit argument, not an env var), so I documented the semantics rather than warning about a value the caller deliberately wrote. Each threshold field says so in its docstring, `docs/faq.md` spells out all three, and `test_threshold_endpoints_disable_their_check` pins the behaviour to the docs.

PR #142, opened 2026-08-12
docs: rewrite the README as a landing page (RES-1228)

Baukebrenninkmeijer:
Closes RES-1228.

The README was a 1029-line manual: it opened by positioning evaluatorq against Orq's own experiment runner — framing that means nothing to a stranger — and buried the runnable example under configuration reference. It is now a 274-line landing page.

## What changed

- **Leads with the pitch.** "Find out how your AI agent breaks — before your users do," then the three questions no test suite answers, then one install line and one runnable example.
- **One short section per surface** — evaluation, red teaming, agent simulation, dashboard — each linking its docs-site guide.
- **The `eq` CLI is mentioned.** The old README had zero mentions of it.
- **Banner, badges and a centred nav row**, promptfoo-style, plus a hero image and a screenshot per surface section.
- **Real numbers, not adjectives.** 19 OWASP categories / 18 vulnerabilities / 45 curated strategies / 16 delivery methods / 18 judges, all read off the registries rather than estimated, plus a coverage table in the red-teaming guide marking which categories get strategies generated per run.
- **Measured cost.** A table of wall clock, tokens and billed cost taken from real Orq traces, so the first question anyone asks about a red-team run has an answer.

## Nothing was orphaned

Everything cut from the README that had no docs-site home got one:

| Cut from README | New home |
|---|---|
| `evaluatorq()` parameters, jobs, async data sources, custom evaluators, CI exit codes | `docs/evaluation-reference.md` |
| `EvaluationResultCell` rubric / sentiment / safety scoring | `docs/structured-results.md` |
| `wrap_langchain_agent`, static vs callable instructions, input modes | `docs/framework-integrations.md` |

Config, env vars, tracing and the API reference already had pages; the README now links them.

## Quick start

The old example converted strings to uppercase — unrelated to anything evaluatorq is for. It is now two versions of a support agent scored on the same three questions, `agent-v1` at 0.33 against `agent-v2` at 1.00, with the terminal output captured from a real run. It needs no API key, and it lives in the repo at `examples/lib/basics/support_agent_eval.py`.

## Review fixes in the last commit

An adversarial review pass found four verified defects, all fixed here: `eq sim run -t` is not a real flag (only `eq redteam run` has the short form), the new reference page had the wrong `data` type, the quick start still uploads to Orq if `ORQ_API_KEY` is set, and the getting-started example exits 1 by design without saying so. `scripts/validate_mermaid.py` now also scans README.md, which just gained the repo's first root-level mermaid block.

## Known open

- 274 lines against the ticket's ~250 ceiling — the coverage and cost tables cost about 35 lines.
- The agent-simulation section has a mermaid diagram rather than a screenshot. Every frame of the existing recordings shows DEBUG log spam, `goal_achieved 0%` and an internal workspace URL, so shipping one would have been worse than shipping none. Tracked with the demo GIF work in RES-1289.

## Verification

`ruff check src`, `ruff format --check src`, `basedpyright`, `pytest -m 'not integration'` (3936 passed), `mkdocs build --strict`, `validate_mermaid.py` — all clean.


currentlycodinng (approved):
Approving. This is a large readability win and, unusually for a docs rewrite, the factual claims hold up. I checked every number against the registries rather than taking the PR body's word for it, and ran the quick start on a clean environment.

## Verified

**All five headline numbers are real.**

| Claim | Source | Result |
|---|---|---|
| 19 OWASP categories | unique keys in `OWASP_CATEGORY_NAMES` (38 with the `OWASP-` aliases) | 19 |
| 18 vulnerabilities | `list_available_vulnerabilities()` / `VULNERABILITY_DEFS` | 18 |
| 45 curated strategies | `known_strategy_names()` | 45 |
| 16 delivery methods | `list_available_delivery_methods()` | 16 |
| 18 LLM judges | distinct callables in the 36-key `OWASP_EVALUATOR_REGISTRY` | 18 |

The coverage table in the red-teaming guide also holds: its per-row counts sum to exactly 45, and its ten `curated` rows match the ten keys in `VULNERABILITY_STRATEGY_REGISTRY`, with the other nine correctly marked `generated`.

**The quick start does what the README says.** With `ORQ_API_KEY` and `OPENAI_API_KEY` both unset, `examples/lib/basics/support_agent_eval.py` gives `agent-v1` 0.33 against `agent-v2` 1.00 and exits 1, which matches the captured terminal SVG and the "this run ends with status 1" line. The README block is the file verbatim minus the `__main__` guard and the two `noqa` comments, so the "stays in sync" promise in the docstring is currently true.

**Nothing is dangling.** All 17 relative links and image sources resolve on disk. All ten docs-site paths the README links (`evaluation-reference`, `structured-results`, `llm-as-a-jury`, `cli-reference/overview`, `configuration`, `tracing`, the four guides) build to real pages, and the `guides/red-teaming/#coverage` deep link lands on a real `id=coverage`. `mkdocs build --strict` is clean, and `validate_mermaid.py` now genuinely covers README.md (34 files, clean). The only pages mkdocs reports as outside the nav are the two pre-existing `wireframe/` ones.

**The new reference signature is accurate**, including the parts that look wrong at a glance: `DataPointInput` and `EvaluatorqResult` are real aliases in `types.py` (`DataPoint | DataPointDict` and `list[DataPointResult]`), so the documented signature matches the source exactly. The three-form parameter equivalence in that page also works, I ran all three shapes live and all three pass.

**The CLI flags are right**, which is what the review-fix commit was about: `-t` is a short form on `redteam run` only, and `sim run` takes `--target` with no short form. `simulate(exit_on_failure=True)` is the real default, and `resistance_rate` / `vulnerabilities_found` are real fields on `ReportSummary`.

## Nits, none blocking

**The model id in the README will not work on the path the README recommends.** Line 150 shows `OpenAIModelTarget("openai/gpt-5.4-mini")`, but the Configuration section says `OPENAI_API_KEY` backs red teaming without Orq. `create_async_llm_client()` prefers Orq and falls back to `api.openai.com`, and `OpenAIModelTarget` passes `model` straight through, so on the no-Orq path that router-prefixed id is not a valid OpenAI model name. The red-teaming guide uses the bare `"gpt-4o-mini"` for the same construct. Worth making the two agree, and the bare form is the one that works either way.

**`eq --help` calls the dashboard a preview, the README does not.** The CLI describes it as "Launch the FastHTML dashboard (preview — still in development)" while the README section presents it plainly. One parenthetical would keep expectations aligned.

**LLM10 is missing from the OWASP LLM Top 10 column and nothing says why.** The registry genuinely stops at LLM09, so the table is accurate, but a nine-row column under a header ending in "Top 10" is the first thing a security reader will notice. A one-clause note that Unbounded Consumption is not covered yet answers it before it gets asked.

**"274 lines" in the Known open section is stale.** The head commit is 245 lines, so the ticket's ~250 ceiling is met and that caveat can come out of the description before merge.

PR #143, opened 2026-08-12
fix(redteam): stop a failed static attack from destroying the whole report

Baukebrenninkmeijer:
## What

`_run_static_target_call` put the raw `AgentResponseError` object into the static job's output dict, but `JobOutputPayload.error` is typed `str | None`. `_coerce_job_output_payload` therefore raised `ValidationError` during report generation, aborting the run.

The damage is out of proportion to the trigger: **one** non-retryable target failure on **one** static attack destroys the results of the entire run — after every attack has already executed, been judged, and been billed. Found while measuring run costs for the README: an 80-attack hybrid run took six 400s from the target and produced no report and no saved artifact.

```
Error: 1 validation error for JobOutputPayload
error
  Input should be a valid string [type=string_type, input_value=AgentResponseError(messag...r', code='target_error'), input_type=AgentResponseError]
```

The dynamic paths already flatten this correctly (`adaptive/orchestrator.py:973`, `adaptive/pipeline.py:458`); only the static path was missed.

## How

Return the same shape the dynamic paths do — a formatted message string plus `error_type`, `error_stage`, `error_code` and `error_details`, all of which `JobOutputPayload` declares. The span attributes read those fields rather than the object.

## Testing

`tests/redteam/test_static_target_job.py` covered the old object contract, so it is updated to the flattened one, and a new `test_static_job_error_payload_survives_report_conversion` runs a failed static attack's payload through `_coerce_job_output_payload` — the exact step that used to raise. Full suite: 4092 passed.


Baukebrenninkmeijer:
## Review — 4 lenses (correctness, silent failures, tests, minimality)

The core fix is verified correct. `_coerce_job_output_payload` validates unguarded at `reports/converters.py:210`, `JobOutputPayload.error` is `str | None` (`contracts.py:1321`), and there is no `try` around the call site at `converters.py:522` — so the diagnosis in the PR body holds exactly. The flattened shape is byte-identical to the single-turn dynamic path (`adaptive/pipeline.py:458-462`); the orchestrator's richer `, on turn N/M, code=…` format is the multi-turn analogue and correctly not copied.

No information is lost: `AgentResponseError` has exactly three fields and all three survive, plus `error_details` — which the static path previously dropped entirely — is now propagated. Scoring is unaffected: the gate is `output_error_text` (`common/output_adapters.py:43-58`), which already handled `str` first, so an errored static attack still short-circuits to inconclusive.

`ruff check src`, `ruff format --check src`, `basedpyright` and `pytest -m 'not integration'` (4092 passed, 2 skipped) all clean. The two `basedpyright` errors on my machine are in a gitignored `.context/` scratch file, invisible to CI.

### Fixed in c6201284

- **`error_turn` parity.** `pipeline.py:486` sets `error_turn=1` on the equivalent single-turn failure; the static path omitted it, so the dashboard's Stage/Turn column showed `-` for static errors only. Now set.
- **The success payload was never validated.** Only the failure branch went through `_coerce_job_output_payload`. `_run_static_target_call` returns an untyped `dict[str, Any]`, so a type slip on `token_usage` / `finish_reason` / `model` kills report generation in exactly the same way, after every attack has run. The success test now coerces too, which closes the whole class rather than this one instance.
- **Success path did not assert the new fields are `None`.** `converters.py:546-549` copies `error_stage` / `error_code` / `error_details` onto `RedTeamResult` without gating on `error`. If a future edit hoists `error_stage = 'target_call'` out of the conditional, every clean attack ships with an error stage and the dashboard starts counting resistant attacks as failures — with the suite fully green.
- **`error_details` had no assertion anywhere.** It is the single field the old code dropped silently, and it backs the dashboard's error-details column (`ui/dashboard.py:2294`) — the exact debugging surface this PR exists to preserve.
- Trimmed the comment at `runner.py:1242` from four lines to two. The incident narrative was already in the commit message and the new test's docstring; three copies is two too many.

### Left open — needs a call, not a rubber stamp

**1. `at_static_job` has the same bug, worse (`runner.py:2426-2434`).** The hybrid static leg returns **no `error` key at all**, despite reading `result.error` for span attributes five lines earlier. So `output_error_text` returns `None` and the judge scores the literal `[ERROR: ...]` marker text as a genuine agent response — a timed-out target in hybrid mode is silently scored RESISTANT, which is worse than the crash this PR fixes because it produces a plausible wrong number instead of no number. It also calls `target_instance.respond()` directly, bypassing `call_target_with_retry`, so it has no retry, timeout or error mapping either.

This is pre-existing and not caused by this PR, so it's a follow-up ticket rather than scope creep here — but it is the higher-severity of the two.

**2. Three hand-rolled copies of the same five-field flatten.** `runner.py:1246`, `pipeline.py:458`, `orchestrator.py:972`. The comment in this PR literally says "the way the dynamic paths do", which is duplication acknowledging itself. The clean version is one method on `TargetCallResult` in `common/target_call.py` — that struct already owns `error`, `attempts` and `error_details`, so it should own their flattened form:

```python
def error_payload(self, *, context: str = '') -> dict[str, Any]:
    """Flat error fields for JobOutputPayload/AttackOutput. Empty on success."""
    if self.error is None:
        return {}
    return {
        'error': f'Target agent failed after {self.attempts} attempt(s){context}: {self.error.message}',
        'error_type': self.error.error_type,
        'error_stage': 'target_call',
        'error_code': self.error.code or 'target_error',
        'error_turn': 1,
        'error_details': self.error_details,
    }
```

Runner becomes `{**base, **call.error_payload()}`, the orchestrator passes `context=f' on turn {turn+1}/{max_turns}'`, and both dynamic sites' defensive `result.error or AgentResponseError(...)` fallbacks delete outright. Net roughly -18 lines, and it makes the `at_static_job` fix above a one-liner. **Not done here** — it touches the orchestrator and pipeline, which is a bigger diff than a fix PR should carry, and I'd rather it land with the `at_static_job` fix it enables.

**3. Two reviewers disagreed about `_coerce_job_output_payload`, and I sided against the change.** One argued for a `field_validator(mode='before')` that stringifies a non-`str` `error`, so a producer bug costs one degraded row instead of the whole report. The other argued that makes the consumer tolerant of producer bugs, loses `attempts`/`error_code`, and would have *hidden* this defect instead of failing loudly. I left the consumer strict — but note the fragility is real and unaddressed: any future job returning a non-`str` `error` still detonates report generation for the whole run at the end. The durable fix is typing the static job's return, not loosening the validator.

**4. Minor divergences from pipeline, both arguable.**
- `token_usage`: pipeline zeroes it on the error path (`pipeline.py:441`); static returns `result.usage` unconditionally. I think static is the more honest one — the tokens were spent — which would make pipeline the side that's wrong.
- The guard is `if err` here vs `if not call.succeeded` in pipeline. They agree today (the dataclass documents `error is None iff succeeded`), but two idioms for one invariant will drift.

**5. An errored static attack can produce zero log output.** With `max_target_retries=0`, `call_target_with_retry` never reaches its retry `logger.warning` (guarded by `attempt + 1 < max_attempts`) and no non-retryable-4xx break fires. A permanently failed target is returned with no log line anywhere — the failure exists only in the final report. A `logger.warning` on the `err is not None` branch would fix it, but it applies to the dynamic path equally, so it belongs with the shared helper in (2).

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

Baukebrenninkmeijer:
## Follow-up: (1) and (2) fixed, (3) explained

Pushed as c701ce00. Full CI checks green — `ruff check src`, `ruff format --check src`, `basedpyright`, `pytest -m 'not integration'` (4105 passed, 2 skipped). The two local `basedpyright` errors remain confined to a gitignored `.context/` scratch file.

### (1) `at_static_job` — fixed

The hybrid static leg now routes through `_run_static_target_call`, the same shared call the non-hybrid leg uses. It gains retry, the per-call timeout, backend error mapping, and the six flattened `error*` fields it previously omitted entirely.

Regression test added: a raising target driven through `_run_dynamic_or_hybrid`, asserting the failure text reaches both the job dict and `report.results[0].error`. Before the fix that test raises `KeyError` on `output['error']` — the key genuinely did not exist.

Note this was the more dangerous of the two bugs. The crash this PR originally fixed is loud: no report, obvious failure, you rerun. The hybrid leg's version was silent — `output_error_text` returned `None`, so the judge scored the literal `[ERROR: ...]` marker string as a real agent reply, and a dead target was reported RESISTANT. A plausible wrong number instead of no number.

### (2) `TargetCallResult.error_payload()` — fixed, and noted in CLAUDE.md

All three copies collapse onto one method in `common/target_call.py`, next to the retry helper that produces the result. Both dynamic sites' `result.error or AgentResponseError(...)` defensive fallbacks delete outright — the method's own `is None` check replaces them. Net −73 lines against +207 (most of the addition is tests).

Two behaviour changes worth flagging rather than burying:

- **The orchestrator's message loses its `code=` fragment.** `...on turn 2/5, code=x, details=boom` becomes `...on turn 2/5: boom`. The code was always duplicated into the `error_code` field, and nothing — no test, no doc, no dashboard column — pinned the old wording. If you'd rather keep it verbatim, say so and I'll thread it through `context`.
- **`error_payload()` always returns all six keys**, `None` on success, rather than `{}`. Call sites index the dict directly (`output['error'] is not None`), so an empty dict on success would turn a clean attack into a `KeyError`. There's a unit test pinning the key set on both branches.

CLAUDE.md gets a new **Target calls and error payloads** section under Key Patterns, stating the rule both bugs violated: every target call goes through `call_target_with_retry`, no job hands an `AgentResponseError` to the report layer, and no job omits the `error` key either.

### (3) The validator disagreement, in full

The question: should `JobOutputPayload` defensively coerce a non-`str` `error` instead of raising?

**The blast radius is real and worth stating precisely.** `_coerce_job_output_payload` calls `JobOutputPayload.model_validate` unguarded (`converters.py:210`), and its call site (`converters.py:522`) sits inside the per-result loop that builds every `RedTeamResult` — with no `try` anywhere around it. So a single malformed row does not degrade that row. It raises out of the loop and kills report generation for the entire run. That is exactly how this bug produced the 80-attack run with no report and no artifact.

**Argument for loosening** (silent-failure-hunter): add a `field_validator('error', mode='before')` that stringifies a non-`str` input, pulling `.message` when present, and log a warning when it fires. Then any future producer bug — a user-supplied job, a new backend — costs one degraded row instead of the whole run.

**Argument against** (minimality): consumer leniency papers over producer bugs. The strict validator is the only reason this defect was ever found; loosen it and the static path ships a lossy payload forever, silently.

**Why I sided against, concretely.** A validator can only fix the field it validates. `error_type`, `error_stage`, `error_code`, `error_turn` and `error_details` are *sibling keys* on the payload, not attributes reachable from the `error` object — stringifying `error` recovers the message and nothing else. Five of the six fields stay lost, and `attempts` never appears at all because it lives on `TargetCallResult`, not on `AgentResponseError`. So the "degraded row" isn't degraded gracefully; it's a row that renders cleanly and is wrong, with blank error-type/code/details columns in the dashboard and nobody alerted. That is a worse failure mode than the crash, for the same reason (1) was worse than the original bug.

**But the objection doesn't answer the concern, and I want to be straight about that.** The fragility isn't the strictness of the schema — it's that a per-row failure is fatal to the run. Those are separable, and the fix that satisfies both is neither proposal: wrap the per-row coercion at `converters.py:522` in a `try` that degrades *that one row* (record the raw output and the validation error as the row's own error) and logs loudly. Strictness preserved, producer bugs still surface, blast radius capped at one attack. There's precedent ten lines up: the `AgentResponse` branch at `converters.py:213-225` already maps fields explicitly rather than letting `model_validate` fail, with a comment explaining why.

I did not do it in this PR. It's a report-layer resilience change with its own test surface, and burying it in a target-call fix would make both harder to review and to revert. Worth its own ticket — and with the producer-side helper now in place, it's the last remaining hole of this shape.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

PR #144, opened 2026-08-12
feat(pairwise): repetition-aware BT-sigma reliability from within-datapoint consistency (RES-1251)

KarinaKKarinaK:
## What

Follow-up to RES-1024 / PR #113, per the RES-1251 design ticket. The global two-item BT fit mixes judge noise with datapoint heterogeneity when a run spans different questions or response pairs, and repetitions stabilized the vote without contributing to reliability. This PR makes reliability come from the one setting where disagreement is unambiguously judge noise: repeated passes of the same prompt.

## Design (recorded on RES-1251 for sign-off)

- **Capture**: the jury already retained raw per-repetition verdicts; the pairwise layer was discarding them. `PairwiseVote.observations` now preserves them as canonicalized entries (a 'B' under the swapped ordering is recorded as 'A'; abstained/failed passes kept as `None`, plus a failure count). Nothing is silently collapsed.
- **Unit of analysis**: one (judge, comparison, ordering) group of repeated passes. Consistency is mean pairwise agreement within a group, averaged to at most ONE observation per judge per datapoint, then across datapoints. By construction: heterogeneous datapoints are never compared to each other, position bias does not read as inconsistency, and repetition count never multiplies panel weight.
- **Aggregation**: when repeats exist, `bt-sigma` winner weights come from repetition consistency (floored at 0.05); judges without repeats vote with the median weight and are named in `fit_warnings`. The pooled fit still supplies the run-level `p_a_beats_b` headline. The hierarchical per-datapoint-latents alternative is documented as the fallback model if this one proves insufficient.
- **Backward compatibility**: `observations` defaults to empty; legacy saved runs load and aggregate exactly as before (pinned by test).

## Cost model

Calls scale linearly with R (judges x orderings x R); wall-clock barely moves (concurrent fan-out); storage adds one small record per pass. Recommended R=2: the minimum that yields consistency evidence. R=3 only refines the per-group agreement scale at +50% cost.

## Docs and tests

`docs/pairwise-judging.md` gains a section stating what consistency estimates and what it must NOT be read as (not task difficulty, not overall judge quality, not accuracy). Sixteen new tests: capture through the real `run_pairwise` pipeline, swapped-order canonicalization, failed passes, the heterogeneity guard, weight normalization, neutral fallback, legacy-run compatibility.

Full suite: 4,145 passed, basedpyright clean, strict docs build.

Ticket: RES-1251.


KarinaKKarinaK:
Deep self-review against RES-1251 plus a test pass, results below. One real gap found and fixed in 0abbeef1.

## Review against the ticket, item by item

1. Per-repetition votes preserved and canonicalized on PairwiseVote: yes, including abstained and failed passes as None entries plus a failure count, and order flips stay visible because the two orderings are separate observation groups.
2. Backward compatibility: the reconciled vote field and the default plurality path are untouched; a legacy serialized vote deserializes with empty observations and legacy runs aggregate exactly as before (both pinned by tests).
3. Unit of analysis: one (judge, comparison, ordering) group of repeated passes of the same prompt. Heterogeneous datapoints are never compared to each other by construction, pinned by test.
4. Grouped alternative implemented and documented (within-datapoint consistency); the hierarchical model is documented as fallback with the identifiability rationale on the ticket.
5. Normalization: at most one observation per judge per datapoint, one weighted vote per judge per comparison; a 5-repetition judge cannot outvote two 2-repetition judges (pinned by test).
6. Feasibility: the cost model (calls = judges x orderings x R) is now verified structurally by a call-count test at R = 1, 2, 3; recommendation R=2 recorded on the ticket.
7. All six required test areas covered, see below.
8. Docs state what consistency estimates and what it must not be read as.

## The gap the review found

A non-converged pooled fit returned uniform plurality BEFORE the repetition path ran, discarding consistency evidence that does not depend on the fit at all. Fixed: consistency is computed first; on a non-converged fit with repeats the winners stay consistency-weighted and only the run-level headline degrades to neutral (p_a_beats_b 0.5, converged false, explicit warning). Sigmas are only reported from a converged fit. Regression test added.

Known boundary, unchanged from before this PR: swap=False runs still fall back to uniform plurality (position-bias symmetrisation unavailable), so repetition weighting currently requires both orderings. Can be lifted later if single-ordering runs need it.

## Test results

- 14 tests in tests/unit/test_pairwise_repetitions.py: capture through the real run_pairwise pipeline, swapped-order canonicalization, failed/abstained passes, per-ordering grouping (position bias does not read as inconsistency), the heterogeneity guard, no-evidence-at-R=1, consistency-weighted winners, weight normalization, neutral fallback with warning, legacy aggregation and deserialization, non-convergence regression, JSON round-trip, end-to-end pipeline at repetitions=3, and cost-model call scaling. All pass.
- Existing BT-sigma and pairwise suites: 135 passed, unchanged behaviour on runs without repeats.
- Full suite: 4,149 passed, 2 skipped, basedpyright 0 errors, ruff clean, strict docs build green. One transient fixture error appeared in a single full-suite run (tests/dashboard/test_downloads.py), passes in isolation, in its own suite, and on the full rerun; unrelated to this change.

KarinaKKarinaK:
Follow-up fixes pushed (0ee19db):

- The one-sided guard warnings now match the actual weighting on the repetition path. When consistency weights are active they say the fit sigma is excluded and the weight comes from repetition consistency, instead of falsely claiming neutral weighting.
- The datapoints-with-repeats count now uses the same per-ordering rule as the estimator: at least two decisive passes within one ordering group. Split passes like ab=[A, None], ba=[A, None] no longer count.
- RepetitionObservation and repetition_consistency are exported from the package root and now show up in the generated API reference.
- Docs note added: abstained or failed passes are excluded from consistency, so a judge that abstains often is not penalized.

Two new tests pin the warning text and the count rule. ruff, basedpyright, pytest, and strict docs build all pass.

src/evaluatorq/pairwise.py
```diff
@@ -219,6 +251,42 @@ def _uniform_plurality_aggregation(
...
+
+    Empty dict when no judge has any qualifying group (e.g. repetitions=1 or a
+    legacy run without observations).
+    """
+    per_judge: dict[str, list[float]] = {}
+    for c in comparisons:
+        for v in c.votes:
+            groups: dict[str, list[str]] = {}
+            for o in v.observations:
+                if o.verdict is not None:
```
Baukebrenninkmeijer, line 349: This filter removes the abstained and failed passes, and nothing later puts them back.

Take a group with observations `['A', None, 'A']`. `groups` keeps `['A', 'A']`, `pairs` is `[True]`, and the group score is `1.0`. A judge that answered all three passes and agreed on all three also scores `1.0`. So a judge that fails one pass in three is scored as fully self-consistent.

`PairwiseVote.repetition_failures` (line 113) holds the count, but neither `repetition_consistency` nor `bt_sigma_aggregation` reads it. A judge's failure rate has no effect on its weight.

Use the failure rate in the group score, or in the floor applied to it.

src/evaluatorq/pairwise.py
```diff
@@ -273,33 +341,80 @@ def bt_sigma_aggregation(comparisons: Sequence[PairwiseComparison]) -> BTSigmaAg
...
 
+    # RES-1251: when per-repetition observations exist, reliability weights come
+    # from within-datapoint consistency instead of the global two-item fit, so
+    # datapoint heterogeneity structurally cannot masquerade as judge noise.
+    # The pooled fit still supplies the run-level headline (p_a_beats_b) - two
+    # different questions, two different estimators.
+    if rep_consistency:
+        weights = {judge: max(c, _MIN_CONSISTENCY_WEIGHT) for judge, c in rep_consistency.items()}
+        voted = {v.model for c in comparisons for v in c.votes if v.vote is not None}
+        neutral = statistics.median(weights.values()) if weights else 1.0
```
Baukebrenninkmeijer: `neutral` is the median of the judges that have evidence, and the warning on the next line calls the result "weighted neutrally".

Take a panel of five judges where only one produces a qualifying group, and that group agrees. `weights` is `{J1: 1.0}`, so `neutral` is `1.0`, and the four judges the run never measured also get `1.0` — the maximum weight the scale holds. The reverse case is the same shape: one measured judge near the `0.05` floor pushes four unmeasured judges down to that value.

The borrowed value can also be lower than a judge's own measured value, so a judge with evidence can weigh less than a judge with none.

Use a fixed weight for judges with no repetition evidence, or require a minimum number of measured judges before the median is used.

src/evaluatorq/pairwise.py
```diff
@@ -543,6 +659,33 @@ def _failed(model: str) -> bool:
...
+            for ordering, jv, swapped in (
+                ('ab', first_votes.get(model), False),
+                ('ba', second_votes.get(model), True),
+            ):
+                if jv is None:
+                    continue
+                failures += jv.repetitions_failed
+                for i, raw in enumerate(jv.repetitions):
+                    value = raw if not swapped else _unswap(raw)
+                    raw_verdict = str(value) if value is not None and str(value) in _PAIRWISE_VALUES else None
```
Baukebrenninkmeijer: A decisive pass whose value is outside `_PAIRWISE_VALUES` becomes `verdict=None` here.

`PairwiseJudgeFn`'s `Prediction.value` is `bool | float | str`, which is wider than `{'A', 'B', 'tie'}`. If one repetition returns `True` while the others return `'A'`, the plurality value stays `'A'`, so `_decisive_value` does not raise. This line then records that pass as `None`. `repetition_failures` does not count it either, because the pass was decisive.

The result is that a contract violation and a clean abstention are the same value. The `RepetitionObservation` docstring says nothing is silently collapsed.

Raise, or count the pass in `repetition_failures`.

src/evaluatorq/pairwise.py
```diff
@@ -177,6 +202,13 @@ class BTSigmaAggregation(BaseModel):
         default_factory=dict,
         description='Per-judge discriminator; smaller = sharper/more reliable. Empty on single-judge fallback.',
     )
+    repetition_consistency: dict[str, float] = Field(
```
Baukebrenninkmeijer, line 231: `repetition_consistency` is the field that decides `winners` when it is not empty, and it stays inside `BTSigmaAggregation`.

`build_report` sets `stats.sigma = bt_block.judge_sigmas.get(stats.model)` (line 497), and the judges table in `pairwise_reports/sections.py` renders that `sigma`. When repetition weighting is active, `judge_sigmas` still holds the pooled two-item fit values, which did not decide this run's winners. The per-judge consistency values reach the reader only as a count inside `fit_warnings`.

Add the per-judge consistency next to `sigma` in `JudgeStats`, or mark `sigma` as inactive when repetition weighting is in effect.

Baukebrenninkmeijer (commented):

**3.5/5** — the estimator is well built and well tested; two inputs it treats as evidence are not evidence, and one design question is for @Baukebrenninkmeijer.

All 13 checks pass. I read `repetition_consistency`, `bt_sigma_aggregation` and `_observations` at the head sha. The group rule is correct: one comparison contributes at most one observation per judge, and the per-ordering grouping does keep position bias out of the consistency number. Legacy runs with an empty `observations` fall back to the pooled fit, and `test_legacy_run_without_observations_aggregates_unchanged` pins it.

### P1

1. **A failed or abstained repetition does not lower consistency.** (`src/evaluatorq/pairwise.py:273`) — `['A', None, 'A']` scores `1.0`, the same as `['A', 'A', 'A']`. `repetition_failures` is stored but no code reads it.
2. **The fallback weight is the median of the measured judges, but the warning calls it neutral.** (`src/evaluatorq/pairwise.py:398`) — with one measured judge at `1.0`, every unmeasured judge also gets `1.0`, the maximum weight.

### P2

- **A decisive repetition with an out-of-contract value becomes `None`.** (`src/evaluatorq/pairwise.py:680`) — it is not counted in `repetition_failures`, and the docstring says nothing is silently collapsed.
- **`repetition_consistency` is stored but the judges table still shows the pooled `sigma`.** (`src/evaluatorq/pairwise.py:205`) — the reader sees a reliability number that did not decide the winners.

### For Bauke

- **Does this estimator need shrinkage?** With `R=2` a group's consistency can only be `0.0` or `1.0`, and the mean over groups is unweighted. A judge with one lucky `R=2` group carries the same weight as a judge with many groups. RES-1251 records the design for your sign-off, and the recommended setting is `R=2`, so this is the common case rather than an edge case.

KarinaKKarinaK:
Pushed ecddacef addressing the review, and merged main in (branch is level with base).

P1:
1. Failed and off-contract repetitions now lower reliability. repetition_consistency discounts a judge's agreement by the share of its passes that errored or came back off-contract, so repetition_failures is finally load-bearing. A genuine abstention still does not penalize (declining honestly should not cost reliability), and that split is now documented. Test: failure lowers consistency (2/3) while a clean abstention stays 1.0.
2. The no-repeats fallback warning no longer calls itself neutral. It names the actual weight, the median measured consistency, with the value. Updated the test that pinned the old wording.

P2:
- An out-of-contract decisive value is now counted as a failure instead of silently collapsing to None. Test pins it through run_pairwise (reps [A, banana, A] -> observations [A, None, A], repetition_failures 1).
- The judges table surfaces per-judge repetition consistency next to sigma, so the reader sees the reliability that actually decided the winners. Added JudgeStats.consistency, the report wiring, and the HTML column.

For Bauke (shrinkage): I did not change the estimator here, since it needs your sign-off. I documented the trade-off in repetition_consistency: at R=2 the per-judge mean over groups is unweighted, so a judge with one lucky R=2 group counts like a judge with many. Shrinkage toward the panel mean by group count would damp it. Happy to add count-weighted shrinkage in a follow-up if you want it, or keep it simple and note the caveat. Your call.

Full suite 4210 passed, ruff and basedpyright clean.

Baukebrenninkmeijer:
Second review pass on `ecddacef`, six independent critics (different lenses: correctness, tests, silent failures, reuse/convention, strategy, goal alignment), each told to skip the four findings already raised and fixed. Everything below is new or an incomplete fix, and I verified each one in the source myself.

Three of the six lenses independently landed on the same two items, and both reopen the earlier thread rather than being new ground.

## Decisions — your call, I'm not deciding these for you

**1. `pairwise.py:427-435` — the median fallback is unchanged; only its warning text changed.**

The earlier comment was that judges with no repetition evidence borrow the median of the measured judges, so a single measured judge sets everyone's weight. `ecddacef` made the warning honest about that ("median measured consistency (1.00)") but kept the behaviour. A panel with one measured judge at consistency 1.0 still hands four never-measured judges the maximum weight the scale holds; the code comment at :430-432 says so explicitly. The reverse still holds too: one measured judge near the 0.05 floor drags four unmeasured judges to the floor.

Naming the number is better than calling it neutral, but a reader who sees "median measured consistency (1.00)" on a five-judge panel still has no signal that the median came from n=1. Options: a fixed prior for unmeasured judges, or a minimum number of measured judges before the median is used at all. If you deliberately want the median, gate it on count and say so in the docstring.

**2. `pairwise.py:340-344` — `swap=False` bypasses repetition weighting entirely, and the docs don't say so.**

The single-ordering early return fires before the repetition path at :419 ever runs. With `swap=False, repetitions=3`, a steady judge returning `A,A,A` and a noisy one returning `B,A,B` carry perfectly usable consistency evidence — same prompt, repeated passes — and the run returns uniform plurality with `repetition_consistency` empty.

You called this a known boundary in the thread above, but `docs/pairwise-judging.md:206-208` states the behaviour unconditionally: "Run comparisons with `repetitions=2` (or more) and the reliability weights stop coming from the global two-item fit". Nothing there mentions `swap`. Consistency needs no position-bias symmetrisation — it is computed per ordering group precisely so orderings never mix — so the estimator could run here. Either lift the boundary or document it; leaving the docs unconditional is the one option that isn't defensible.

**3. Shrinkage at R=2** — the question you deferred. Two lenses reached it independently. At R=2 a group score is only 0 or 1, so a judge with one lucky `[A,A]` group scores 1.0 while a judge with ten groups and one disagreement scores 0.91, and the sparse judge wins the row. Not a blocker: the ticket's simple across-datapoint mean is implemented and the caveat is documented. It is a real follow-up risk at the R you recommend.

## Fixes worth making

**4. `docs/pairwise-judging.md:239-241` is now false.** It says "Abstained or failed passes are excluded from consistency ... `['A', None, 'A']` scores 1.0". After `ecddacef` that shape scores **2/3** when `repetition_failures=1` (`pairwise.py:301-310`). Only a clean abstention still scores 1.0. The paragraph needs to distinguish the two cases the code now distinguishes — which is the whole point of the change.

**5. `docs/pairwise-judging.md:180-182` is stale for the same reason.** It says the fit "falls back to uniform plurality when the optimizer does not converge". Since `0abbeef1` that is only true without repeats: with repeats, the winners stay consistency-weighted and only the pooled `p_a_beats_b` headline degrades to neutral.

**6. `common/jury.py:394` — one hole left in "failures lower reliability".** `failed_count` counts only `p.error is not None`. A `Prediction(value=None, abstained=False)` is neither decisive nor an abstention nor an error: it lands in `repetitions_raw` as `None`, `repetition_failures` stays 0, and `pairwise.py:717` records it as a genuine abstention. So a mechanically unusable pass keeps consistency at 1.0 — the exact case #4 was supposed to close. Counting `not p.decisive and not p.abstained` as failed closes it. Shared jury layer, so it touches redteam and simulation too; worth checking those before changing it.

**7. `pairwise.py:721-726` degrades without announcing itself.** An off-contract decisive value collapses to `None` and increments the failure count, but no `logger.warning` names the judge, the repetition, or the offending value — so the user sees a lowered consistency with no way to find out why. Related: the `repetition_failures` field description at :115-118 still says "passes that raised an error", which is no longer what it counts.

**8. `pairwise.py:376-380` — a third silent fallback.** When observations exist but no ordering group holds two decisive passes (`ab=[A, None]`, `ba=[A, None]` at R=2), `repetition_consistency` comes back empty and the run uses pooled-fit weights with no warning at all. The user asked for repetition weighting, paid 2x the calls, and got the old estimator with nothing in `fit_warnings` to say so. The other two fallbacks on this path warn; this one should match them.

**9. `pairwise.py:189-192` — `BTSigmaAggregation`'s docstring states the wrong contract.** It still says winners are re-derived "as a reliability-weighted vote (weight `1/sigma`)". On the repetition path the weight is `max(consistency, 0.05)` (:425). This is the public model docstring, so it is what a user inspecting the API reads.

## Tests

**10. Two of the new tests pass on a broken implementation.**

- `test_judge_without_repeats_gets_fallback_weight_and_an_honest_warning` (:199-208) asserts only the warning string. An implementation that emits the warning and then applies uniform weight passes. Assert the weighting: with that fixture, uniform plurality gives A on the second row and the median fallback gives B.
- `test_failed_pass_lowers_consistency_but_a_clean_abstention_does_not` (:211-219) calls `repetition_consistency` directly, so it would still pass if `bt_sigma_aggregation` ignored the penalty when assigning winner weights. Add a row where a clean judge votes A and a flaky judge with `[B, B, None], repetition_failures=1` votes B, and assert the winner.

**11. Coverage gaps that let a regression through silently.**

- No test drives the swapped ordering's failure path (`pairwise.py:707-713`); the capture fixture uses `swap=False`, so dropping `repetitions_failed` for `ba` votes breaks nothing.
- No `run_pairwise(repetitions=1)` assertion through the real pipeline — the cost test checks call count only, so storing no observation at R=1 passes.
- The new `Consistency` column (`export_html.py:133-146`) is tested at the `JudgeStats` field level only; removing the header/value mapping leaves the suite green.

---

Nothing here is a correctness bug in the estimator itself — the maths does what the ticket asked. The pattern across the whole batch is narrower and worth naming: **`ecddacef` changed behaviour in three places and the prose describing that behaviour moved in none of them.** Items 4, 5, 7 and 9 are all the same failure, and it is the failure most likely to reach a user, because the docs are what they will trust when the number surprises them.

Baukebrenninkmeijer:
Re-checked my review above against `b69f60ab`, which landed alongside it. Two of the three decisions are answered; the fix list is untouched, and the shrinkage change adds one new instance of the same docs-drift pattern.

## Answered by `b69f60ab`

**Decision 3 (shrinkage) — done.** Empirical-Bayes toward the panel mean, `k=1` pseudo-observation, with a test that pins the direction (`many > few`, `few < 1.0`). This is the right shape for the R=2 problem.

**Decision 1 (median fallback) — mostly.** `_MIN_MEASURED_JUDGES = 2` means a single measured judge can no longer set the whole panel's weight, and the skip is now announced in `fit_warnings`. Residual: at exactly two measured judges the median is just those two averaged, and it is still borrowed by arbitrarily many unmeasured judges. Fine as a deliberate choice — worth one line in the docstring saying quorum 2 is the floor, not a claim of sufficiency.

**Item 8 — half.** The new warning covers `rep_consistency` non-empty but below quorum. The case I filed is the one where it comes back **empty** despite observations existing (`ab=[A, None]`, `ba=[A, None]` at R=2): `if rep_consistency and not rep_active` short-circuits on the empty dict, so that run still silently uses pooled-fit weights after paying 2x the calls.

## New in `b69f60ab`

**12. Shrinkage changed what the published number means, and nothing that describes it moved.** `docs/pairwise-judging.md:229` still says "Read `bt_sigma.repetition_consistency` for the per-judge values (1.0 = always agrees with itself)". After shrinkage a perfectly self-consistent judge cannot read 1.0 unless the whole panel is at 1.0 — your own tests had to relax `== 1.0` to `> 0.9` in five places to accommodate that. The same value is what `JudgeStats.consistency` publishes and what the new HTML column renders under the header `Consistency`, so a reader comparing two runs with different panels sees a judge's number move without that judge changing.

Either rename it to what it now is (a shrunk reliability estimate, not raw self-agreement) or publish the raw mean alongside. The docs also mention neither shrinkage nor the quorum at all, so the section now describes the pre-`b69f60ab` estimator.

**13. `pairwise.py:318` — the shrinkage anchor is the unweighted mean of the raw per-judge means.** A thin-evidence judge with one lucky `[A, A]` group contributes to the panel mean at full strength, and is then shrunk toward an anchor it helped set — the noise being damped is part of the target. On a two-judge panel with one comparison each, both judges move halfway toward each other, compressing exactly the separation the weights are derived from. An evidence-count-weighted panel mean (`sum(n*mean)/sum(n)`) removes the self-reference for one line of code.

**14. `k = _SHRINKAGE_PSEUDO_OBS = 1.0` is undocumented as a choice.** The docstring says "signed off in review" — what was signed off is shrinkage, not this magnitude. At `n=1, k=1` a judge is pulled halfway to the panel mean, which is a strong prior. One line on why 1.0 and what moving it does.

## Still open, unchanged

- **Decision 2** — `swap=False` bypasses repetition weighting while `docs/pairwise-judging.md:206-208` states the behaviour unconditionally. `b69f60ab` adds a second undocumented exception (the quorum) to the same paragraph.
- **Items 4, 5** — the two stale docs paragraphs, now three with #12.
- **Item 6** — `common/jury.py:394` still counts only `p.error`, so a non-abstained valueless pass keeps consistency at 1.0.
- **Item 7** — off-contract collapse still logs nothing.
- **Item 9** — `BTSigmaAggregation` docstring still promises weight `1/sigma`.
- **Item 10** — `test_judge_without_repeats_gets_fallback_weight_and_an_honest_warning` still asserts only the warning string; an implementation that warns and then weights uniformly passes. `test_failed_pass_lowers_consistency_but_a_clean_abstention_does_not` got weaker, not stronger: the exact `2/3` became `careful > flaky`, and it still never goes through `bt_sigma_aggregation`, so the penalty could be ignored in the weighting and the test would pass. Shrinkage forced the loosening, which is a reason to assert the weighting outcome rather than the value.
- **Item 11** — no swapped-ordering failure test, no `repetitions=1` capture assertion through the real pipeline, no render test for the new column.

`tests/unit/test_pairwise_repetitions.py`: 21 passed on `b69f60ab`.

The estimator is in good shape now. What is left is almost entirely that the code has moved three times and the prose describing it has moved once.

KarinaKKarinaK:
Thanks for the two detailed passes. All open items addressed in `4accc510`. Point by point against your latest list.

**Decisions**
1. **Median fallback / quorum** — added the one-line docstring note that `_MIN_MEASURED_JUDGES = 2` is a *floor*, not a sufficiency claim (at exactly two the median is those two averaged and still borrowed by any number of unmeasured judges).
2. **`swap=False`** — documented, not lifted. `docs/pairwise-judging.md` now states the repetition path needs both orderings (`swap=True`) *and* a quorum, and falls back to the pooled fit (with a `fit_warnings` line) when either is unmet. No longer an unconditional claim.
3. **Shrinkage** — already landed in `b69f60ab`; the residual doc/anchor points are items 12/13 below.

**Fixes**
- **4 (docs `['A', None, 'A']` = 2/3)** — rewritten: a clean abstention scores 1.0, but an errored/off-contract `None` counted in `repetition_failures` discounts by the failed share, so the same shape scores 2/3. The doc now names what separates the two identical-looking `None`s.
- **5 (docs convergence fallback)** — fixed: uniform-plurality fallback applies *only without repetition weights*; with repeats the winners stay consistency-weighted and only `p_a_beats_b` degrades to neutral.
- **6 (`jury.py` failed_count)** — fixed in the shared layer: counts `not p.decisive and not p.abstained`, so a `value=None, abstained=False` pass is a failure, not a free abstention. `decisive` already excludes ties and clean abstentions, so nothing legitimate is miscounted. Verified redteam + simulation suites stay green (2761 passed; the only failures are missing optional extras `huggingface_hub`/`fasthtml`, unrelated).
- **7 (off-contract collapse)** — now logs a `logger.warning` naming judge, ordering, repetition and the offending value; `repetition_failures` field description updated to "errored OR off-contract".
- **8 (empty `rep_consistency` despite observations)** — added the missing warning for the `ab=[A,None], ba=[A,None]` case (observations exist, no ordering group has 2 decisive, dict empty). It no longer falls back silently.
- **9 (`BTSigmaAggregation` docstring)** — fixed to say the repetition-path weight is `max(shrunk_consistency, 0.05)`, not `1/sigma`.
- **12 (shrinkage renamed the published number)** — went with **publish raw alongside**: new `repetition_consistency_raw` (un-shrunk self-agreement) on the model, `JudgeStats.consistency_raw`, and a split HTML column (`Consistency (shrunk)` / `Consistency (raw)`). Docs updated: the shrunk value is a reliability weight (below 1.0 unless the panel is), the raw value is self-agreement, compare within a run.
- **13 (self-referential anchor)** — the shrinkage anchor is now the evidence-count-weighted panel mean `sum(n*mean)/sum(n)`, so a thin-evidence judge cannot help set the anchor it is shrunk toward.
- **14 (`k = 1.0` undocumented)** — added the rationale: at n=1 group, k=1 pulls a judge halfway to the panel mean (a deliberately strong prior for the 0/1-per-group R=2 regime); raising k shrinks harder, lowering it lets a lucky group count for more.

**Tests**
- **10** — `test_judge_without_repeats_...` now asserts the winner (row 1 is B under the median fallback, A under uniform, so it pins the weighting, not just the warning string). Added `test_failed_pass_penalty_decides_the_winner_through_aggregation` which drives the penalty through `bt_sigma_aggregation` and asserts careful beats flaky.
- **11** — added: swapped-ordering failure count, a `repetitions=1` assertion through the real pipeline (no consistency), and an HTML render test asserting both new columns and distinct shrunk/raw values.
- Plus new coverage for items 6, 7, 8, 12 and 13.

`tests/unit/test_pairwise_repetitions.py`: 30 passed. `ruff`, `ruff format` and `basedpyright` clean on the changed source.

currentlycodinng (approved):
Approving. The estimator does what the description says, and I checked it by running it rather than reading it, including a live panel. The one thing I would fix before or just after merge is a comment that misstates what the shrinkage constant is doing, since it is the constant a reader would reach for first when tuning this.

## Verified live

I ran a four-comparison panel of `gpt-5.4-mini`, `claude-haiku-4-5` and `gpt-4o-mini` at `repetitions=2` against the real API. Capture is exact: 48 observations for 4 comparisons x 3 judges x 2 orderings x 2 passes, no drops, no double counts. All three judges came back at raw 1.0 on these cases, the weights stayed at 1.0 because the whole panel was self-consistent (which is the caveat in the docstring behaving as documented), and the winners matched the intended better answer on all four.

To see the estimator separate judges rather than agree with everyone, I built a panel through the real `run_pairwise` with one judge that alternates its verdict every call and two that decide from content. Raw self-agreement came out `{X: 1.0, Y: 0.0, Z: 1.0}` and the weights `{X: 0.952, Y: 0.095, Z: 0.952}`, so the inconsistent judge is down-weighted by a factor of ten while still casting a vote. The position-bias property holds by construction as claimed: a judge that always picks the first slot produces `ab=[A,A]` and `ba=[B,B]` after un-swapping, each group scores 1.0, and it reads as consistent rather than noisy.

I also checked the `swap=False` claim in the docs, because `_repetition_stats` groups per ordering and a single `ab` group of two decisive passes qualifies on its own, so I expected the doc to be overstating it. It is not: `bt_sigma_aggregation` bails to uniform plurality at the `v.completed` guard before any of this runs, and end to end a `swap=False` run at `repetitions=2` returns an empty `repetition_consistency` with the single-ordering warning. The doc is right and my reading of the estimator in isolation was the wrong level to check it at.

The failure discount matches the worked example in the docs exactly: `['A', None, 'A']` scores 1.0 when the middle pass is a clean abstention and 2/3 when it is a failure. Tightening `failed_count` in `jury.py` from `p.error is not None` to `not p.decisive and not p.abstained` is the right call and closes the case where a mechanically unusable pass counted as a free abstention. The HTML table renders `-` for both new columns on runs that have no consistency, and there is a test pinning the headers.

The full suite is 4,567 passed with one failure, `test_pairwise_panel_picks_the_better_answer`, which is only the missing-credentials guard. It passes with `ORQ_API_KEY` set. The description says 4,145, so that number is stale rather than wrong.

## The shrinkage comment describes a different `n` than the code uses

`_SHRINKAGE_PSEUDO_OBS` is documented as "k = 1.0 means one group of evidence and the prior carry equal weight, so at the R=2 the ticket recommends (n = 1 group) a judge is pulled halfway to the panel mean". But `n` in `(n * mean + k * panel_mean) / (n + k)` is not the group count and has nothing to do with R. `_repetition_stats` appends one entry per comparison, so `n` is the number of datapoints on which that judge had a qualifying group. The prior is therefore only worth half the weight on a one-datapoint run, and on any realistic run it is close to inert:

```
Judge X always agrees with itself, judge Y never does, R=2 throughout.

  datapoints=1    raw {X: 1.0, Y: 0.0}   shrunk {X: 0.750, Y: 0.250}   weight ratio  3.0
  datapoints=2                           shrunk {X: 0.833, Y: 0.167}   weight ratio  5.0
  datapoints=5                           shrunk {X: 0.917, Y: 0.083}   weight ratio 11.1
  datapoints=20                          shrunk {X: 0.976, Y: 0.024}   weight ratio 40.7
  datapoints=50                          shrunk {X: 0.990, Y: 0.010}   weight ratio 99.0
```

The mechanism still serves its stated purpose, since a judge measured on one datapoint out of fifty is genuinely pulled toward the panel mean relative to one measured on all fifty, and that is the "one lucky R=2 group cannot dominate the run" case. What is not true is the numeric story attached to it, and the field descriptions inherit it: "a self-consistent judge reads below 1.0 unless the whole panel is" is technically right and practically 0.99 by fifty datapoints, which is a smaller caveat than the wording implies. Rewriting the comment in terms of datapoints-with-evidence, and saying explicitly that k trades off between judges within a run rather than damping R, would let whoever tunes k next reason about it correctly.

## Smaller

`repetition_consistency` is exported from `evaluatorq/__init__.py` and added to `__all__`, but `repetition_consistency_raw` is not, even though the docs send readers to the raw number as the one to compare within a run. Both are public functions on the module; exporting only the shrunk one makes the value the docs recommend the harder of the two to reach.

The off-contract branch logs one warning per bad repetition per ordering. On a large run with a judge that is reliably off-contract, that is judges x orderings x R lines saying the same thing. Not wrong, and it is the behaviour the earlier review asked for, but a per-vote summary would read better at scale.

Baukebrenninkmeijer:
Third pass, on `4accc510` (+ the `main` merge). Everything I filed is addressed — the docs now describe the estimator that exists, both silent fallbacks announce themselves, `repetition_consistency_raw` publishes the un-shrunk signal, the shrinkage anchor is evidence-weighted, `k=1.0` is justified in-place, and the two weak tests now assert winners. Locally: ruff clean, `ruff format --check src` clean, basedpyright 0 errors, 4583 passed / 2 skipped.

Decision 2 answered by documenting the boundary rather than lifting it, which was one of the two options — `docs/pairwise-judging.md:211-221` now lists both gates up front. Good place to land it.

Seven new items, all introduced by this commit. One is on the default path.

## P1

**15. Every `repetitions=1` run now emits a false "repeated passes exist" warning.** `pairwise.py:458-462`

The item-8 guard fires whenever `any(v.observations)` and `rep_consistency` is empty. Observations are captured at R=1 too (one per ordering), so the default path trips it. Verified on a 9-comparison, 2-judge, R=1 run:

```
repetition weighting skipped: repeated passes exist but no single ordering held >= 2 decisive passes to measure consistency; using the pooled two-item fit
```

There were no repeated passes and nothing was skipped. Every plain bt-sigma user now sees a warning about a feature they did not ask for, which is the fastest way to teach people to ignore `fit_warnings`. Gate on evidence of an actual repeat — some ordering group holding ≥2 observations — not on observations existing. `test_single_repetition_through_pipeline_has_no_consistency` asserts the empty dicts but not the absence of this warning, which is why it slipped through.

**16. `repetition_consistency_raw` is not exported.** `__init__.py`

```python
>>> import evaluatorq
>>> hasattr(evaluatorq, 'repetition_consistency'), hasattr(evaluatorq, 'repetition_consistency_raw')
(True, False)
```

`from evaluatorq import repetition_consistency_raw` raises `ImportError` while its sibling works, and the API reference is generated from top-level `__all__`, so the new public helper does not appear there. `docs/pairwise-judging.md:244-250` points readers at the raw metric. The new tests import from `evaluatorq.pairwise`, so they don't catch it — the same shape as the `RepetitionObservation` export you fixed in `0ee19db`.

## P2

**17. The "raw" value is not raw.** `pairwise.py:347-355`, `:368`

`_repetition_stats` applies the failure discount (`agreement * completion`) before returning, and `repetition_consistency_raw` publishes that. So `['A', <failed>, 'A']` reads **0.667** under a field described as "RAW within-datapoint self-agreement ... 1.0 = this judge always agreed with itself on repeated passes" — and it did always agree with itself, on every pass it completed. The field contradicts its own description and the docs sentence at :244-246.

Pick one: keep pure agreement in `*_raw` and apply the discount only where the weight is formed, or rename it (`consistency_unshrunk`) and say it is failure-adjusted. This matters more than it looks, because the whole point of item 12 was that the published number should mean what it says.

**18. The widened failure semantics stop at the jury's internals.** `contracts.py:783`, `docs/tracing.md:353`

`common/jury.py:394-400` now counts any non-decisive, non-abstained pass, and the reasoning is written out in the code. But `JuryVote.repetitions_failed` still declares "Count of repetitions that raised an error (non-decisive due to error)", and the `judge.repetitions_failed` span attribute is documented the same way. That is the public schema for redteam and simulation, which now emit the wider count. The behaviour is right; the contract now lies about it.

**19. Both early returns still drop the new diagnostic.** `pairwise.py:434-444`, `_uniform_plurality_aggregation` at `:266-283`

`repetition_consistency_raw` is now a *diagnostic*, not a weight, so there is no longer a reason to withhold it on paths where weighting is off. A `swap=False, repetitions=3` run, and a run whose judges are position-biased into no reconciled vote (`ab=[A,A]`, `ba=[B,B]`), both return `repetition_consistency_raw={}` while `repetition_consistency_raw(comparisons)` computes fine. The report renders `-` for evidence the run collected and paid for. Populating it on the fallback paths is the point of having split raw from shrunk.

## P3

**20. The one-sided-judge rule is still stated unconditionally.** `docs/pairwise-judging.md:194-199`

It says such a judge "votes with a neutral weight instead". On the repetition path every weight is replaced by consistency — your own comment at `pairwise.py:519-521` says the neutral assignment "never survives" there. Same qualification the other two rules just got.

**21. `error` and `abstained` together still slips the failure count.** `common/jury.py:400`

`Prediction(value=None, abstained=True, error='boom')` is constructible and yields `decisive=False, abstained=True`, so the new predicate does not count it. Verified. Nothing in the repo builds that combination today, so it is latent, not live — `not p.decisive and (not p.abstained or p.error is not None)` closes it.

**22. Two test gaps in the new batch.**

- `test_html_judges_table_shows_shrunk_and_raw_consistency` asserts both headers and `'1.00' in html`. Swapping the two values at `export_html.py:149` leaves all three assertions true, so the column mapping is unpinned — assert the cell at each header's index.
- No test drives `swap=False, repetitions=2`. It is now a *documented* gate, and removing the guard at `pairwise.py:415-419` would leave the suite green.

---

Fifteen of my items closed in one commit with the reasoning written into the code, which is the right place for it. The pattern in what is left is narrower than last round and worth naming: **item 8's fix and item 12's fix each shipped a claim that is slightly wider than the code** — a warning that says "repeated passes exist" when one pass exists, and a field that says "raw" when it is failure-adjusted. Same class of drift as before, just now inside the fixes rather than the docs.

KarinaKKarinaK:
Thanks — third pass items all addressed in `e6e328f1`. The pattern you named (a fix shipping a claim slightly wider than the code) is exactly right; each one is now narrowed to what the code does.

**P1**
- **15 (false R=1 warning).** The item-8 guard was gated on `any(v.observations)`, but R=1 captures one pass per ordering, so the default path tripped it. Now gated on an actual repeat — some ordering group with ≥2 passes. Added an assertion to `test_single_repetition_through_pipeline_has_no_consistency` that the warning is absent, which is what let it slip.
- **16 (not exported).** `repetition_consistency_raw` is now in `evaluatorq.__init__` imports and `__all__`, so `from evaluatorq import repetition_consistency_raw` works and it lands in the generated API reference. New test imports it from the package root.

**P2**
- **17 (the "raw" value was not raw).** Split it: `_repetition_stats` now returns `(raw_agreement, discounted, n)`. `repetition_consistency_raw` publishes pure self-agreement — `['A', <failed>, 'A']` reads **1.0** (it did agree with itself on every completed pass) — and the failure discount lives only in the shrunk `repetition_consistency` weight (**2/3**). Field descriptions and the docs worked example updated to distinguish the two. Test pins raw 1.0 while the weight drops.
- **18 (contract lagged the behaviour).** `JuryVote.repetitions_failed` field description and the `judge.repetitions_failed` span-attribute doc now describe the wider count (an error OR a non-decisive non-abstained pass; a clean abstention is not counted).
- **19 (fallbacks dropped the diagnostic).** `repetition_consistency_raw` is now populated on both the uniform-plurality path and the no-decisive-votes (position-biased) early return, since it is a diagnostic rather than a weight. Test asserts it survives a `swap=False` run.

**P3**
- **20 (one-sided rule stated unconditionally).** Docs now qualify it: neutral weight on the pooled-fit path, replaced by the judge's consistency weight on the repetition path.
- **21 (error + abstained slipped the count).** `failed_count` is now `not p.decisive and (not p.abstained or p.error is not None)`, so a pass that errored while flagged abstained counts as a failure. Latent today, closed.
- **22 (test gaps).** The HTML test now finds each header's column index and asserts the value under it, so swapping the two cells fails. Added a `swap=False, repetitions=2` test that asserts the single-ordering fallback (empty consistency + the warning), so removing the guard fails.

Also fixed the shrinkage-constant comment @currentlycodinng flagged: `n` in `(n*disc + k*panel_mean)/(n+k)` is the number of datapoints with a qualifying group (one per comparison), not the group/repetition count and unrelated to R. The comment now says that, and gives how fast the prior fades (2 datapoints → 1/3 toward the anchor, 5 → 1/6).

34 tests in `test_pairwise_repetitions.py`; ruff, `ruff format --check` and basedpyright clean on the changed source.

Baukebrenninkmeijer:
@karina.kalicka looking good. One last thing, could you simplify the language in the docs? It's quite technical and complex phrasing.

KarinaKKarinaK:
Done - simplified the repetition-aware reliability section in `docs/pairwise-judging.md` (ca371445). Shorter sentences, dropped the jargon (empirical-Bayes, canonicalized, within-datapoint heterogeneity), kept all the facts and numbers. Merging now.

PR #148, opened 2026-08-13
fix: make simulation criteria actually able to fail (RES-1308)

Baukebrenninkmeijer:
Closes RES-1308.

## The bug

Simulation criteria never failed. Every scenario came back all-green regardless of what the agent actually did.

Three independent causes, each sufficient on its own.

### 1. Transport — the judge never saw the agent

The Orq Responses router **silently drops** an assistant input item whose content is a bare string or a list of `input_text` parts. Only `output_text` parts survive. Some backends 400 instead:

```
cannot unmarshal string into Go struct field messageItemRaw.content of type []json.RawMessage
```

So the judge was handed a transcript with zero agent turns in it. Dumping its raw tool arguments during a live run showed it saying, in plain text, *"the agent has not yet responded"*. No criterion about agent behaviour could be evaluated, because there was no agent behaviour in the input.

This is also why the ticket's Repro C (swap in a stronger judge) changed nothing — a stronger judge cannot score a transcript it was never shown.

Fixed in `openresponses/input_items.py`, the shared renderer used by `OrqResponsesTarget`, `OpenAIAgentTarget` and red-team multi-turn replay. `simulation/agents/base.py` was hand-building its own input list; it now calls the shared renderer.

### 2. Schema — `must_happen` was structurally unfailable

The judge was never asked for a per-criterion verdict. Pass/fail was inferred from a criterion id being *absent* from `rules_broken` — but `rules_broken` is a continue/finish control signal, and a `must_happen` criterion that simply never occurred is not a "rule broken". No judge ever emits one, so it could never fail.

Both judge tools now require a `criteria_verdicts` audit with one entry per criterion.

The audit asks **only "did this occur?"**, never "did this pass?". A pass/fail flag means the opposite thing for `must_happen` and `must_not_happen`, and gpt-5.4-mini reliably inverted it — marking a satisfied `must_happen` as unmet while its own `reason` field said the opposite. Occurrence is one factual question with one answer for both types; the type-to-verdict mapping happens in code. A probe went 0/3 → 3/3 correct on this change alone.

New house rule in `CLAUDE.md`: *never ask a judge for a verdict that inverts between types.*

### 3. Aggregation — mid-run violations were discarded

`criteria_results` was built from the **final turn's** judgment only, so a violation the judge reported on turn 2 disappeared if turn 5 ended the run. And `continue_conversation` hardcoded `rules_broken=[]`.

`_CriteriaTracker` folds every turn's audit into one run-level verdict. Occurrence is sticky (False→True only) — something that happened on turn 2 still happened when the run ends. Everything downstream (`criteria_results`, `criteria_meta`, `SimulationResult.rules_broken`, the `criteria_met` scorer) derives from that one list, so they cannot disagree.

## Also fixed

`criteria_met` returned **1.0** for runs that errored or timed out — a dead target scored a perfect run and inflated the average. Those now score **0.0** with a warning: the run terminated before any criteria audit, so the outcome is unknown, which is not the same as met.

## Degradation

Per the project's "a degraded path announces itself" rule, every partial-evidence case warns rather than quietly passing:

- judge returned no audit for any turn → warn, fall back to `rules_broken`, results flagged unverified
- judge skipped some criteria → warn, treat as not observed (`must_happen` fails, `must_not_happen` passes)
- malformed audit entries → warn with a count, drop them
- non-text assistant content parts → warn, they cannot be represented as `output_text`

The judge's free-text `rules_broken` is consulted **only** for criteria the audit never covered. Letting it override an audited verdict would put the unreliable channel back in charge of the thing this design exists to stop trusting.

## Verification

Unit tests would not have caught either of the first two causes, so this was verified against the live API as well.

- Judge inversion probe: **0/3 → 3/3** correct
- Deterministic end-to-end run: 4/4 assertions, `criteria_met: 0.5` (was 1.0)
- Real Orq agent, multi-turn: a `must_happen` BANANA criterion correctly FAILED, and a `must_not_happen` criterion correctly FAILED when the agent said "card account"
- New `tests/simulation/test_criteria_verdicts.py`: 19 tests over the tracker fold, judge parsing, runner end-to-end and the scorer
- Full suite: 4333 passed, 2 skipped. ruff, whole-repo basedpyright, `mkdocs build --strict` all clean

`eq sim` was checked for parity: the CLI builds no runner or judge of its own, routes through the same `_simulate_run` as the SDK, and never passes `judge=` — so it picks the fix up unchanged.

## Not breaking

`criteria_verdicts` is optional on `Judgment` and defaults to `None`. A custom `judge=` that does not report verdicts still runs; it warns and its criteria results are flagged unverified. Documented in the simulation guide.


Baukebrenninkmeijer:
## Review: 11 agents (`/pr-review-stolen` + `/hate`), recommendations applied

Ran seven `pr-review-stolen` lenses (code, tests, comments, silent failures, type design, spec-alignment, minimality) and four `hate` critics (Skeptic, Pedant, Historian, Realist), all on sonnet, then verified every finding against the source before acting on it.

### Ratings

| Critic | Lens | Rating | Key findings |
|--------|------|--------|--------------|
| The Skeptic | strategy | 🟠 | `rules_broken` union keeps the untrusted channel alive; per-turn audit cost undocumented; audit-coverage info discarded at `resolve()` |
| The Pedant | quality | 🟢 | judge safety-terminate paths score a malfunctioning run 1.0; `max_turns` with no judgment scores 1.0 |
| The Historian | reuse | 🟢 | none — `base.py` now routes through the canonical `messages_to_responses_input`; no new bypass of `common/` |
| The Realist | goal | 🟠 | "flagged unverified" was false as written — a `logger.warning` was the entire mechanism |

`code-reviewer`, `comment-analyzer` and `minimality` each returned no findings (minimality: *"net: -0 lines possible"*). `spec-alignment` confirmed all three of RES-1308's repros are addressed, and that the transport bug is a better explanation for Repro C than the ticket's own hypothesis.

### The one real problem, raised independently by three critics

The Realist, the type-design lens and the Skeptic converged on the same hole, and the Pedant found the concrete exploit for it: **the fix reproduced RES-1308 one layer up.**

`_CriteriaTracker.resolve()` computed exactly which criteria were audited and whether any audit arrived at all — then threw that away after logging it. Nothing on `SimulationResult`, `criteria_meta`, or any span attribute said the verdicts were unverified. So the docs' and CHANGELOG's claim that such a run is *"flagged unverified"* was false as written: a loguru warning was the whole mechanism, and `tests/simulation/test_criteria_verdicts.py` asserted `'unverified' in caplog.text`, encoding "a log line is the design".

The Pedant then found the path that makes it bite the **built-in** judge, not just a custom one. `JudgeAgent._parse_judgment` has three safety-terminate branches (no tool call, unparseable arguments, unknown function) that all return `rules_broken=[]` with `criteria_verdicts` left at `None`. Trip one on the first turn and `_any_audit` never flips, `resolve()` falls back to `[]`, every criterion passes, and `criteria_met` returns **1.0** — while `terminated_by` is `judge`, so the new `_UNEVALUATED` guard never fires. A judge that failed to parse its own tool call scored a perfect, unverified run.

Applied the structural fix in 3b97fda2 — thread the coverage state out instead of discarding it:

- **`SimulationResult.criteria_verified`**, set from the tracker on both termination paths. `None` on runs saved before the field existed, so their scores don't change.
- **`criteria_met_scorer` returns 0.0 with a warning when it's `False`** — the same treatment error/timeout runs already get, reusing this PR's own machinery rather than adding a third flavour of "unverified".
- **`_CriteriaTracker.observe` warns on a verdict id the scenario never defined** (raised by the silent-failure and type-design lenses). Such an entry is well-formed, so the judge's parser doesn't count it as malformed and it vanished silently — an off-by-one across a whole payload just looked unaudited.
- Docs and CHANGELOG updated so the "flagged unverified" claim is now true, with the field to check.

Also closed the e2e coverage gap that let this ship, which the test lens and the Pedant both flagged: the fallback was tested on the tracker and on the parser separately, never wired through `SimulationRunner`, and the three safety-terminate branches had no test at all. Seven new tests (26 total in the file), including the three parametrized safety-terminate paths and a `max_turns`-with-no-judgment case.

Checks: `ruff check src` / `format --check src` clean, whole-repo `basedpyright` 0 errors, **4344 passed / 2 skipped**, `mkdocs build --strict` clean.

### Still open — your call, not applied

**1 — Should `finish_conversation` keep reading the judge's free-text `rules_broken` at all?**

`criteria_verdicts` is `required` in both tool schemas, so `rules_broken`'s only remaining job is rescuing a criterion the audit skipped — the channel this PR exists to stop trusting, kept as a second source of truth.

- **A — derive `rules_broken` purely from the audit:** drop the union at `judge.py:388`, matching what `continue_conversation` already does. Deletes ~15 lines of merge logic and 2–3 tests, and closes the door on free text re-inverting an audited criterion. Costs the partial per-criterion rescue, for which no test shows a real model actually doing it.
- **B — keep the union:** costs the special case and the two tests pinning down how the channels reconcile.

**Lean: B for now.** The rescue is already fenced to unaudited ids only and can only add failures, so it can't undo an audited verdict — and with `criteria_verified` now on the result, a run leaning on that channel is visible. Worth revisiting once there's field data on whether real judges skip entries.

**2 — Should the per-turn audit be trimmed as criteria settle?**

The Skeptic notes the judge re-audits every criterion on every turn, including `continue_conversation` ("even when nothing has changed"). At `max_turns=10` with 8 criteria that's up to 80 structured entries with evidence strings per conversation, and no doc line on the token/latency cost. Trimming to unsettled criteria saves tokens but locks in an early hallucinated `occurred=True` with no chance for the judge to correct it. Genuine trade-off, nobody has made the call — flagging rather than deciding it.

**3 — Per-criterion `audited` flag in `criteria_meta`?**

Partial audits currently only warn. A failed `must_happen` the judge genuinely confirmed and one that defaulted to False because the judge skipped it render identically in reports and the dashboard. Left out because partial audits already fail safe and the run-level flag covers the unsound case; say the word if the dashboard should distinguish them.

Not adopted: splitting this into four PRs (correct in principle, but the transport and schema fixes were each necessary for verifying the other), and dropping the per-turn `rules_broken=violated` in `continue_conversation` — dead today, but harmless and it documents the hardcoded-`[]` bug it replaced.

Baukebrenninkmeijer:
## All three open decisions resolved — ecf64445

### 1. Free text dropped

`finish_conversation` no longer takes a `rules_broken` argument at all. Violations are derived in code from the occurrence audit + `Criterion.type`.

```python
# judge.py — the finish tool's schema
'required': [
    'reason',
    'goal_achieved',
    'goal_completion_score',
    'criteria_verdicts',
],
# No `rules_broken`. The judge is asked what OCCURRED; which occurrences
# count as violations is derived in code from `Criterion.type`.
```

```python
# _parse_judgment, finish branch — same derivation continue_conversation already used
return Judgment(
    should_terminate=True,
    reason=str(args.get('reason', '')),
    goal_achieved=goal_achieved,
    rules_broken=violated,   # was: reported + [cid for cid in violated if cid not in reported]
    ...
)
```

The unaudited rescue in `_CriteriaTracker.resolve()` went with it — `broken += [r for r in judgment.rules_broken if r in unaudited]` is gone. That line handed the free-text channel exactly the criteria the audit had no answer for, i.e. the ones where it was least checkable. A skipped criterion now keeps its not-observed default.

Two things deliberately kept:

- **`Judgment.rules_broken` / `SimulationResult.rules_broken` as outputs.** Only the tool *input* is gone. A custom judge that fills the field itself still works.
- **The total-absence fallback.** If no audit ever arrived there is nothing to prefer it to, so the judgment passes through untouched and the run is marked `criteria_verified=False`. That is the custom-judge path, not the two-channels-competing path.

`test_free_text_rules_broken_still_rescues_an_unaudited_criterion` is inverted to `..._does_not_rescue_...`, and `test_finish_conversation_unions_reported_and_audited` became `..._derives_rules_broken_from_the_audit_alone` — it now sends a `rules_broken` anyway and asserts it does *not* survive.

### 2. Audit trimmed to what serves early stopping

Kept the per-turn re-audit; stripped only the part that cannot contribute to a stop decision. Occurrence is sticky, so a criterion already confirmed to have occurred is **settled** — a later turn cannot change its run-level verdict, and re-reporting it cannot trigger a stop that the turn it first occurred on did not already justify.

A settled criterion stays in the prompt — the judge still needs it to weigh whether to end the conversation — but is excluded from the audit payload:

```
MUST HAPPEN:
- criteria_0: <...> (must_happen) [ALREADY CONFIRMED — do not re-report in criteria_verdicts]
- criteria_1: <...> (must_happen)
```

That drops an id, a boolean and an evidence quote per settled criterion per remaining turn. The Skeptic's objection — "locks in an early hallucinated `occurred=True` with no chance to correct" — is already true of stickiness itself (`_occurred[cid] or occurred`), so trimming commits to nothing the fold hadn't.

Two implementation notes worth flagging in review:

- `mark_settled` **rebinds** a frozenset rather than mutating one. The runner does `copy.copy(self._injected_judge)` per simulation — a shared mutable set would leak occurrence between concurrent runs, which is the CLAUDE.md house rule on shared state.
- Wired **duck-typed** (`_implements(judge, _SETTLEABLE_JUDGE_METHODS)`), not `isinstance`. This module already carries a comment on why isinstance-against-Protocols is avoided here, and the duck-typed form is testable — `test_runner_tells_the_judge_which_criteria_have_settled` asserts the exact per-turn settled sets. A custom judge without the method keeps auditing everything.

Also guarded the "no `criteria_verdicts` returned" warning so it counts *unsettled* criteria — otherwise it cried wolf on the legitimately-empty payload once everything settled.

### 3. `audited` on each criterion

`criteria_meta` entries gain `audited`, sourced from the tracker's `_seen`:

```python
for c in result.metadata['criteria_meta']:
    if not c['passed'] and not c['audited']:
        print(f"{c['id']} failed by default — the judge never reported on it")
```

A `must_happen` the judge confirmed never occurred and one it silently skipped both render `passed: False`; only this separates them. `None` when no tracker was available, which reads as unknown rather than as audited — `test_build_criteria_meta_without_a_tracker_reports_audited_unknown` pins that.

---

Docs and CHANGELOG updated for all three. Checks: ruff clean, whole-repo basedpyright 0 errors, **4353 passed / 2 skipped**, `mkdocs build --strict` clean.

No decisions left open.

Baukebrenninkmeijer:
## Judge tool schemas are now generated from pydantic models

Ten commits since the last review round. The score-side RES-1308 fix was already in; this round replaces the hand-written schemas that caused it and then chases the same bug through every reporting surface it was still hiding in.

### The core change

`judge.py` asked the LLM for its terminate decision through two alternative function tools whose JSON schemas were hand-written dicts, and read the reply back with `json.loads` plus four hand-rolled coercion helpers. The schema half and the parser half drifted, and the drift was invisible — **a field the parser stopped reading looks exactly like a field the judge chose not to report.** That is how RES-1308 happened.

The schemas are now generated from `_JudgeToolArgs` / `ContinueConversation` / `FinishConversation`. One definition, no second copy to fall behind.

### Why `WIRE_REQUIRED` is explicit rather than pydantic's own `required`

The wire contract and the parser contract are deliberately different contracts, and this is the part worth reviewing carefully.

Every field carries a parser-side default, so one malformed field cannot cost a whole turn's audit. But that means `model_json_schema()` emits **no `required` key at all** — which silently un-requires `criteria_verdicts` and reintroduces the exact bug this branch fixes. An earlier draft of this refactor did precisely that, and three independent reviewers caught it.

So `required` comes from an explicit `WIRE_REQUIRED: ClassVar[frozenset]`, with `_assert_wire_fields_exist()` running at import to check it against `model_fields` so a rename cannot half-land. Same story one level down: `evidence` has a parser default, so pydantic omitted it from the nested item schema too — hence `_NESTED_WIRE_REQUIRED`. `default` is stripped from the generated schema for the same reason: advertising a default tells the model the opposite of what `required` does.

`tests/simulation/test_judge_schema.py` pins all of it, and an AST guardrail in `tests/test_reuse_guardrails.py` bans hand-written `'properties': {...}` literals under `src/` so a third copy can't grow back.

### Identity

`criterion_id` is the identifier everywhere — not an integer index. A string id is *copied* from the prompt: an off-by-one goes out of range, gets dropped, and warns. An integer is *counted*: an off-by-one lands on the neighbouring, usually opposite-typed criterion and **silently inverts a verdict**. `criterion_id_for()` and `CRITERION_ID_PATTERN` are now the only way to build or read one; nine inline f-strings are gone.

### What the whole-branch review turned up

The score side was correct end-to-end. The reporting side had outrun it — an unaudited run still displayed as clean in four separate places:

- the `criteria_met` evaluator returned `pass=True` beside its own `0.0`, and that flag lands on the evaluator span and the uploaded Orq experiment
- three renderers (markdown/HTML, the FastHTML dashboard, the Streamlit `eq sim ui`) rendered "2/2 criteria met — PASS" next to `criteria_met 0.0`
- a target failure discarded the entire audit, so a `must_not_happen` violation the judge had already confirmed on turn 2 vanished if the target died on turn 4
- span inputs recorded assistant turns as a Python repr after the `output_text` fix — the defect CLAUDE.md's `content_to_text` row names

Fixing that fold then introduced the mirror-image bug twice, both caught and fixed: a dead target fabricating *confirmed failures* out of criteria nobody audited, and — one notch narrower — a cut-short run reporting an unmet `must_happen` as failed when it simply hadn't had its chance yet. **Unknown is its own state in both directions**, and `criteria_meta` now carries `audited` and `evidence` so every surface can say so.

### Verification

`ruff check src`, `ruff format --check src`, bare `basedpyright`, `pytest -m 'not integration'` (4433 passed), `mkdocs build --strict` — all green.

### Two things worth a reviewer's eye

**Not covered by any test here:** whether the judge actually complies with the prompt's "skip ALREADY CONFIRMED criteria" instruction, and whether ids come back misattributed. That is a live-model question — it needs `scripts/live_trace_validation.py`, not a fake. Worth running before merge.

**Accepted deliberately:** `pattern: ^criteria_\d+$` now reaches the wire. Fine for OpenAI; the narrower schema dialects behind the Orq router that motivated `$defs` inlining are the same ones most likely to reject it. Flagging it as a conscious choice rather than a pydantic side effect.

PR #149, opened 2026-08-14
fix(cost): record whether a cost is billed or estimated (RES-1307)

Baukebrenninkmeijer:
Closes [RES-1307](https://linear.app/orqai/issue/RES-1307/cost-figures-dont-record-where-the-price-came-from-estimate-vs). Was stacked on #145 (RES-1295), now **merged** — the diff here is everything after `8cb2352`.

## Why

After #145 a cost figure comes from one of two very different places and lands in an identical shape either way:

- **provider-reported** — the Orq router prices `/v3/router/responses` and returns `usage.input_cost`/`output_cost`/`total_cost`. This is the billed number.
- **client-side estimate** — anything on chat completions is priced by `model_catalogue.price_usage()`, token counts × the catalogue's per-1k rate.

The estimate set `total_cost` and `priced_calls = calls` exactly like the real thing, so no consumer could tell them apart. Two ways that misled:

1. The estimate is knowingly approximate — a flat input rate, while cached-read and cache-write tokens bill at a discount/premium the catalogue does not publish. A cache-heavy call reads high.
2. It defeated the one partial-cost signal we had. `cost_is_partial` is `0 < priced_calls < calls`, and the docs tell users to compare those two counts — but an estimate registers as *complete*, so the coverage clause went silent exactly when the estimates arrived.

Same theme in the second half: `JudgeOutcome.endpoint` (added in #145 so a missing cost can be diagnosed — "model not in the catalogue" vs "Responses 400'd and fell back to chat") reached the saved artifact only on the single-judge path. Static-mode runs use the panel path, which dropped it: `judge endpoints: [null, null, null, null]` on a live 4-attack run.

## Cost provenance

`Usage` gains `estimated_calls: int` — how many of `priced_calls` were client-side estimates — and a derived `cost_source` property (`provider` / `catalogue` / `mixed` / `None`).

A **counter**, not a bool, and derived rather than stored: it sums through `__add__` and clamps through `__sub__` with no special case, the way `priced_calls` already does, and a run mixing billed and estimated calls reports `mixed` instead of collapsing to one of them. Saved reports lacking the key default to 0 → `provider`, which is correct: client-side estimation is itself unreleased, so no shipped report can be mislabelled.

Producer side is one line. `price_usage` is the only place a catalogue estimate is ever produced (verified: `cost_per_1k` has no users outside `model_catalogue.py`), so stamping it there covers judges, targets and jobs with no per-surface plumbing.

Render side is one function. `cost_coverage(priced_calls, calls, *, estimated_calls=0)` in `common/reports/md_helpers.py` is the single label helper every surface already called, so coverage and provenance now compose instead of one masking the other:

| | all provider | all estimated | mixed |
|---|---|---|---|
| fully priced | `''` | `' (estimated)'` | `' (partly estimated)'` |
| partly priced | `' (3 of 10 calls)'` | `' (3 of 10 calls, estimated)'` | `' (3 of 10 calls, partly estimated)'` |

`estimated_calls` is then carried through the red-team and simulation report serializers and their renderers, and through every dashboard accumulator, stats dataclass and label — the renderers read Usage-shaped dicts from saved reports, so a value the serializer omits is invisible downstream. One shared `resolve_cost_source(priced, estimated)` in `contracts.py` backs both `Usage.cost_source` and `cost_coverage`, so the three-way rule exists once.

## Judge endpoint on the panel path

`common/jury.Prediction` gains `endpoint`, `JuryDeliberation` aggregates it (`mixed` when a panel used both, `None` when nothing recorded one), and all three judge-fn adapters stamp `outcome.endpoint`. Additive and `None`-defaulting, since `common/jury.py` is shared machinery and older artifacts must still deserialize. The catch-all `Prediction(error=str(exc))` inside `jury.py` has no `JudgeOutcome` to read and stays `None` rather than inventing a value.

Three artifact writes, all under the same `raw_output['endpoint']` key the single-judge path already used, so consumers read one key:

- `redteam/adaptive/evaluator.py` panel path
- `redteam/frameworks/owasp/evaluatorq_bridge.py` — **both** legs. This is what `runner.py` uses for static mode, and it is where the ticket's `[null, null, null, null]` actually came from; fixing `adaptive/evaluator.py` alone would not have moved it.
- `llm_jury()` — computed the endpoint and dropped it, which would have left the jury surface the last one blind.

## Docs

`docs/guides/red-teaming.md` "What a run costs" rewritten: coverage and provenance are separate axes, and a total can be fully covered and still be a client-side guess. CHANGELOG entry under `### Notable defaults`.

## Verification

`ruff check src`, `ruff format --check src`, `basedpyright` (bare, whole repo), `pytest -m 'not integration'` → 4419 passed, 2 skipped. Every task passed a spec + quality review; a whole-branch review found the `llm_jury()` drop and a dashboard label where the provenance clause displaced the priced count — both fixed in `f29ff8a` and re-reviewed.

~~Not covered: per-run dashboard rows still render a bare `$x.xxxx` with no provenance, matching how per-run coverage was already handled.~~ Fixed in the audit follow-up below. The estimate itself is still a flat input rate — splitting it by cache tier needs `/v2/models` to publish the tiers.

## Follow-up: five-chain provenance audit

After the work above landed, five read-only auditors swept the other chains — secondary generation, core red-team/simulation, judging, recommendations, dashboards — against six invariants (usage captured; unpriced calls priced or honestly unpriced; provenance survives aggregation; hand-built usage dicts complete; every rendered dollar qualified; judge paths stamp their endpoint). None of what they found was a regression from the work above; it is the same defect class in places this PR had not reached. Six further commits fix it.

**Usage that was billed and then dropped.** Judge calls that error, time out or return unparseable JSON are billed like any other. Three sites threw that usage away — the OWASP bridge's panel `judge_fn` failure branch, its single-judge error/timeout legs, and `common/judge.py`'s legacy `json_object` parse-failure handler, where `usage` was a local that was gone by the time the handler ran. All three now carry it, matching `adaptive/evaluator.py`, which already did.

**Target usage from every attempt, not just the last.** `call_target_with_retry` returned only the surviving `AgentResponse`, so an attempt that returned an `.error` marker *with* a real usage block, followed by a successful retry, had its tokens dropped. `TargetCallResult.billed_usage` now sums every attempt that reported usage. The four consumers replace their `response.usage` read with it rather than adding to it — on the single-success path the two are the same value, so joining them would double every run total; a test pins that equality.

**Call counters, fixed at source.** The four consumers disagreed about what `calls=0` from a target means: the orchestrator repaired it two different ways, one of which widened `priced_calls` alongside `calls` — turning "two attempts, one of which reported a cost" into a `'provider'` cost claiming full coverage — while the other three surfaces took `calls=0` verbatim and reported zero calls for a real exchange. `call_target_with_retry` now normalises each attempt's counters before summing: a target that tracks its own `calls` is left alone (orq rolls tool-continuation rounds into one block, so its count beats anything we could infer), and one reporting `calls=0` gets `calls=1`, `priced_calls=1` iff that attempt reported a cost. Both orchestrator fallbacks became dead code and are gone. A custom `AgentTarget` that hand-builds a cost with no priced call now gets a warning rather than an unqualified figure.

**Endpoint provenance on the two surfaces that still lacked it.** `llm_jury()` wrote `raw_output` only under `assignment='cyclic'`, so the documented **default** computed the endpoint and discarded it; it is now written on both paths. `pairwise` gained a `PairwiseComparison.endpoint` that reaches the saved artifact, folded across orderings the same way the jury folds across judges. `common/judge.py`'s terminal error returns stamp the endpoint where control flow makes it knowable — the chat span is unreachable while `responses_model` is live, and the responses leg's only clearing path sets it to `None` before falling through, so the inference is total. The genuinely-unknowable catch-all stays `None`.

**Every rendered dollar figure qualified.** The per-run dashboard Cost column (called out as not-covered above) now carries a compact `~` marker with the full wording in its tooltip; the legacy red-team branch that added dollars to Total spend without adding any coverage counters now counts those calls as unknown-coverage; "Avg cost / job" is qualified, and "Spend by job type" no longer subtitles itself "Real cost across runs" — `cost_by_kind` genuinely has no per-kind provenance, so it says so instead of faking six per-bar qualifiers. The input/output split is qualified on all six renderers: a split shown unqualified next to a qualified total implies the split is exact. LLM spans emit `gen_ai.usage.cost_source` alongside the cost.

**Counters, cache tiers and dead plumbing.** The openai-agents integration read `calls=1` off an aggregate that carries a real `.requests` count, undercounting any tool-using run and making `priced_calls == calls` look complete when it was not; it now matches its crewai and pydantic-ai siblings. LangChain's `cache_creation` input detail was never read, so cache-write tokens stayed 0 on every LangChain-fronted Anthropic call. `Usage.__sub__` clamped its counters independently and could yield an invalid triple; the chain `estimated_calls <= priced_calls <= calls` now holds by construction. `generate_single_prompt` and `token_usage_summary` had no callers and are deleted, along with the dead read that could never fire.

Each task was implemented and reviewed independently, and the reviews were adversarial: one overturned an assertion an earlier task had written (a judge PARSE failure now records `'chat'`, because the fallback that served it is derivable, not fabricated), one traced all four consumers to their run totals to rule out double counting, and one reproduced the mixed-cost-reporting case that drove the source fix. `ruff`, `basedpyright` (bare), `pytest -m 'not integration'` (4481 passed, 2 skipped) and `mkdocs build --strict` are green.

Out of scope by decision: `common/apply.py` discards the `Usage` from its recommendation-merge completions. That is a dashboard action, not a report path.

---

## Review decisions

A 7-agent review of the branch, then a 4-critic adversarial pass over the resulting fixes. Recorded here and on [RES-1307](https://linear.app/orqai/issue/RES-1307) so the next reviewer doesn't re-litigate them.

**1. A catalogue-estimated cost is stamped onto the span, rather than the docs narrowed to match the code.**
`docs/tracing.md` documented `gen_ai.usage.cost_source` as taking `catalogue`/`mixed`, but all three helpers in `common/llm_call.py` recorded the span from the raw response and only *then* called `price_usage` — so an estimate never reached a trace and the attribute could only ever read `provider`. A report showed dollars the trace viewer did not. The cheap fix was deleting the wording; instead `_price_and_record_cost` prices first and stamps after, which makes the documented values real. Costs a second write to the same span per call.

**2. The judge endpoint is exposed as a span attribute, not rendered in reports.**
`JudgeOutcome.endpoint` reached the saved artifact (the ticket's third criterion) but nothing in running code read it back, so diagnosing meant grepping JSON. Rendering it in the token-usage section was rejected as new report surface on a ticket already 2× its scope. `gen_ai.judge.endpoint` lands next to `cost_source`, where the missing cost is actually observed. Noted for the record: the span already carried `operation='responses'`/`'chat'`, so it is partly redundant — kept because it names *why* a judge span lacks a cost instead of leaving that to be inferred.

**3. The call is counted at the parse site; `Usage.with_calls` is deleted.**
`from_completion` parsed chat usage as `calls=1` while `from_openresponses` parsed the same thing — one response object, one API call — as `calls=0`, leaving each caller to patch the 1 back on. Both callers did so on the line after the parse. Since `extract` already derives `priced_calls` exactly as `with_calls` did, the two forms were the same value by two routes. This also removes one of the two `model_copy(update=...)` paths that bypass the counter-chain validator, and with it a way for a call site to forget the stamp and report a real exchange as zero calls.

**4. The remaining `model_copy` bypass stays, and the `Usage._with()` funnel is dropped.**
Two critics argued the validator's "by construction" claim was misleading while `model_copy(update=...)` skips it, and proposed funnelling every mutation through a revalidating helper. With decision 3 removing one bypass, the only one left is `target_call._attempt_usage`, which rewrites counters on a usage it did not parse and so genuinely cannot go through the parse path. It clamps by hand and says why. Not worth building the funnel for one justified site.

**5. The PR stays combined rather than being re-split.**
The branch is roughly 2× the ticket — the audit follow-up carries billing-accuracy fixes and two chains (LangChain cache tokens, dead-code removal) with no connection to provenance. Unwinding a reviewed, green stack costs more than it recovers. Going forward, an audit sweep that turns up adjacent fixes belongs in a stacked follow-up PR.

### Findings rejected

- **"Clamp negative counters in the validator instead of raising."** `tests/redteam/test_review_followup.py` deliberately asserts negative counts are *rejected* — a negative is a corrupt payload, not a stale one. Applied, caught by that test, reverted; the docstring was scoped instead.
- **"Delete `TestAdaptToolCall`/`TestAdaptStaticOutput` as unrelated scope."** They are the only coverage for two still-live functions in `common/output_adapters.py`.
- **"`Usage.with_calls` self-corrects now the validator exists."** It does not — `model_copy(update=...)` bypasses pydantic validators entirely. `with_calls` was genuinely broken and needed its own clamp before decision 3 removed it outright.
- **`record_token_usage(judge_endpoint=...)`** was added, then found to have zero callers and to document values that disagreed with production. Deleted with its two unit tests; replaced by a test on the live path.

### Open at merge

The judge catch-all now stamps `_failed_endpoint()` where a previous author explicitly declined to. The inference holds for a failure that dispatched a request; a pre-flight failure (bad `extra_kwargs`) has no test pinning it.


ChielToto:
#### What worked / tested

- All checks pass: Lint, Build (strict), Examples, SDK compat 4.11/4.12/4.13, and Typecheck + test on 3.10 through 3.13.
- `resolve_cost_source` as one shared function backing both `Usage.cost_source` and `cost_coverage` is the right call. The three-way rule exists in exactly one place, and `md_helpers` can't call the property because it only has raw ints.
- `_PROVENANCE_LABELS` as a mapping instead of if/elif/else means a fourth literal renders an empty clause rather than falling into whichever branch happened to be `else`.
- A counter rather than a bool, derived rather than stored: it sums through `__add__` and clamps through `__sub__` with no special case, `mixed` is representable, and old reports default to 0 → `provider`, which is correct since client-side estimation is unreleased.
- I traced `_failed_endpoint()` (`common/judge.py:452`) and the inference holds: `responses_model = None` is set inside the `BadRequestError` handler before the chat span opens, so a live `responses_model` when an exception escapes really does mean the Responses leg raised. Not a guess.
- Test coverage is real, not decorative. `tests/common/reports/test_md_helpers.py` parametrizes all six coverage/provenance combinations plus the inconsistent-caller case; `test_target_call.py` pins `billed_usage == response.usage` on the single-success path, which is the assertion that makes "replace, don't add" safe at the four consumers.
- Two genuine bugs picked up along the way: LangChain's `input_token_details.cache_creation` was never read, so cache-write tokens were 0 on every LangChain-fronted Anthropic call; and openai-agents read `calls=1` off an aggregate that carries a real `.requests` count.

#### Issues

1. `Usage.with_calls()` is removed (`contracts.py`) with no CHANGELOG entry. It's a public method on `Usage`/`TokenUsage`, which is exported and documented. All three in-repo callers are updated correctly, but external code that hand-builds usage for a custom `AgentTarget` gets an `AttributeError`. The PR documents `generate_single_prompt` and `RedTeamReport.token_usage_summary` removals in detail, so this one just slipped.
2. `AgentResponse.from_openresponses` now returns `usage.calls == 1` instead of 0 (`contracts.py:1145`). That's a semantic change to a public parser: any external caller that patched `+1` afterward, which is exactly what evaluatorq itself did in two places until this PR, now double counts. The failure is silent and lands in the worst place — inflated `calls` alongside inflated `priced_calls` makes a partial cost read as fully covered. Also not in the CHANGELOG.
3. Scope. Base is another unmerged PR, and the description's second half ("five-chain provenance audit", six more commits) is by your own framing a different defect class from the RES-1307 fix. Cost provenance, judge-endpoint plumbing, target-usage accumulation, call-counter normalisation, two integration bugs and dead-code removal in one 4.2k-line unit is more than one reviewer can hold. Three PRs stacked would have reviewed the same and merged incrementally.
4. `Usage._clamp_call_counts` (`contracts.py:545`) reads `data.get('calls') or 0`, and `calls` itself defaults to 0. So `Usage(total_cost=0.02, priced_calls=1)` with no `calls` now silently clamps `priced_calls` to 0, flipping the figure from "priced" to "no coverage data". I checked every `priced_calls=` construction in `src/` and none does this, so nothing in-repo is affected, but it's a silent clamp on a public model and it's the shape a hand-built target usage most plausibly takes.
5. `call_target_with_retry` logs at WARNING for every error-marker attempt carrying usage (`common/target_call.py:266`). On a target that legitimately retries, that's per-attempt noise about a condition the code now handles correctly. `logger.debug` reads better.
6. The same custom-target shape (cost with `priced_calls=0`) trips both `_attempt_usage`'s warning and `record_token_usage`'s new "cost recorded without provenance" warning, so that user gets two WARNING lines per call for one upstream mistake.

#### Nits

- The coverage clause renders three times on the landing KPI screen: under `Avg cost / job`, appended to `Total spend`, and again in the "Spend by job type" subtitle. Each is individually justified in the comments; together it's the same parenthetical three times in one viewport.
- `_run_grid`'s new empty state ("No runs on this page.") is a good change with nothing to do with cost provenance.
- `_attempt` resets `usage = None` per retry, so tokens billed on attempt 1 are dropped when attempt 2 raises pre-flight. Consistent with how `raw_content` is scoped, and it under-reports rather than over-reports, but given the PR is specifically about not losing billed tokens it's worth a line in the docstring.

#### Rating

Rating: 4/5. The design is right and the reasoning is unusually well documented in the code itself, tests back each claim, and CI is green — held back by two undocumented public-API changes and a size that makes independent verification harder than it needed to be.

Baukebrenninkmeijer:
Thanks — four of the six are now fixed in `ce0c7729`, two I'm leaving with reasoning.

**4 (the real bug).** You're right, and it's worse than a nit: `calls` defaults to 0, so the clamp was reading a field that hadn't been set. An omitted `calls` now widens to the priced count instead of clamping it away; an explicit `calls=0` still wins the clamp, since a caller who states it means it. Both branches pinned in `tests/redteam/test_review_followup.py`. Nothing in `src/` hit it, as you found, but it's exactly the shape a custom `AgentTarget` produces.

**1 and 2.** Both in the CHANGELOG under Breaking Changes now. Agreed that 2 is the dangerous one — the failure mode you describe (external `+1` still applied, inflated `calls` *and* `priced_calls`, partial cost reading as fully covered) is stated explicitly in the entry rather than left as "semantics changed".

**5.** Dropped to `logger.debug`, with your reasoning in a comment so it doesn't get "fixed" back up. Counting a failed attempt's tokens is correct handling, not a degraded path.

**Nit 3.** Traced it: `call_target_with_retry` does accumulate across attempts, so target-side retries aren't lossy. The reset is in `judge._attempt`, and there it's deliberate — the outcome's `usage` has to describe the call that produced its verdict, so a retried judge under-reports rather than mis-attributes. That trade-off is now written down at the reset.

**6.** Leaving both warnings. They fire from different layers and only overlap for one target shape: `_attempt_usage` names the fix (`set priced_calls on the usage this target returns`), `record_token_usage` is the generic backstop for callers that never go through `target_call` at all. Downgrading either loses coverage for a path the other doesn't see. If it turns out noisy in practice, the right fix is dedup at the sink, not dropping a signal.

**Nit 1.** Also leaving, and I think it's forced rather than sloppy: the CHANGELOG now guarantees that an unqualified dollar figure means every call behind it was provider-billed. `Avg cost / job`, `Total spend` and the per-kind bars are three separate figures, so dropping the clause from any one of them makes that figure a false claim. Three repetitions in a viewport is the cost of the invariant. A band-level footnote would fix the redundancy without breaking it — worth doing, but as a UI change on its own rather than inside this diff.

**3 (scope).** Fair, and conceded — it's recorded as a decision in the PR body and on RES-1307 rather than argued away. The rule I'm taking forward: an audit sweep that turns up adjacent fixes gets a stacked follow-up PR, not appended to the ticket that inspired it. Unwinding it now costs more than it recovers, but it shouldn't have been one unit.

PR #153, opened 2026-08-17
fix: bug-hunt remediation — retry amplification, Python 3.10, error reporting (RES-1320)

Baukebrenninkmeijer:
Remediation of the verified defects from the map-reduce bug hunt (RES-1320), plus the two shipstoppers a final 11-critic review round caught in the remediation itself.

## What this fixes

**Retry amplification.** Every target call path now makes exactly `max_target_retries + 1` HTTP attempts. Measured at transport level (`httpx.MockTransport`) across all eight paths — openai auto/injected, openresponses auto/injected, orq auto/injected, the simulation client, the judge. Before this, the injected OpenAI/OpenResponses paths made **9 attempts where 3 were intended**: the SDK's own `max_retries` and `with_retry` were stacked, and they multiply.

> **Correction (`ca581dab`).** As originally written that claim was false for one of the eight legs. `SimulationAgent._call_chat_completions` / `_call_responses` called `with_retry` with no `max_attempts`, so the simulation client always made **5** transport attempts regardless of `LLMCallConfig.retry_count` — and the matrix leg for that path monkeypatched `with_retry` itself to force the expected budget, so its assertion held no matter what production did. It was the only leg of the eight not actually measured at transport level. The budget is now threaded from config, the monkeypatch is gone, and the leg was confirmed to fail against its parent commit (observed `attempt 2/5` … `4/5`). **The simulation agent default drops from 5 attempts to 2**; see the CHANGELOG's `Notable defaults`. The chat-completions path additionally retries once *within* a transport attempt on an empty response — a content-level retry, so a model that keeps returning nothing still costs up to `2 × (retry_count + 1)` calls. That is now documented rather than invisible.

**Python 3.10.** `Task.cancelling()` is 3.11+, and it had landed in a `finally` block despite `requires-python = ">=3.10"` and a CI matrix that runs 3.10. Removed; the full suite passes on 3.10.

**Errors reported as clean runs.** A dead target came back RESISTANT because the job emitted no `error` key, so the judge scored the literal `[ERROR: ...]` marker as a genuine reply. Both static legs now emit the key unconditionally, both converter paths flatten through `TargetCallResult.error_payload()` (so `error_stage` / `error_code` / `error_turn` / `error_details` survive alongside the message), and a malformed `criteria_meta` entry scores `0.0` and lands in the errors section instead of reporting PASS.

**Billed runs discarded by display faults.** A `BrokenPipeError` on the final progress update threw away a run whose datapoints had all completed and been paid for. Every progress-display call on a success path now warns and continues.

**Report and dashboard accounting.** Pre-execution failures were double-counted, so bucket counts did not sum to the attack total, and the HTML export rendered **zero resistant attacks**. Both fixed and pinned. Timed-out simulation runs no longer report `$0`.

**Prompt-injection boundary.** `common.sanitize.delimit` had several escape gaps around closing-tag prefixes, split tags and attribute variants — including one backtracking case. Closed, with the linear-time property measured.

Plus tool calls surviving into judged transcripts in a valid wire shape, structured output rendered as JSON rather than a Python repr, cancellation propagating as `CancelledError`, and the library no longer calling `sys.exit`.

## How it was built

Subagent-driven: 31 tasks, fresh implementer each, adversarial review gate per task, executed across five isolated worktrees in parallel and cherry-picked back. Then a final review at real scale — 11 critics (4 opus Claude, 3 sonnet Claude, 4 Codex `gpt-5.6-luna`) — which found the retry amplification and the 3.10 break that the whole prior process had missed. Each of those got its own fix round and scoped re-review before merge.

A further 13-critic round (4 Claude critics, 1 Codex `gpt-5.6-luna` outsider, 8 PR-review lenses) then found the fake retry-matrix leg above, which 11 critics and every prior gate had passed over — the test was green and the claim was wrong. That round landed three more commits:

- `ca581dab` — the retry-budget fix and its now-real matrix leg.
- `4dc9b5d5` — **one** `extract_status_code` (two existed on the same call path and disagreed on text-parsed statuses, so an exception could be *reported* as a 4xx and *retried* anyway); `_warn_ignored_target_retries` moved out of `common/retry.py`, which held the package's only `common/ → redteam/` import; `TargetCallResult.succeeded` derived from `error` instead of stored alongside it, since as independent fields they could be constructed disagreeing and `error_payload()` keys off `error` — the same dead-target-reads-RESISTANT shape this PR fixes elsewhere, reachable through the type built to prevent it.
- `f88fd29f` — an empty deployment reply now always logs (empty string / empty list / list with no text parts returned `''` silently while the sibling `content=None` case warned); `criteria_met` logs on a present-but-empty `criteria_meta`; the `parallelism` docstring corrected from "three independent scopes" to the one shared per-datapoint semaphore it actually is, plus the `Notable defaults` entry the evaluator-bounding change should have had.

Decisions and rejected findings from that round are recorded on RES-1320.

## Verification

- `uv run ruff format --check src` — clean
- `uv run basedpyright` (whole repo, unscoped) — **0 errors**
- `uv run pytest -m 'not integration'` — **4722 passed / 2 skipped** on 3.13 (4718 before the three review commits; 3.10 last measured at **4687 passed / 5 skipped**, before those commits)

Every behavioural fix is pinned by a test proven to fail against the parent commit; tests that turned out to pass on the parent were deleted rather than left as false coverage. The one exception found — a leg that passed only because it monkeypatched the function under test — is corrected above.

## Parked

`redteam/contracts.py` flushes text preceding a tool call to its own assistant row — wire shape is paired and valid in both directions, only the same-row invariant is unmet. A leaked `Live` object if `Live.stop()` itself raises. 11 pre-existing printf-style loguru calls elsewhere in the repo — loguru interpolates `{}`, not `%s`, so these render the placeholder literally (`redteam/reports/converters.py` ×4, `redteam/adaptive/objective_generator.py` ×3, `simulation/tracing.py` ×2, `openresponses/tracing.py`, `redteam/runner.py`). None are introduced by this branch: the printf-style calls it *does* add are all in modules that use stdlib `logging`, where the syntax is correct. Early stopping is split out as RES-1328.

Closes RES-1320.


currentlycodinng (approved):
Approving. I checked the headline claims by reproducing them rather than reading them, and the ones that matter hold. One line in the CHANGELOG says something the code does not do, and there are a few loose ends worth a follow-up, but none of it should hold the merge.

## Verified by running it

**Retry amplification is real, and the fix lands.** I wrote my own transport-level probe rather than borrowing the branch's, so the measurement is independent of the tests being added. Against `origin/main`, an injected OpenAI client at the SDK default with `max_target_retries=2`:

```
main:    intended=3   actual HTTP attempts=9   result.attempts reported 3
pr153:   intended=3   actual HTTP attempts=3
```

So the reported attempt count was wrong too, which is the part that makes this hard to notice in the field: the run tells you it tried three times while the bill says nine. The clone through `with_options` is the right shape, since transport, auth, base URL and headers survive and the caller's client is untouched.

**`sanitize.delimit` holds.** Zero escapes across roughly 35 hand-built breakout attempts (closing-tag prefixes, split tags, attribute variants, nested and doubled forms) and 60,000 fuzz cases. The linear-time claim measures out: `'<' * 60000 + '/data'` in 3.43ms and `'</dat' * 60000` in 5.94ms, which is the backtracking case behaving.

**The 3.10 number is exact.** `UV_PROJECT_ENVIRONMENT=.venv310 uv run --python 3.10 --all-extras python -m pytest tests/ -q -m 'not integration'` gives 4687 passed and 5 skipped, matching the description, and `cancelling()` is gone from `src/`.

**"Pinned by a test that fails on the parent" survives a spot-check.** Copying thirteen of the new tests onto `origin/main`, eleven fail there and the two that pass include the explicit control `test_successful_judge_call_is_unaffected`, which is meant to. `test_transcript_tool_calls.py` cannot even import on the parent.

**The display-fault handlers only wrap display work.** Every new `except Exception` sits inside `_safe_progress_call` or the polling loop and guards a `progress_service` call and nothing else, and the streaming path re-raises `CancelledError` ahead of the broad clause, so a cancelled caller keeps asyncio's semantics. The rewritten gather block collects task failures separately from the poller it deliberately cancelled, which is the fiddly part and it reads correctly.

**The donut fix works, by the fallback rather than the branch it looks like it takes.** `_render_donut_chart` reads `unevaluated_attacks` first, but `_build_summary_section` never puts that key in `summary_data`, so it always computes `total - evaluated`. That happens to be the right number and it agrees with `dashboard.py:923`, so the output is correct, but the lookup is dead and the next reader will assume it fires.

## The CHANGELOG says the judge is unchanged, and one cell of it is not

"Judge, simulation and pipeline calls are unchanged, each already had a single owner." I ran the full matrix of `retry_count` against an injected client's `max_retries` on both sides:

```
                                              main    pr153
retry_count=0, injected max_retries=0     1        1
retry_count=0, injected max_retries=3     4        1     <-- changed
retry_count=1, injected max_retries=0     2        2
retry_count=1, injected max_retries=3     2        2
retry_count=3, injected max_retries=0     4        4
retry_count=3, injected max_retries=3     4        4
```

Five cells match, one does not. The removed `_without_client_retries(client, retry_count)` had an `if retry_count <= 0: return client` guard, and the new parameterless `without_client_retries` drops it, so a caller who set `retry_count=0` precisely because their own client was doing the retrying now gets a single attempt and no rate-limit protection at all.

I think the new behaviour is the defensible one, since `retry_count=0` should mean zero retries and the single-owner rule is the whole point of the branch. What is not defensible is the sentence claiming nothing changed, because that is the sentence someone will read when their judge starts failing on 429s after upgrading. Either restore the guard or say plainly in the breaking-changes list that `retry_count=0` now disarms an injected client's SDK budget as well.

## Smaller

`check_pass_failures` no longer has a single production caller. Removing `sys.exit` from a library is right, and the breaking-changes entry covers the decision, but this PR also *improved* that function in the same diff (an errored evaluator now counts as a failure) while removing the last thing that called it, so the improvement ships as dead code pinned only by unit tests. It is also not exported from `evaluatorq/__init__.py`, so a caller told to "inspect `pass_` and choose their own gate" has to reach into `evaluatorq.evaluatorq` for the helper that already does it. Confirmed end to end: a run with a failing evaluator exits 1 on main and 0 here, with nothing in the return value flagging it beyond the raw scores.

The semaphore change is a real concurrency change that neither the description nor the CHANGELOG mentions. Evaluators used to run unbounded (the deleted comment said so) and now share the per-datapoint semaphore with the job. That is the better default and the `evaluatorq()` docstring documents the new "three scopes" wording properly, but at `parallelism=1` evaluators now serialise where they used to fan out, which will read as a slowdown to anyone who set 1 to serialise jobs. Worth a line in the changelog. While you are there, `types.py:248` still describes `parallelism` as "Number of jobs to run in parallel", so the two docstrings for the same field now disagree.

PR #154, opened 2026-08-17
Docs: streamline red-team and simulation guides

Baukebrenninkmeijer:
## Summary

- Refocus the red-team and agent-simulation guides on first-run workflows and user decisions.
- Add an interactive cost estimator using `fixed calls + attacks × turns × 2`, with Frontier, Balanced, Cheap, and Custom model tiers.
- Document the fixed-call baselines from the implementation: static/replay runs start at 0; dynamic/hybrid runs add resource inference, tool classification when applicable, strategy planning when enabled, and optional report analysis.
- Add visual cookbook cards and inline links for smoke tests, reproducible datasets, domain-specific attacks, report inspection, configuration, and recommendations.
- Use keyword arguments consistently in the red-team and simulation cookbook examples, linked docs snippets, and framework integration examples.
- Shorten report inspection and cost guidance; remove internal accounting, security, and deprecated Streamlit details.
- Clarify CI coverage, simulation failure behavior, dashboard history, and recommendation eligibility.
- Remove stale legacy UI references from the dashboard, CLI reference, and FAQ.

## Linear

- [RES-1353 — Docs: streamline the red-team and agent-simulation guides](https://linear.app/orqai/issue/RES-1353/docs-streamline-the-red-team-and-agent-simulation-guides)
- [DOC-653 — Update screenshots in the docs for red teaming](https://linear.app/orqai/issue/DOC-653/update-screenshots-in-the-docs-for-red-teaming)
- [DOC-655 — Add screenshots to the Agent Simulation Documentation](https://linear.app/orqai/issue/DOC-655/add-screenshots-to-the-agent-simulation-documentation)

## Validation

- `uv run ruff check src`
- `uv run basedpyright`
- `node --check docs/js/cost-calculator.js`
- `git diff --check`
- Live docs server responds with HTTP 200 at http://127.0.0.1:55190/evaluatorq/

The full test suite reached 4,521 passing tests; six tests require optional packages not installed in this workspace (`huggingface_hub`, `pandas`, and the legacy Streamlit UI). The strict docs build still reports the repository’s existing unnav and git-follow timestamp warnings, but completes successfully.


currentlycodinng (approved):
Approving. Docs-only, and the new material is accurate where I could check it against the code, which is most of it. Three things I would tidy, none of them worth holding the merge.

## What I verified against the code

Every flag in the new CLI quickstart exists and means what the surrounding prose says: `--max-static-datapoints`, `--no-executive-summary`, `--no-recommendations`, `--report`, `--yes` and `--category` all come back from `eq redteam run --help`, and `--category` really does take a bare `LLM01`. Same for the defaults quoted elsewhere: `min_evaluation_coverage` is 0.8, `--generated-strategy-count` is 2, `red_team()` saves at `SaveMode.FINAL`, `simulate()` and `generate_and_simulate()` default `save=False`, `upload_results` defaults to `True`, and `eq sim run --no-save` is the right escape hatch for the auto-save under `.evaluatorq/sim-runs/`.

The new simulation note ("dropped simulations raise by default; ordinary failed goals remain in the returned results") is correct against main, where `_simulate_core` already passes `_exit_on_failure=False` down and reserves the raise for `SimulationDroppedError`. It also stays correct after #153 lands, which is the one place I expected these two branches to collide.

Nothing dangles structurally. All ten new dashboard screenshots exist under `docs/assets/dashboard/`, the four `dashboard.md` anchors the guides now point at resolve, and the deleted `_snippets/dashboard-tip.md` has no remaining references in `docs/` or `mkdocs.yml`. The two knobs the trimmed cost section used to carry, `min_evaluation_coverage` tuning and judge `retry_count`, both survive elsewhere: the FAQ keeps the full "how to get warn-only back" answer and `llm-as-a-jury.md` keeps the per-attack versus run-level distinction. Moving `dashboard.md` out of `not_in_nav` and into the nav is the right call now that both guides send readers there.

## `eq redteam ui` and `eq sim ui` still ship

Both commands are still registered and still show up in `eq redteam --help` and `eq sim --help` with their own one-line descriptions, but this PR removes the last mention of them from the docs, including the deprecation notes. So the CLI advertises a command that the CLI reference no longer lists at all, and a reader who copies it out of `--help` and searches the docs learns nothing, not even that they should be using `eq dashboard`. Dropping the flag tables is right; I would keep a single row in the `eq redteam` / `eq sim` command tables saying deprecated, use `eq dashboard`, until the commands themselves come out.

## The `audited` tri-state loses its explanation but keeps its consequences

The simulation guide drops the paragraph explaining that `audited` is `None` on runs saved before the field existed, and with it the code comment justifying `.get('audited') is False` over `not c['audited']`. The example still uses the careful form, so the code is fine, but the reason a reader would not simplify it is gone. Two sentences later the page still says `evidence` "is `''` when the criterion never occurred and `None` when no tracker was available, same as `audited`", which now points at a third value the page no longer describes. Either restore one sentence naming the three states or drop the "same as `audited`" comparison so nothing refers to it.

## The focus-areas screenshot contradicts the section it illustrates

The new `docs/assets/dashboard/README.md` is a genuinely useful addition, and it says plainly that `redteam-05-focus-areas.png` predates the apply flow and shows no **Apply…** buttons, with the instruction not to delete the prose caveats standing in for the missing shots. The simulation guide honours that, calling out the empty **TRACES** column and steering readers on turn quality and compare. Red teaming section 5 does not: it describes the Apply button, the diff preview and the confirm step over a screenshot that has none of them, with nothing telling the reader the shot is older than the feature. One sentence there would close the gap the README asks you to keep open. `sim-09-compare.png` is a milder version of the same, since its red "low overlap (0%)" banner is an artefact of the capture run and the section text does not say so.

## Smaller

The cost section is now a paragraph, which is the right size for a first-run guide, but `EvaluatorConfig(api='chat_completions')` goes with it and is not documented in any guide afterwards. The new text explains the symptom, that a cost with `priced_calls < calls` is a lower bound, without the lever that changes it. Worth a line in the FAQ or `llm-as-a-jury.md` rather than restoring the block here.

The FAQ's coverage-gate answer still ends with "see Red Teaming › In CI" for the `min_successful_judges` distinction, and that is the part the In CI section no longer explains. The FAQ carries the explanation itself, so the pointer is just redundant now, but it reads as if there is more to find.

Baukebrenninkmeijer:
Thanks — all four picked up in `9aa9a7c8`, plus two of your points went stale in the meantime.

**`eq redteam ui` / `eq sim ui`** — right, `--help` was advertising a command the reference no longer listed at all. Both CLI reference pages get a short deprecated section pointing at `eq dashboard`, rather than restoring the flag tables.

**The `audited` tri-state** — restored, as one sentence naming all three states (`True` / `False` / `None` on runs saved before the field existed) and why the example tests `is False` rather than `not c['audited']`. That also repairs the dangling "same as `audited`" in the `evidence` line.

**`EvaluatorConfig(api='chat_completions')`** — now a FAQ entry rather than back in the guide, as you suggested. Worth flagging that I first wrote it backwards: the default is `responses` *because* that is the endpoint the router prices, so the lever is not "switch to chat completions to get prices" — the chat-completions fallback and the explicit opt-out are both what leave judge calls unpriced. The entry says that.

**The FAQ pointer to "Red Teaming › In CI"** — dropped; the FAQ carries the `min_successful_judges` explanation itself.

**Now stale, both fixed before your review landed:**

- The focus-areas screenshot no longer contradicts its section — `redteam-05-focus-areas.png` was removed outright, along with `sim-05-transcripts.png`, `sim-07-turn-quality.png` and `sim-09-compare.png`, all four for the reason your quote from `docs/assets/dashboard/README.md` gives. Prose and caveats stayed; the images are still on disk and the README now says they are unreferenced and must be re-added in the same commit that replaces them. So `sim-09-compare`'s red "low overlap" banner is gone with it.
- The dashboard walkthroughs then moved wholesale out of both guides into `docs/dashboard.md` as the canonical copy, with a condensed tab table left behind in each guide. Section numbering you reviewed against has shifted accordingly.

Also worth a look since it changed after your review: the cost calculator was rebuilt on real per-1k token prices, and a follow-up adversarial pass then found three modelling bugs in that rebuild — judge calls were growing the transcript for each other (a five-judge jury compounded), judge output was priced as a full 1k-token turn, and fixed setup calls were getting the 90% cache discount despite sharing no prefix. Fixed in `74928c7f`; the default estimate moved $0.4842 → $0.3642.

PR #155, opened 2026-08-17
feat(cache): keep the Anthropic prompt cache across a replayed conversation

Baukebrenninkmeijer:
Linear: [RES-1352](https://linear.app/orqai/issue/RES-1352/retain-the-anthropic-prompt-cache-across-replayed-conversations-agent)

## What

Conversation growth in agent simulation and red teaming was already append-only — no trimming, no summarization, no sliding window. But Anthropic caching is **opt-in**: without a `cache_control` breakpoint the whole transcript is re-encoded every turn however stable the prefix is.

Evidence before the change: **715 `claude-sonnet-4-6` router traces** over the last 30 days, effectively zero cache reads. 740 of the sampled traces had prompts ≥1024 tokens, so it was never a size problem.

## How

New `common/prompt_cache.py`. Both APIs take a **positioned, per-item** marker, so the same placement rule applies to each:

| Path | Helper | Wire shape |
|---|---|---|
| Chat Completions | `apply_cache_breakpoints(messages, *, volatile_tail)` | system + end-of-prefix become text blocks with `cache_control: {"type":"ephemeral"}` — 2 of the 4 allowed breakpoints |
| Responses | `mark_responses_input(input, *, volatile_tail)` | `cache_control` on the last content part of the end-of-prefix `input` item |

Applied in `simulation/agents/base.py` (both API paths) and the red-team `OpenAIModelTarget`. Both are gated on `client_routes_through_orq`: `cache_control` inside a content part is not in the direct OpenAI schema (OpenAI's own schema names `prompt_cache_breakpoint`), so a client pointed at `api.openai.com` or a self-hosted server is left untouched.

**No `prompt_cache_key`.** OpenAI caches automatically and the router *ignores* `cache_control` on non-Anthropic models rather than rejecting it, so there is no per-provider branch to maintain.

### `volatile_tail` — a required keyword, no default

A cache **write** costs 1.25x and is read back only by a later request repeating the marked prefix byte-for-byte. Marking a message the caller *rebuilds* each turn is therefore a pure loss: the next turn puts different content at that position, the prefix diverges immediately after the system message, and the whole transcript pays a write nothing ever reads.

`volatile_tail` is how a caller declares how many trailing messages it rebuilds (`0` when the whole list persists). It has **no default** on purpose — a caller that rebuilds its last message and forgets to say so gets a silent per-turn write and no read, which surfaces as a bill rather than a bug.

`JudgeAgent.evaluate` appends a synthetic per-turn instruction, so it passes `volatile_tail=1`. Without that, its breakpoint landed on the synthetic turn and the cached figure stayed pinned at the system-prompt size while the transcript grew — the worst case, since the judge runs every turn on the longest input.

### The Responses path is positioned, not top-level

An earlier revision sent a **top-level** `cache_control` body field on the Responses path. That marks the end of the *whole* input, including the trailing item the judge rebuilds, so it wrote on every call and read on none. `input` is a list of items with content parts exactly like a chat message list, and the router honours a per-item marker — nothing about the shape forced the limitation. Probed directly:

```
top-level only (the earlier revision)   reads=[0, 0]     no read
per-item on prefix end                  reads=[0, 4764]  HIT
per-item + top-level                    reads=[0, 4761]  HIT
neither (control)                       reads=[0, 0]     no read
```

`responses_cache_body()` is gone; the judge's "Responses cannot cache, switch to `api='chat_completions'`" warning is deleted with it.

### Judge system prompt is now byte-stable

The `[ALREADY CONFIRMED]` markers for settled criteria moved out of the criteria listing and into the trailing user message `JudgeAgent.evaluate()` appends. They sat at token position 0, so every `mark_settled()` call invalidated the whole cached prefix — system prompt, transcript and tool schemas — to save a few dozen output tokens. Behaviour unchanged; a custom judge is unaffected.

`tracing._serialize_messages` now flattens text blocks, and `with_llm_span` was deduplicated onto it — it carried a second inlined copy of the same extraction, so a marked message reached the red-team target span as a Python repr with `cache_control` in it.

## Verification

Live against `anthropic/claude-sonnet-4-6` on the Orq router.

**Both judge API paths, three judgements each, uuid-salted cold prefix** (`scripts/manual_tests/prompt_cache_judge_check.py` — the salt matters, without it a second run inside the 5m TTL reads the first run's entry and every number lies):

```
                   call 1   call 2   call 3
chat_completions        0    6,998    7,890
responses               0    7,424    8,313
```

`cached` tracks the transcript rather than standing still. Responses reads *more* because its system prompt lives in `instructions`, outside the marked input, so the whole input prefix caches.

Full `SimulationRunner` run with target, user simulator and judge **all** on Anthropic:
```
run total: {'calls': 3, 'input_tokens': 7009, 'cached_tokens': 6150, 'output_tokens': 362}
```

Local checks on the merged tree: `ruff check src`, `ruff format --check src`, `basedpyright` (whole repo, 0 errors), `pytest -m 'not integration'` (4805 passed).

## Notes for the reviewer

- **`cache_creation_tokens` stays 0.** The router's OpenAI-shaped `usage` reports reads but never the write, so the reports' "Cache-Write Tokens" row will be permanently empty for router Anthropic traffic. Our parser handles the field if it ever appears — do not read the zero as "no cache write happened".
- **No minimum-size guard.** Breakpoints are placed per request regardless of prompt size, so a short or one-shot exchange can pay a 1.25x write it never reads back. Anthropic caches nothing below ~1024 tokens. Both helpers document this; neither enforces it.
- **Never set `ttl`.** The 5m default is right; `1h` costs more and only Anthropic honours it.
- **Coverage is partial and deliberately so.** The adaptive red-team orchestrator (whose attacker transcript is itself append-only and grows every turn), `redteam/runtime/jobs.py` and `common/judge.py` replay conversations through `execute_chat_completion` and get no breakpoints. The default red-team target backend is `orq`, not `openai`, so it is untouched too. Centralizing the gate in `execute_chat_completion` is [RES-1360](https://linear.app/orqai/issue/RES-1360/centralize-prompt-cache-breakpoints-in-execute-chat-completion).
- **Orq's documented minimums are wrong for Sonnet 4.6.** Traces show caching engaging at 1319 tokens, below Orq's published 2048 floor; Anthropic's own table says 1024. Docs fix tracked separately.
- Marked content is sent as a **text-block list** rather than a plain string. A target that inspects the raw request body will see that shape; three existing red-team tests were updated to flatten it.


currentlycodinng (changes requested):
Requesting changes on two things, one of which is the judge change itself. The mechanism is right and I reproduced your numbers, but the breakpoint placement on the judge path caches the system prompt and nothing else, which is the opposite of what the PR description claims for it.

## The judge's breakpoint lands on a message that never becomes a prefix

`evaluate()` appends a synthetic trailing user message, `apply_cache_breakpoints` marks the last message, so the second breakpoint always lands on that synthetic turn. That turn is thrown away and rebuilt on the next judgement, so the cache entry written at it is never read back, and no entry is ever written at the end of the transcript. The only reusable breakpoint left is the one on the system prompt.

Live on `anthropic/claude-sonnet-4-6` through the router, judge shape exactly as this branch builds it, transcript growing:

```
judge  transcript=2 turns      in=3811  cached=0      (0%)
judge  transcript=4 turns      in=4007  cached=3603   (90%)
judge  transcript=6 turns      in=4203  cached=3603   (86%)
judge  transcript=8 turns      in=4399  cached=3603   (82%)
judge  transcript=10 turns     in=4595  cached=3603   (78%)
judge  transcript=12 turns     in=4791  cached=3603   (75%)
```

`cached` is pinned at 3603, which is the system prompt, and never moves. Every token of transcript is re-encoded on every judgement, so the cached share decays as the conversation grows and keeps decaying.

The simulation target path does not have this problem, because there the marked last message is a real user turn that stays in history and so is a genuine prefix next turn:

```
target transcript=3 turns      in=3851  cached=3750   (97%)
target transcript=5 turns      in=4047  cached=3946   (98%)
target transcript=7 turns      in=4243  cached=4142   (98%)
```

Moving the second breakpoint to the last real transcript message and leaving the synthetic trailing turn unmarked fixes it. Same model, same shape, only the placement changed:

```
variant B t1 (2 turns)                 in=5011  cached=0
variant B t2 (3 turns)                 in=5109  cached=4999
variant B t3 (4 turns, settled a)      in=5216  cached=5097
variant B t4 (5 turns, settled a,b)    in=5318  cached=5195
```

`cached` now tracks the transcript instead of standing still, at 97 to 98% throughout, and the settled note changing between turns costs nothing because it sits past the breakpoint. That also makes the `_settled_note` move pay off the way the description says it does: right now the note is already past the last cacheable boundary, so byte-stabilising the system prompt buys only the 3603 tokens above and not the transcript behind it.

Your 88% figure for the full run is real but it averages three callers, and the judge is the one called on every turn with the longest input, so its shortfall is the one that compounds.

## The span still gets the `cache_control` repr on the red-team path

`_serialize_messages` is fixed, but `with_llm_span` has a second, inlined copy of the same role/content extraction at [tracing.py:479-488](src/evaluatorq/common/tracing.py#L479-L488) that still calls `truncate_for_span` directly. `OpenAIModelTarget` passes the marked list into it as `input_messages` ([openai.py:150](src/evaluatorq/redteam/backends/openai.py#L150), via [redteam/tracing.py:184](src/evaluatorq/redteam/tracing.py#L184)), so the exact noise the tracing fix exists to prevent still reaches the target span. Real exported span, in-memory exporter, marked messages:

```
span: chat anthropic/claude-sonnet-4-6
  gen_ai.input.messages = [{"role": "system", "content":
    "[{'type': 'text', 'text': 'you are a target', 'cache_control': {'type': 'ephemeral'}}]"}, ...]
```

Simulation is unaffected because it records input through `record_llm_input`, which goes through the fixed path. One line: call `_span_content_text` at 483 too.

## One question before merge

The Responses path gates `responses_cache_body()` on `client_routes_through_orq`, and the Chat Completions path applies the breakpoints unconditionally, including for an injected non-router client. `client_routes_through_orq` exists precisely so an injected OpenAI client does not receive router-specific fields, and its docstring says so. OpenAI's own schema for a text content part carries `prompt_cache_breakpoint`, not `cache_control` ([openai-openapi, ChatCompletionRequestMessageContentPartText](https://github.com/openai/openai-openapi/blob/master/openapi.yaml)), so a direct `AsyncOpenAI()` on the simulation or red-team path is sending a field OpenAI does not define. I have no OpenAI key here so I could not test whether it is ignored or rejected. If you verified it, say so in the description, and if not, gating the Chat Completions call the same way costs nothing.

Nit: `apply_cache_breakpoints([])` raises `IndexError`, because index 0 survives the `index < 0` guard on an empty list. Not reachable from either call site since both prepend a system message, so it is only worth the guard if you expect the helper to be used elsewhere.

## What I verified and what held

Your core claim reproduces exactly. Same filler prompt, marked versus unmarked, through the router:

```
anthropic/claude-sonnet-4-6   marked t1  in=5521 cached=0
anthropic/claude-sonnet-4-6   marked t2  in=5521 cached=5518
anthropic/claude-sonnet-4-6   plain      in=5521 cached=0
```

Caching really is opt-in, and the unmarked request gets nothing however stable the prefix.

The thing I most expected to bite does not. When the transcript grows, the previously marked message reverts to a plain string, which changes the wire shape of the cached prefix, and I assumed that would miss. It does not: the grown request still read 5518 back, so a single text block and a bare string normalise to the same prefix.

`openai/gpt-5.4-mini` accepts the marker and caches on its own, as you say. `google/gemini-2.5-flash` accepts it too and actually does better with it than without, 5013 cached against 4065 on the identical unmarked request, which is worth a line in the CHANGELOG since it currently reads as though the marker is inert on Gemini. One Gemini call returned a 500 empty response on the grown shape, but it did not reproduce over three repeats and the plain request has the same flake, so it is not this change.

`uv run pytest -m 'not integration'` gives 4559 passed, 2 skipped, matching your number. `ruff check src` and `ruff format --check src` are clean.

The `_MARKABLE_ROLES` guard, the non-mutating copy, and the decision not to set `prompt_cache_key` all look right to me, and leaving tool-calling assistant turns alone is the correct call.

currentlycodinng (approved):
Re-reviewed 8d501bdf...ce05d07683b8. Both P0s from my last pass are fixed, verified by reading the actual call paths rather than trusting the commit messages.

**Finding #1 (judge breakpoint on the synthetic turn) — fixed.** `JudgeAgent.evaluate()` (`simulation/agents/judge.py:733-748`) now passes `volatile_tail=1` through `_call_llm`, and `apply_cache_breakpoints`/`mark_responses_input` (`common/prompt_cache.py`) exclude exactly the last `volatile_tail` items when picking the prefix-end breakpoint. The breakpoint now lands on the last real transcript message, not the rebuilt instruction turn, on both the chat and Responses paths. `tests/redteam/test_orchestrator_prompt_cache.py::test_marker_follows_the_growing_transcript` pins the growing-prefix behavior, which would catch a regression to a hoisted/frozen marker.

**Finding #2 (cache_control repr on spans) — fixed.** `with_llm_span` in `common/tracing.py` no longer has its own inlined role/content extraction; it now calls `_serialize_messages`, which routes through `_span_content_text` -> `coerce_content_text` -> `truncate_for_span`, same as the already-fixed `_serialize_messages` path. `redteam/tracing.py`'s `with_llm_span` was already a thin delegate to this function, so there was only one copy of the bug and it's gone.

**Open question (Chat Completions gating) — resolved.** Both `base.py`'s chat path and `openai.py`'s `OpenAIModelTarget._call` now gate `apply_cache_breakpoints` behind `caching_applies(client, model)`, matching the Responses path. An injected non-router client no longer gets cache_control blocks.

**Nit (empty-list IndexError) — resolved.** `_prefix_index` on an empty list returns -1 without indexing; `apply_cache_breakpoints([])` returns `[]`.

New work since my review: the adversarial-loop caching in `redteam/adaptive/orchestrator.py` recomputes the breakpoint every turn against the live transcript with `volatile_tail=0`, correct since the attacker loop is pure append, unlike the judge's rebuilt trailing turn, and it's guarded by the same growing-transcript test against the hoisting failure mode. The cached-token report rows (`fmt_cached_tokens`, `CACHED_TOKENS_LABEL`) are shared between the redteam and simulation renderers and guard the zero-denominator case. No new P0 found.

Two non-blocking notes, not gating this approval: no test/lint CI has run against this head SHA, only the `label` check is green, so I'm verifying from code paths and their tests rather than a green suite. Separately, the branch currently shows as CONFLICTING with main, so it needs a rebase before it's mergeable regardless of review state.

Approving to clear my stale change-request.

PR #156, opened 2026-08-18
style(docs): make both sidebars read as a hierarchy (RES-1368)

Baukebrenninkmeijer:
Closes RES-1368.

## Problem

Neither sidebar read as a hierarchy, and the two disagreed about how to show one.

- **TOC (right):** Material indents levels with cumulative link padding only, on a 12.1rem rail. On a 4-level page most entries wrapped and every level rendered at the same size and weight.
- **Nav (left):** section labels differed from the pages under them by font weight alone, at the same indent — a group read as a flat list. And it ran the *opposite* convention to the TOC: small grey labels above full-strength page links, where the TOC had dark parents above lighter children.

## Change

All in `docs/stylesheets/extra.css`, scoped to `min-width: 76.25em` for the nav (below that the same markup is the mobile drawer, where Material's own spacing applies).

One convention in both rails: **parent larger and darker, children indented and lighter.**

- Secondary sidebar 12.1rem → 14.5rem, so TOC entries mostly fit one line.
- TOC keeps a guide rule per nesting level — its levels differ by ~1px of type size and a heading can sit 4 deep, so the rules do real work. The nav has two levels and a size jump, so it gets indent and whitespace instead.
- Nav section labels at 1.08rem/600; pages 0.94rem, muted, indented 0.6rem. Groups get equal margin top and bottom.
- Group headings use a new `--eq-nav-heading` (70% of the theme foreground) — dark grey rather than near-black, and since it mixes `--md-default-fg-color`, slate inherits the same relationship without a second rule.
- Active item: accent colour + weight 600 in both rails.

`mkdocs.yml`: Agent Simulation and LLM as a Jury now take the section shape Red Teaming already had — landing page becomes "Overview", and Cyclic Judge Assignment + Pairwise Judging move under the jury group. Agent Simulation is deliberately a one-page group, as a placeholder for pages to come.

`CLAUDE.md`: two traps that cost time here — live-reload does not fire in a Conductor worktree (the server keeps serving the build it started with, which reads like a browser cache), and a bare `mkdocs serve` can bind a port a sibling worktree already owns, so you end up reviewing their build. Plus: never `pkill -f` a command name, it kills sibling sessions' servers.

## Verification

`mkdocs build --strict` green. Every state was checked against a rendered screenshot at 1600px, light scheme; dark mode follows because the colours are theme-variable mixes rather than literals.


Baukebrenninkmeijer:
## Before / after

Both builds rendered at 1600px, light scheme. `BEFORE` = `origin/main`, `AFTER` = this branch.

### Left sidebar — nav hierarchy
Group labels get size and weight, pages indent and lighten. Agent Simulation and LLM as a Jury are now groups; their landing pages are named Overview.

left sidebar before/after

### Right sidebar (TOC)
Wider rail, uppercase title, h2s carry weight, a guide rule at every depth, accent rule on the active entry.

toc before/after

### Reference tables
Tables were pinned to the body font size, so a three-column table filled the screen and code spans broke mid-token. Also fixes two cells where an empty double-backtick span rendered its backticks literally.

table before/after

<sub>Images live on the `screenshots/res-1368` branch (nothing added to this PR's diff); delete it after merge.</sub>

PR #157, opened 2026-08-20
feat(simulation): run the whole simulation pipeline on the Responses API

Baukebrenninkmeijer:
## What

Every simulation LLM call now goes through the Responses API. Previously the judge and target used Responses while the persona/scenario/first-message generators and the user simulator used chat completions, so a run's spans came back as a mix of `span.chat_completion` and `span.generic` — scanning a trace for router spans made the simulator and generator calls look absent.

Verified on staging (412 project): a full run's trace contains **zero** `chat ` spans.

```
Evaluatorq - Agent Simulation
  orq.simulation.persona_generation       → responses openai/gpt-5.4-mini
  orq.simulation.scenario_generation      → responses openai/gpt-5.4-mini
  orq.simulation.first_message_generation → responses openai/gpt-5.4-mini
  orq.simulation.turn
    orq.simulation.user_simulator_call    → responses openai/gpt-5.4-mini
    orq.simulation.target_call            → responses agent/<key>
    orq.simulation.judge_evaluation       → responses openai/gpt-5.4-mini
```

- `generate_structured` gains `api=`. The Responses branch uses `responses.parse(text_format=...)` and returns `None` — after a warning naming the cause — on 400/404 or an unparsed result, so a provider without Responses structured output degrades to the existing chat legs rather than losing the call. Truncation raises the same actionable error as the chat leg. **The default is unchanged**, so the red-team report paths (`redteam/reports/recommendations.py`, `simulation/reports/recommendations.py`, `simulation/traces.py`) keep chat completions; flipping them is now a one-word change per call site.
- Persona/scenario generators pass `api='responses'`; the first-message generator moves off a direct `chat.completions.create`; the user simulator defaults to `responses` as `JudgeAgentConfig` already did.

## Three defects fixed along the way

Found while getting a 30-scenario run to complete — each one blocked the run.

**1. The user simulator was fed the target's tool traffic.** It received the assistant turn's `tool_calls` plus the `role='tool'` result rows. After role inversion that becomes a `user` message carrying `tool_calls` followed by orphan tool rows, which the provider rejects:

> `Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'`

Every simulation against a tool-using agent died at turn 1. The canonical transcript is untouched (inversion builds a throwaway list), so the target and judge still see the tool traffic — `test_does_not_mutate_the_canonical_transcript` pins that.

**2. `ScenarioGenerator` had a flat `max_tokens=6000`.** At ~15+ scenarios the structured output truncated and raised an unrecoverable `LengthFinishReasonError`. The budget now scales with `num_scenarios`.

**3. The simulator call closed the wrong turn.** A turn span read `target → judge → next user message`. It now opens the following turn: `user → target → judge`.

## Tests

Three new regression tests (`test_turn_boundaries.py`, `test_scenario_token_budget.py`, plus the transcript ones). The rest of the diff is test doubles reshaped from chat completions to Responses (`choices[0].message.content` → `output_text`, `response_format` → `text_format`).

`ruff check` / `ruff format --check` clean, `basedpyright` 0 errors, **4780 passed**.

---

## Review follow-up — [RES-1400](https://linear.app/orqai/issue/RES-1400)

Second commit. A parallel review (7 lenses, plus 5 adversarial critics — one from a different model family) found the migration shipped with a fallback that laundered unrelated errors, a truncation guard that could not fire, the token-budget fix applied to 2 of 7 call sites, and two hand-rolled copies of transport that already exists. All fixed here; tracked in **RES-1400**.

**The Responses fallback caught every 400 and 404.** It logged `Responses structured output rejected` and re-billed the same request on chat completions — so a bad parameter, an over-length context or a content-policy rejection came out blaming the provider, at double the cost, with the real error only surfacing from the wrong endpoint. A 400 is now gated on the error body, reusing the keyword test the chat leg already had; the test is extracted into one predicate so the two legs cannot drift again. 404 stays unconditional, since no endpoint *is* a capability signal.

**The truncation guard could not fire for the case it was written for.** `responses.parse` runs `model_validate_json` on the body before returning, so a payload cut off at `max_output_tokens` raises `ValidationError` from inside the SDK and never reaches the `incomplete_details` check — confirmed against the SDK's own `parse_text` → `model_parse_json`. Only a response truncated with *no* text at all (a reasoning model that spends the budget before answering) came back as an object. Since the generators catch `json.JSONDecodeError` only, a truncated run died on a raw pydantic error reading "malformed JSON" instead of "your budget was too small". Both shapes now raise one actionable error, which also collapses the third copy of that message in the file.

**The reserved-key guard knew only chat vocabulary.** `_STRUCTURAL_KEYS` reserves `model`/`messages`/`response_format`, but the Responses leg's structural fields are `input`/`text_format`/`max_output_tokens` and it ended with a bare `params.update(extra_kwargs)`. `extra_kwargs={'text_format': OtherModel}` passed the guard, got a well-formed `output_parsed` of the wrong type, and returned it through the `cast` — no error until an attribute access failed frames away. Reserved keys are now per-API, and an `api='responses'` call reserves *both* sets, because it may still fall through to the chat legs.

**The scaling budget reached 2 of 7 call sites.** `generate_edge_cases`, `generate_boundary_scenarios` (flat 4000 — *below* the new 6000 floor), `generate_security_scenarios` and both `PersonaGenerator` methods kept flat caps, so `generate_security_scenarios(num_scenarios=30)` reproduced the exact truncation defect #2 above set out to fix. The arithmetic moved to `common.structured_output.token_budget_for_items`; each generator binds its own per-item cost. Deliberately unbounded at the top — a clamp trades an actionable "raise the budget" error for silent truncation at an arbitrary ceiling.

**Both new Responses call sites bypassed `common.llm_call.execute_response`.** They rebuilt params, trace headers and pipeline metadata by hand, dropping the `llm_slot()` concurrency bound (so generation escaped `llm_parallelism` entirely), the `asyncio.wait_for` timeout, the reasoning-block drop-and-retry memo, and `price_usage` — the last of which `structured_output.py`'s own RES-1295 docstring already complained was missing. Both now route through it, which also binds the generic properly instead of asserting it with a `cast`.

### Tests

Every Responses branch is now covered: happy path with no chat call, 404 and schema-400 degradation with their warnings asserted, the unrelated-400 re-raise, unparsed output, both truncation shapes, and both reserved-key sets. `test_scenario_token_budget.py` drives all five scenario methods and both persona methods through a capturing client and asserts the real `max_output_tokens` — the previous version proved the helper, not the wiring, so reverting a call site to a flat literal left it green.

Docs ship with it: `docs/tracing.md` still showed the turn children in the pre-fix order, and `CHANGELOG.md` gains the two `### Notable defaults` this PR changes (the `api='responses'` default plus the tool-row drop; the budget scaling and its spend increase).

`ruff check src`, `ruff format --check src`, `basedpyright` (whole repo, 0 errors), `pytest -m 'not integration'` (**4797 passed**) and `mkdocs build --strict` are all green.

### Known gaps, deliberately not taken here

- **First-message generation still re-raises a 400/404 from `/responses`.** On a provider without the endpoint that aborts generation for the datapoint. `generate_and_simulate` gathers with `return_exceptions=True` and skips the pair with a warning, and the trace-derived path catches and replays the recorded opening — but `DatapointGenerator.generate_datapoints` gathers *without* it, so one failure aborts the whole batch before any simulation runs. The narrow fix is a 404-only degrade to chat completions.
- **A target turn that is purely tool calls with no text** inverts to a content-less `user` row for the simulator. It no longer crashes, but the simulator is asked to reply to a blank turn. Rendering the tool traffic into the inverted row's text (via `contracts.tool_result_to_text`) is the structural fix and makes the `function_call`/`function_call_output` pairing problem impossible by construction.


Baukebrenninkmeijer:
## Outcome

Implemented and pushed in merge commit `7b7a55f8` after merging `origin/main`.

- Responses handling now checks provider stop/incomplete metadata before accepting partial output, so max-output-token failures surface an actionable error without relying on EOF indicators.
- Refusals are propagated from Responses/OpenAI-compatible payloads through `AgentResponse` and `LLMResult` into the judge paths.
- Structured Responses fallback is limited to capability signals (404, schema-unsupported 400, or unparsed output); unrelated 400s are preserved, and Responses calls use the shared transport helper.
- Regression coverage includes truncation, refusals, fallback behavior, tool-call transcripts, and shared defaults.
- Verification passed: ruff check, format check, basedpyright with 0 diagnostics, and `pytest -m 'not integration'` with 4,823 passed and 2 skipped. Live research-profile calls also verified normal parsed Responses output and real truncation handling.

Known deferred items remain documented: first-message generation still lacks a 404 fallback, and tool-only target turns still produce a blank inverted user row.

PR #162, opened 2026-08-22
feat(config): expose internal configuration on every execution path

Baukebrenninkmeijer:
## What

An audit of `evaluatorq` for *"internal configuration items not exposed to users on one of the paths"* turned up ten classes of defect: config fields accepted by a pydantic model and then ignored at the call site, knobs honoured on one execution path and dropped on another, hardcoded module constants with no override, three drifted copies of the reserved-request-key guard, and `reasoning_effort` reaching exactly one of many call sites. This PR fixes all ten.

## Changes

**Config plumbing**
- `EvaluatorConfig` inherits `LLMCallConfig` rather than redeclaring its fields. An import-time guard fails the package load if the field sets ever desync — `extra='forbid'` previously rejected any newly added `LLMCallConfig` field and broke the documented `LLMConfig(evaluator=LLMCallConfig(...))` pattern.
- `check_reserved_keys` + `_RESERVED_COMPLETION_KEYS` / `_RESERVED_RESPONSES_KEYS` in `contracts.py` replace three drifted copies of the same guard.
- `LLMCallConfig.responses_params()` joins `completion_params()`, so a config renders correctly for both endpoint shapes (`max_output_tokens` vs `max_completion_tokens`, `reasoning={'effort': ...}` vs flat `reasoning_effort`).

**`reasoning_effort`**
- Reaches every call site: `execute_chat_completion`, `execute_chat_parse`, `execute_response`, `OrqResponsesTarget`, `BaseAgent`, the OpenAI backend.
- Settable per role — target, judge, attacker, simulator — via each role's own `LLMCallConfig`.
- The preflight is scoped to targets that actually receive the setting. Other target kinds warn and skip instead of failing the run with `ValueError`.
- Every path drops the reasoning block and retries once when a model rejects it, memoized per process.

**`extra_body`**
- New first-class `LLMCallConfig` field, merged **over** the call-site router body per key, so a caller-supplied `thread`/`retry` hint no longer wholesale-replaces the call site's. Still rejected inside `extra_kwargs` — it is a dedicated parameter on all three executors.

**Execution-path parity**
- Red team's static leg honours `target_agent_timeout_ms`, `max_target_retries` and the same error taxonomy as the hybrid leg.
- Simulation's `BaseAgent` honours the whole `LLMCallConfig` it accepts (temperature, max tokens, timeout, reasoning effort, `extra_kwargs`, `extra_body`) on both the chat and Responses paths, with env fallbacks resolved at call time instead of import time.
- `ORQAgentTarget` takes `max_tool_continuations` at construction instead of reading a module global.

**Token usage**
- `generate_structured` returns a frozen `StructuredResult(parsed, raw, usage)` and sums usage across all four fallback rungs. 11 call sites migrated. *Not marked breaking* — the helper is internal, but callers outside the package that unpacked the old 2-tuple would need updating.

**Exposed constants**
- `SimulationScoringConfig` (turn-efficiency cliffs, decay, floor, composite weights) with validated invariants. **Behaviour change:** the tail decay now continues from the last cliff, so a longer conversation no longer scores *higher* than a shorter one (7 turns scored 0.9 while 6 turns scored 0.7).
- Backoff parameters on `common/retry.py`, budget parameters on `common/apply.py`, the executive-summary model knobs, and a user-writable model-catalogue override registry (`register_model`) plus retry on a failed catalogue fetch.

**Docs**
- New `docs/tuning.md` (every knob, what it means, per-role examples) and `docs/guides/targets.md` (writing your own `AgentTarget`, including an OpenAI-backed one).
- `docs/configuration.md`, both CLI references, the simulation and red-teaming guides, `CHANGELOG.md` and `.claude/skills/docs-coverage/axes.md` updated.

## Verification

```
uv run ruff check src                  # clean
uv run ruff format --check src         # clean
uv run basedpyright                    # 0 errors, whole repo
uv run pytest -m 'not integration'     # 4931 passed
uv run python scripts/validate_mermaid.py
```


Baukebrenninkmeijer:
### `extra_body` on the jury helpers — before / after

`llm_jury()`, `llm_jury_pairwise()` and `PairwiseComparator` build their `LLMCallConfig` internally, so body fields had no seam on the jury path at all.

**Before** — no `extra_body=` parameter. The obvious route silently misbehaved:

```python
from evaluatorq import llm_jury

# extra_body is one of the structural fields the call site owns, so this raises
# inside the param builder — but run_judge catches a judge failure and turns it
# into a verdict, so the symptom is a judge that always fails, not a crash.
jury = llm_jury(
    name='helpfulness',
    criteria='Is the answer helpful?',
    judges=['openai/gpt-5.6-luna'],
    extra_kwargs={'extra_body': {'my_router_field': 'x'}},
)
```

The only real workaround was a hand-rolled client proxy — and it had to override `with_options`, because `without_client_retries` clones the client via `client.with_options(max_retries=0)` and a plain `__getattr__` proxy is silently unwrapped there, dropping the injection with no error:

```python
class _BodyInjector:
    def __init__(self, inner, body): self._inner, self._body = inner, body
    def __getattr__(self, n): return getattr(self._inner, n)
    def with_options(self, **kw):
        return _BodyInjector(self._inner.with_options(**kw), self._body)
    # ...plus proxying .responses / .chat.completions to merge self._body
```

**After**:

```python
jury = llm_jury(
    name='helpfulness',
    criteria='Is the answer helpful?',
    judges=['openai/gpt-5.6-luna'],
    extra_kwargs={'top_p': 0.9},          # top-level SDK arg, replaces the key
    extra_body={'my_router_field': 'x'},  # request body field, merged per key
)
```

Same on the pairwise side:

```python
llm_jury_pairwise(name='ab', criteria='Which is better?',
                  judges=[...], extra_body={'my_router_field': 'x'})

PairwiseComparator(model='openai/gpt-5.6-luna',
                   extra_body={'my_router_field': 'x'})
```

The two seams are not interchangeable, and the merge semantics are the reason: `extra_kwargs` sets a top-level argument on the SDK call and **replaces** the key; `extra_body` is **merged** per key by `LLMCallConfig`, so router-owned body fields (`thread`, `memory.entity_id`, the retry hints) survive alongside the caller's.

Shipped in `8414cf99` — 22 lines of source threaded through the single `_run_single_judge` path so pointwise and pairwise can't drift. Tests drive the real scorer and `PairwiseComparator.compare` rather than `run_judge` directly; each of the three forwarding sites was mutation-checked (deleting any one fails a test).

PR #165, opened 2026-08-23
docs: explain red-team replay [axes]

Baukebrenninkmeijer:
`gap: entry point (red_team()) × data source (replay)`

Red-team replay had no prose anywhere on the docs site. The Modes section of the red-teaming guide listed dynamic / static / hybrid and stopped, `previous_run=` appeared in zero docs pages, and `--from-run` existed only as a flag-table row in the CLI reference. This adds a `## Replay a previous run` section to `docs/guides/red-teaming.md`.

## What the page covers

- what replay is for: it holds the attack set fixed, so a moved resistance rate means the agent moved, which no fresh dynamic or hybrid run can tell you;
- what is replayable, and the ways a saved run is not (`save='none'`, a `--save detail` artifacts directory, a pre-replay-support or newer-format run);
- creating a run to replay, then replaying it, with all five `previous_run=` reference forms;
- what replay restores (pipeline, `max_turns`, `attacker_instructions`) and that an explicit argument still wins;
- the nine data-selection arguments replay rejects, and that they are rejected by name rather than by value;
- `### Use replay in CI` — the run store does not survive a fresh checkout, so a build must persist it itself. Includes a worked replay-plus-rate-gate snippet.

Replay is not added as a fourth `## Modes` bullet. `.claude/skills/docs-coverage/axes.md` says explicitly that replay crosses with the modes rather than being one, so it is a trailing sentence after the list.

There is no fenced `eq redteam run --from-run` block. No `EVALUATORQ_AGENT_KEY` was available and `--target` accepts only `agent:`/`deployment:` — so it could not be executed. The CLI form is prose plus a cross-link to the flag table instead.

## Review pass

The branch was rebased onto main and reviewed across four lenses: quality, docs accuracy, minimality, and spec alignment. One Critical and four Important findings were confirmed against source and fixed in `326c7bde`:

| finding | fix |
|---|---|
| The `mode` admonition documented a bug main had already fixed (`17be8978` gave `red_team()` and `--mode` a `None` default), and contradicted the rejected-set sentence above it | deleted |
| `save='none'` writes no manifest either, so it is not listed by `eq redteam runs` at all | claim corrected |
| `EVALUATORQ_DIR` names the store root, not the `runs/` directory inside it, so the page's advice produced the failure it exists to prevent | both paths named separately |
| The CI gate formatted `resistance_rate` after guarding on `no_verdict`, which is `False` when `total_attacks == 0`, giving a `TypeError` instead of a clean build failure | added the `rate is None` guard, matching the page's existing In CI idiom |
| `print(report.pipeline)` yields `Pipeline.STATIC` on 3.10, where the `StrEnum` polyfill has no `__str__` | prints `.value` |

The same pass cut a 17-line program whose only output was a printed `ValueError`, a one-command bash fence, and a false cross-reference to Fastest first run (different target, filter and cap). Those cuts were roughly offset by the fixes and by two lists that used to be run-on sentences — the section is 185 lines against the pre-review branch's 204.

Two later commits worked on the prose rather than the facts. `db7fe4a5` acts on a writing review by three roleplaying readers (a first-timer, a copy editor, and a practitioner scanning under deadline): the CI subsection leads with its two actions instead of filesystem background, the `previous_run=` reference forms and the nine rejected arguments are lists rather than run-on sentences, `cases` and `run store` are defined where they are first used, and the four `###` headings are parallel and action-shaped. `08390273` is a humanizer pass over the same prose.

`src/evaluatorq/redteam/README.md` gets a one-line fix: its copy of the rejected-argument set was missing `max_per_category`. That set is now written out in three places (guide, README, CLI help) and has already drifted once. A guardrail asserting the documented names equal the conflict tuple in `runner.py` would be worth having, but is out of scope here.

## Checks

`mkdocs build --strict`, `scripts/validate_mermaid.py`, `ruff check src`, `ruff format --check src`, `basedpyright` (0 errors), and `pytest -m 'not integration'` (5117 passed) are all green.

The three Python blocks in the new section were executed in document order against an empty run store before the review pass. The two blocks the review changed (`.value`, the `rate is None` guard) were re-verified statically against `RedTeamReport` and `ReportSummary` rather than re-run, because no API key was available in this environment.

## axes.md

Two edits, both unrelated to replay. One row records the retired `eq redteam ui` and `eq sim ui` Streamlit viewers as `N/A`, since they are deliberately de-documented in `cli-reference/dashboard.md` and re-reporting them weekly is noise. The other removes a phantom `eq ui` alias, which is registered nowhere in the CLI.

## Open questions for the reviewer

1. Single source of truth for the rejected set. The guide, `redteam/README.md` and the CLI `--from-run` help each write it out. The drift is not hypothetical — README was already missing one. Worth a `tests/test_reuse_guardrails.py`-shaped check?
2. Should `bt_sigma_aggregation` be recorded as a false-positive gap? It is internal, called only from `build_report(aggregation='bt-sigma')`, and users reach the concept through `report.bt_sigma`, documented at length in `pairwise-judging.md`. The coverage matrix flagged it on a symbol-name grep. A Tier note in `axes.md` would stop it resurfacing weekly — but narrowing the routine's own workload needs a human's approval.
3. Should the docs-autofill receipt be a hard gate? A nonzero block exit currently gets written into the table and the routine carries on. This run shows the risk: an early block genuinely failed (`exit=127`, `eq` not on `PATH`) and the stale artifact survived until two critics caught it. That change belongs in `.claude/skills/docs-autofill/SKILL.md`, not this PR.

## Known gaps, not addressed here

- Every run auto-uploads its results to `my.orq.ai`. Real and undocumented — but page-wide rather than replay-specific, so it belongs in `configuration.md` or the CLI reference.
- There is no CLI resistance-rate threshold (`--min-resistance-rate`). A feature gap; the page says the gate is Python and shows it.
- The dashboard has no red-team Compare view, though the simulation surface does — also a feature gap.
- The `02_static_dataset` docstring shows a flat sample schema while the fixture it points at is nested. Real docs drift — but on a different page, so it needs its own PR.
- Deployment-key discovery in the Orq UI is documented for `agent:` targets but not for `deployment:`. Pre-existing, different page.
- There is no replay cookbook under `examples/redteam/`.

---

Opened by the weekly docs-autofill routine.


Baukebrenninkmeijer:
## `claude-review` is red, and it is not this PR's failure — it needs a repo secret

Not pushing a fix, because there is nothing in this diff to fix. Root cause, from [the job log](https://github.com/orq-ai/evaluatorq/actions/runs/32628240565/job/97166859193):

```
Action failed with error: Environment variable validation failed:
  - Either ANTHROPIC_API_KEY, CLAUDE_CODE_OAUTH_TOKEN, or workload identity
    federation (ANTHROPIC_FEDERATION_RULE_ID and ANTHROPIC_ORGANIZATION_ID)
    is required when using direct Anthropic API.
```

`.github/workflows/claude-code-review.yml:29` passes `claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}`, and that secret resolves to empty in this repo. The action installed Claude Code fine (v2.1.241), obtained its OIDC app token fine, and then refused at credential validation — it never read the diff. **A re-run cannot change this**; a missing secret is deterministic, so I have not burned a cycle on one.

This diff touches three markdown files and nothing else — no `src/`, no workflow, no dependency — so it cannot have caused an auth failure in a review action.

**What a human needs to do:** add `CLAUDE_CODE_OAUTH_TOKEN` to the repo (or org) secrets. I have deliberately not touched the workflow to make the check pass — weakening a review gate's auth to get green is not a fix.

Worth noting this workflow is new — it merged three commits ago in #163, and this is its first run against a repo where the secret was never configured. Its three earlier runs were on the branch that introduced it. So this will fail on **every** PR until the secret exists, not just this one. The workflow's own prompt says it is advisory and never blocks a merge, which matches: it is not a required check.

### The rest of CI is green

`Build (strict)` — the docs build, the one that actually exercises this change — passed, along with `Lint`, `Examples (static)`, `Resolve SDK matrix`, and all three `SDK compatibility` jobs. The four `Typecheck + test` jobs were still running when I checked; I am subscribed and will follow up if any of them goes red.

---

currentlycodinng (approved):
Verified this against source, not just the PR's own receipts.

- The 9-item rejected set in "Replay refuses to be steered" (mode, dataset, categories, vulnerabilities, strategies, delivery_methods, max_per_category, max_dynamic_datapoints, max_static_datapoints) matches the `conflicting` list in `runner.py:698-708` exactly, including the label strings used in the ValueError message.
- The `mode` sentinel bug the note describes at `runner.py:698` (`Pipeline(mode) != Pipeline.DYNAMIC`) is real: since `mode` defaults to `Pipeline.DYNAMIC`, `mode="dynamic"` is indistinguishable from omitting it, so it passes the conflict check and gets silently overwritten by `replay.pipeline` two lines later. The note is accurate, not hedged into vagueness.
- `max_turns` and `attacker_instructions` restore-if-None behavior (`runner.py:719-723`) matches "explicit argument still wins."
- `--from-run` on `eq redteam run` (`cli.py:392-395`) and the CLI help epilog example at `cli.py:53` both check out. `ReplayError` text ("No saved {surface} runs found", `common/replay.py:151`) matches the doc.
- `vulnerability_rate`/`resistance_rate` complement claim: `converters.py` computes `resistant = eval_total - vulns` then feeds both into `_verdict_rate`, so `resistance_rate` is exactly `1 - vulnerability_rate`. The "reports the inverse metric" warning is correct.
- Retired-UI claims: `eq redteam ui` and `eq sim ui` are both still registered and print a deprecation warning pointing at `eq dashboard` (`redteam/cli.py:654`, `simulation/cli.py:2054`), matching the axes.md N/A row wording verbatim. Grepped for a standalone `eq ui` registration in `cli.py` and found none, so the "phantom alias" removal is correct, it was never real.
- All internal doc anchors resolve to real headings in the target files.

CI is green except `claude-review`, which is the known-flaky bot check, not a real gate. mkdocs strict build, all four Python-version typecheck+test jobs, and both example-run jobs all passed.

One non-blocking flag: this PR carries five "Open decisions" in the description that read like separate follow-up tickets (the mode-sentinel code fix, README/guide duplication, whether replay belongs under Modes, the bt_sigma false-positive, the receipt-gate skill change). None of them need to hold up this PR since the doc content is accurate either way, but they should get filed as actual issues or they'll get lost once this closes.

Approving. The replay mechanics, CI section, and axes.md edits all match what the code does.

Baukebrenninkmeijer:
Thanks for verifying against source rather than the receipts, @currentlycodinng — and good call on the open decisions.

Filed them so they survive this PR closing: **#167** tracks all five, each with its lean and source pointers. Two need a human call by nature (the `mode` sentinel is a code change; whether `replay` belongs under `## Modes` is a judgement call); the other three are mechanical once someone signs off.

Not merging — that's the routine's rule, and the decisions are yours to make.

---

arianpasquali (approved):

PR #166, opened 2026-08-23
docs: harden docs-autofill, fix replay mode conflict

Baukebrenninkmeijer:
Hardens the weekly docs-autofill routine against the failures its own first live run reported, and fixes a red-team bug that run found but was forbidden to touch.

## The rules the routine was missing

The first live run wrote a self-review listing six things that went wrong. Three now have a mechanism rather than a sentence:

- **The receipt is generated, never authored.** A block genuinely failed, was re-run out of band, and the file on disk was never regenerated — so four critics were briefed on a lie. The runner is now written *before* anything runs, lives at a tracked path, and is the only writer of the receipt. Hand-patching it is forbidden.
- **Blocks run against a fresh store** (`EVALUATORQ_DIR=$(mktemp -d)`). Round 1 ran against a store the agent's own probes had populated, which is how a hidden `previous_run="latest"` dependency survived review.
- **Nonzero exit is a hard gate**, aborting before the build step instead of becoming a table row nobody has to act on. Bash blocks also get the venv on `PATH` — the original `exit=127` was a broken runner, not a broken example.

Two more are written rules with no mechanism, and are labelled as such rather than claimed as fixed: hand reviewers artifacts instead of conclusions, and treat an unverified reviewer claim as a claim.

## The boundary

**The routine may never edit `src/`.** Not even a one-character docstring fence. A code/docs inconsistency is reported in the PR body *and* on Slack, and the gap is dropped with `skipped-code-defect` if the page cannot be written honestly around it. Writing a `!!! note` that explains a bug as though it were a design launders it into documented behaviour, which is much harder to undo later.

## Prose rules, in `CLAUDE.md` so they bind everyone

- **Never cut a sentence off** — pages, table cells, docstrings, commit messages, PR titles and bodies. Truncation belongs to captured output only.
- **Do not hard-wrap prose in a Markdown file you are editing.** A sentence split across source lines is the mechanism by which the first rule gets broken by accident. Counter-rule included: do not reflow a file you are not otherwise editing.
- PR titles capped at ~60 characters, one clause. `[BLOCKED]` and `[axes]` are the only permitted tags.

## `fix(redteam)`: an explicit `mode=` alongside `previous_run=`

`previous_run=` rejected eight data-selection arguments by name and silently swallowed the ninth. `mode` defaulted to `Pipeline.DYNAMIC`, so `Pipeline(mode) != Pipeline.DYNAMIC` could not distinguish "caller passed dynamic" from "caller passed nothing"; the value was then overwritten by the stored pipeline. It now uses the sentinel `max_turns` already used two lines away, on both the Python API and the CLI. Non-breaking.

## `replay` is a data source, not a fourth mode

`axes.md` modelled `replay` as a mode value. `runner.py` does `mode = replay.pipeline`, so a replayed run already *has* a mode — `--mode` conflicts because the value would be overwritten, not because the two are rival spellings. Moved to the data-source axis, with the impossible pair listed in the N/A table.

## Review

Five critics (four Claude, one Codex through a different model family). Every finding was verified against the file before acting; the fixes are in `fdae176c`. Highlights:

- The `Caps` summary still permitted the `src/` docstring-fence edit that step 3 had just closed by name — the section an agent skims under pressure, stating the opposite of the body.
- `.context/` is gitignored, so a receipt runner "committed with the branch" could never be staged.
- The ledger wrote `opened` before `gh pr create`, so a failed PR creation buried the gap permanently. It now writes `prepared` and amends on success.
- One new test was a byte-identical duplicate of the test below it and pinned nothing. Deleted.
- **A rule I had added was simply wrong.** It claimed `bt_sigma_aggregation` is reachable only via `build_report(aggregation='bt-sigma')`. It is in `evaluatorq.__all__`, which makes it Tier 1 by this repo's own definition. Removed — the coverage matrix should keep flagging it.

Rejected: the unwrapping was requested, not drive-by; `ORQ_WORKSPACE` isolation is display-only and not a gate; the `resolved_*` naming nit changes no behaviour.

## Deliberately not done

- The `## Modes` list in `docs/guides/red-teaming.md` still lists `replay` as a fourth mode, and the replay single-source-of-truth refactor is not started. Both belong to unmerged #165, which owns that section; doing them here would fight that PR in the same file. #165 also edits `axes.md`'s mode row, so expect one small conflict there.
- Moving the routine's mechanical steps into scripts is tracked as RES-1417 rather than done the night before a scheduled run.

## Timing

The routine fires from `main` Mondays at 06:08 UTC. Until this merges, it runs the old rules.


currentlycodinng (approved):
Checked the mode/replay fix against source, not just the description.

`src/evaluatorq/redteam/cli.py:258-262` and `runner.py:488,527` change `mode` from defaulting to `Pipeline.DYNAMIC` to defaulting to `None`. The conflict check at `runner.py:745` now reads `mode is not None` instead of `Pipeline(mode) != Pipeline.DYNAMIC`, and the default gets applied after the replay branch at `runner.py:754-755` (`elif mode is None: mode = Pipeline.DYNAMIC`). That's the right fix for the bug described: previously `--mode dynamic --from-run latest` silently passed the conflict check and got overwritten by `replay.pipeline`, because dynamic was indistinguishable from "not supplied." Confirmed `Pipeline` only has `static`/`dynamic`/`hybrid` (`contracts.py:278-283`), so there's no risk of a real fourth value colliding with the sentinel. New test `test_previous_run_rejects_an_explicit_dynamic_mode` (`test_replay.py:767-781`) asserts exactly this case and passes.

Spot-checked the docs claims that are checkable against source rather than taking them on faith:
- `axes.md`'s reframe of `replay` as a data source not a mode matches `runner.py:722` (`mode = replay.pipeline`) and the `Pipeline` enum having no `replay` member.
- `.gitignore:83` does have `.context/`, backing the SKILL.md claim that a receipt runner written there can't be committed.
- `mkdocs.yml` does have both `nav:` and `plugins.llmstxt.sections`, with a build hook enforcing parity, matching the "goes in twice" rule.
- `.github/CODEOWNERS` does request 3 named reviewers on every PR and skip the author, matching the CLAUDE.md claim.
- `bt_sigma_aggregation` is in `evaluatorq.__all__` (`__init__.py:115`), so the self-correction described in the PR body (deleting the wrong Tier 2 claim) is accurate, and the wrong claim is in fact gone from `axes.md` in this diff.

Claude Code Review check failure is infra, not the PR: missing `ANTHROPIC_API_KEY`/`CLAUDE_CODE_OAUTH_TOKEN` in that workflow run. Not a finding on the code. All required CI (SDK compatibility x3, Examples, Typecheck+test x4 Python versions, Lint, Docs build --strict) is green.

Rest of the diff is prose reflow/hard-wrap removal in markdown, which checks out everywhere I sampled it for meaning changes.

Approving.

PR #169, opened 2026-08-24
fix(tracing): enforce the OTEL flush timeout, and document ORQ_OTEL_* batching

Baukebrenninkmeijer:
## What this covers

Documents the four OpenTelemetry `BatchSpanProcessor` tuning environment variables —
`ORQ_OTEL_MAX_QUEUE_SIZE`, `ORQ_OTEL_MAX_BATCH_SIZE`, `ORQ_OTEL_SCHEDULE_DELAY_MS`,
`ORQ_OTEL_FLUSH_TIMEOUT_MS` — which previously existed only in the
`src/evaluatorq/tracing/setup.py` docstring, i.e. the generated API reference, which
`docs-coverage` does not count as discovery (Tier 1: every env var needs prose).

Writing that prose surfaced defects in the code it described, so the PR grew a `fix(tracing):`
half. The scope is wider than the docs-autofill routine's default and is described below.

### Docs

A **"Batching and flush"** section in `docs/tracing.md`, covering how the processor queues and
drains spans, the two ways spans are silently lost, what each variable actually controls and when
it is read, and a CI recipe in both shell and GitHub Actions form. The four variables are added to
the canonical env-var table in `docs/configuration.md`, and an FAQ entry gives the symptom
("some spans are missing from my traces") a discovery path.

Two pre-existing inaccuracies in the same page are corrected: the red-team span tree drew an
`orq.redteam.security_evaluation` span that does not exist, and showed the multi-turn spans
(`attack_turn`, `adversarial_generation`, `context_retrieval`, `datapoint_generation`) as the
general shape when static mode emits none of them. Judge LLM spans were documented as
`chat {model}`; `EvaluatorConfig.api` defaults to `'responses'`.

### Code

Three behaviour fixes, each one a claim the docs could not have made truthfully otherwise:

- **`ORQ_OTEL_FLUSH_TIMEOUT_MS` did nothing.** `BatchProcessor.force_flush` ignores its
  `timeout_millis` argument and unconditionally returns `True`
  ([open-telemetry/opentelemetry-python#4568](https://github.com/open-telemetry/opentelemetry-python/issues/4568)),
  so the value was passed down and discarded and the timeout-warning branch was unreachable. A run
  whose collector stopped responding waited for the export to finish, however long that took.
  `flush_tracing` now enforces the bound with `asyncio.wait_for`.
- **A set-but-empty value fell back silently.** `_env_int` used a falsy check, so
  `ORQ_OTEL_MAX_QUEUE_SIZE=""` — what an unresolved workflow variable expands to in a CI `env:`
  block — took the same path as an unset variable. Empty and whitespace-only now warn like every
  other rejected value.
- **The batch-size clamp was silent.** `max_export_batch_size` was clamped to the queue size with a
  bare `min()` next to a helper that warns on every rejected value: two adjacent branches differing
  in whether they log, which is what the "a degraded path announces itself" house rule exists to
  prevent. It warns now.

## Verification

| Check | Result |
|---|---|
| `uv run ruff check src` | pass |
| `uv run ruff format --check src` | pass (229 files) |
| `uv run basedpyright` | 0 errors, 0 warnings, 0 notes |
| `uv run pytest -m 'not integration'` | 5042 passed, 2 skipped |
| `uv run --group docs mkdocs build --strict` | exit 0 |
| `uv run python scripts/validate_mermaid.py` | `mermaid OK — 39 file(s)` |

CI's lint/typecheck/test jobs run on every PR (neither `ci.yml` nor `docs.yml` has a `paths:`
filter) and are listed here because this PR now changes `src/`. The job that gates the docs half is
`Docs / Build (strict)`.

Behaviour claimed in prose is pinned by tests rather than asserted:

- the three documented defaults (`4096` / `512` / `5000`) are asserted where the processor consumes
  them, so changing the literals fails instead of silently contradicting the page;
- an all-default initialization must not trip the clamp warning;
- a hanging force-flush is bounded and warns;
- `_env_int`'s empty and whitespace cases are in both parametrizations.

The two live claims that prompted the corrections were reproduced directly against the pinned SDK
(1.42.1): `WARNING:opentelemetry.sdk._shared_internal:Queue full, dropping Span.` from a
`max_queue_size=2` processor with a blocking exporter, and
`ORQ_OTEL_MAX_QUEUE_SIZE= → default 4096, no warning` before the fix.

## Review

Reviewed by two rounds of parallel agents, both read-only:

1. Six lenses over the diff (docs accuracy, verification adequacy, persona readability, minimality,
   repo standards, structure). Four of five independently flagged the same Critical defect — the
   section claimed queue overflow had "no first-party signal" and sent readers to hand-count spans
   in the Orq trace UI, when the SDK logs on every drop.
2. Six Codex reviewers (`gpt-5.6-luna`, read-only): four writing lenses plus two fact-checkers that
   validated every assertion in the changed pages against source. They found the `force_flush`
   no-op, the wrong drain model (the worker wakes immediately at the batch threshold, so the
   "512 spans per 5 seconds" arithmetic and the advice to lower the schedule delay under a burst
   were both wrong), the wrong logger name and dedup window, and a `dashboard.md` paragraph
   asserting a shared tracer provider for a process that never initializes tracing.

Every finding was verified against source before being acted on. Two house-voice claims were
checked and **rejected**: the corpus is majority American spelling (`behavior` 19 / `behaviour` 4),
and statement-form FAQ headings already exist. Two were checked and **applied**: `›` for section
breadcrumbs in link labels (8 existing uses, no `→`) and `**Label** —` for list labels (31 existing
uses, no `**Label.**`).

## Scope

Beyond the routine's default "one section on an existing page":

- `docs/configuration.md` — the gap is not closed unless the canonical env-var reference lists the
  variables, or next week's `docs-coverage` re-derives the same gap.
- `docs/faq.md`, `docs/dashboard.md` — discovery path from the symptom.
- `.claude/skills/docs-coverage/axes.md` — the ledger row named a `surface` value that does not
  exist. Env vars are a Tier-1 obligation, not a matrix cell; the file now says how to record one so
  the free-text dedupe matches. `CLAUDE.md` requires an `axes.md` change to land in the same PR.
- `src/`, `tests/`, `CHANGELOG.md` — the three fixes above.

## Follow-ups, deliberately not in this PR

- **`docs/tracing.md` has further pre-existing drift** found by the fact-check sweep and not yet
  verified by hand: persona/scenario LLM calls shown without their
  `orq.simulation.persona_generation` / `scenario_generation` wrappers; `orq.score` described as
  JSON when scalars go through `str()`; the `single_trace=False` paragraph ignoring an ambient
  parent; the `.orq.ai` auth check using `netloc`, so `my.orq.ai:443` is not recognised as an Orq
  domain.
- **`docs-drift` runs scoped-to-diff**, so it never checks `configuration.md` when an env-var gap is
  filled on a surface page. Handled manually here; the durable fix is a tooling change.

---

_Opened by the weekly docs-autofill routine; scope extended during review._


Baukebrenninkmeijer:
The `claude-review` check is red, but it is not caused by this change and cannot be fixed from this PR.

The [failing job](https://github.com/orq-ai/evaluatorq/actions/runs/32697974472/job/97343651448) died in ~16s at the "Run Claude Code Action" step with `ANTHROPIC_API_KEY:` empty in its step environment — a missing-secret configuration issue in the review workflow, before any review of the diff ran. It's deterministic (not a flake) and would fail identically on any PR, so re-running it won't help while the secret is unset, and there is no code or docs change here that would turn it green.

All checks relevant to this docs-only change are passing: **Build (strict)**, **Lint**, **Examples (static)**, and **SDK compatibility**. Leaving `claude-review` for a maintainer, since fixing it means setting the workflow secret, not editing this branch.

---

currentlycodinng (approved):
Verified this against src/evaluatorq/tracing/setup.py on the PR branch (feb073f2), not just the diff.

Var names and defaults, all exact matches:
- ORQ_OTEL_MAX_QUEUE_SIZE, default 4096 (setup.py:214)
- ORQ_OTEL_MAX_BATCH_SIZE, default 512 (setup.py:215)
- ORQ_OTEL_SCHEDULE_DELAY_MS, default 5000 (setup.py:219)
- ORQ_OTEL_FLUSH_TIMEOUT_MS, default 5000 (setup.py:259)

Read timing: docs say the first three are read "at init," the fourth "per run." Confirmed, the first three are only read inside `init_tracing_if_needed()`, gated by `_initialization_attempted` and runs once per process; `ORQ_OTEL_FLUSH_TIMEOUT_MS` is read fresh in `flush_tracing()` every call (setup.py:259). Correct.

Batch-size clamping: docs say `ORQ_OTEL_MAX_BATCH_SIZE` is "clamped down to the queue size." setup.py:220 does exactly that: `max_export_batch_size=min(requested_batch_size, max_queue_size)`. This isn't cosmetic, the underlying otel-sdk `BatchSpanProcessor` raises `ValueError` if batch size exceeds queue size, so the clamp is load-bearing, not just tidiness. `tests/common/test_tracing_lifecycle.py:197-206` exercises this directly: `MAX_BATCH_SIZE=200` with `MAX_QUEUE_SIZE=100` asserts `max_export_batch_size == 100`. Docs match.

Invalid-value fallback: docs say a non-numeric or non-positive value logs a WARNING and falls back to default. Matches `_env_int()` at setup.py:51-68 exactly (the ValueError branch and the `value <= 0` branch both warn and return default).

Silent queue overflow: docs say overflow drops spans with no log and no exception. Confirmed at the otel-sdk level, its queue is a `collections.deque(maxlen=...)` that evicts silently on append. The doc's "no first-party signal" framing is accurate.

Flush timeout warning text: docs quote it verbatim as "OTEL span flush timed out after <ms>ms; some spans may not have been exported." Matches the `logger.warning()` format string at setup.py:263-266 character for character.

No stray vars either direction: grepped `ORQ_OTEL_` across src/ and tests/, the only hits outside tracing/setup.py and tracing/context.py are the four in `tests/common/test_tracing_lifecycle.py` exercising these same four vars. Nothing documented that doesn't exist in code, nothing in code left undocumented.

`configuration.md` changes are consistent with `tracing.md` and don't duplicate content, just cross-reference it. All internal anchors resolve to real headings.

CI green except `claude-review`, the known-flaky bot check, not the diff's fault.

Approving.

arianpasquali (approved):

PR #170, opened 2026-08-24
test(guardrails): fail when a Sphinx role reappears in a docstring

Baukebrenninkmeijer:
## What

Adds one AST-based guardrail to `tests/test_reuse_guardrails.py`: no docstring under `src/evaluatorq` may contain a Sphinx cross-reference role (`:class:`, `:func:`, `:meth:`, `:attr:`, `:mod:`, `:data:`, `:exc:`, `:obj:`).

## Why

#139 stripped all 228 role usages, but left no way to stop them returning. mkdocstrings does not understand roles: the prefix and the `~` shorthand render as literal text on the API reference pages, and `mkdocs build --strict` stays green while it happens — so the regression is invisible until someone reads the site.

The ticket suggested hanging the check off the `htmlproofer` plugin. A source-level test is stricter (it catches docstrings that have no reference page yet, which are the majority) and fails in the normal `pytest` run rather than only in a docs build.

## Verification

- Passes on the current tree (0 roles in `src/`).
- Fails as intended: dropping a module with `See :class:\`Foo\`.` into `src/evaluatorq/` turns the test red, naming the file and line.
- `ruff check src`, `ruff format --check src`, `basedpyright`, `pytest -m 'not integration'` all clean (5038 passed).

Closes RES-1278.

### DIFF
--- a/tests/test_reuse_guardrails.py
+++ b/tests/test_reuse_guardrails.py
@@ -13,6 +13,7 @@ not be a way around the rule by accident.
 from __future__ import annotations
 
 import ast
+import re
 from functools import cache
 from pathlib import Path
 
@@ -283,4 +284,35 @@ def test_exception_usage_goes_through_the_shared_helper() -> None:
         + '. Use evaluatorq.common.structured_output.usage_from_exception — it owns '
         'the harvest rule (why getattr rather than an except clause, and when the '
         'exception carries no total so the result must not be double-counted).'
+    )
+
+
+# Sphinx cross-reference roles (`:class:`Foo``) render as literal text on the
+# MkDocs site — the role prefix and the `~` shorthand both print. RES-1278.
+SPHINX_ROLE = re.compile(r':(?:class|func|meth|attr|mod|data|exc|obj):`')
+
+
+def _sphinx_role_sites() -> list[str]:
+    """``path:line`` for every docstring carrying a Sphinx cross-reference role."""
+    sites: list[str] = []
+    for path in sorted(SRC.rglob('*.py')):
+        rel = path.relative_to(SRC).as_posix()
+        tree = ast.parse(path.read_text(encoding='utf-8'))
+        for node in ast.walk(tree):
+            if not isinstance(node, ast.Module | ast.ClassDef | ast.FunctionDef | ast.AsyncFunctionDef):
+                continue
+            doc = ast.get_docstring(node, clean=False)
+            if doc and SPHINX_ROLE.search(doc):
+                sites.append(f'{rel}:{getattr(node, "lineno", 1)}')
+    return sites
+
+
+def test_docstrings_carry_no_sphinx_roles() -> None:
+    sites = _sphinx_role_sites()
+    assert not sites, (
+        'Sphinx cross-reference role in a docstring: '
+        + ', '.join(sites)
+        + '. mkdocstrings has no idea what a role is, so `:class:` and the `~` '
+        'shorthand render as literal text on the API reference pages. Use a plain '
+        'code span, or an mkdocstrings autoref: [Message][evaluatorq.contracts.Message].'
     )
--- a/tests/test_reuse_guardrails.py
+++ b/tests/test_reuse_guardrails.py
@@ -289,22 +289,55 @@ def test_exception_usage_goes_through_the_shared_helper() -> None:
 
 # Sphinx cross-reference roles (`:class:`Foo``) render as literal text on the
 # MkDocs site — the role prefix and the `~` shorthand both print. RES-1278.
-SPHINX_ROLE = re.compile(r':(?:class|func|meth|attr|mod|data|exc|obj):`')
+# Covers the autodoc object-reference roles plus the three doc-level roles
+# (`ref`, `paramref`, `term`) that break the same way. Domain-specific roles
+# nobody here writes (`:option:`, `:envvar:`, …) are out of scope until one shows up.
+SPHINX_ROLE = re.compile(r':(?:class|func|meth|attr|mod|data|exc|obj|ref|paramref|term):`')
+
+
+def _sphinx_role_lines(source: str) -> list[int]:
+    """Line numbers of every docstring in ``source`` carrying a Sphinx role.
+
+    A module docstring reports line 1: ``ast.Module`` has no ``lineno``.
+    Roles in comments or in ordinary strings are not docstrings and not hits.
+    """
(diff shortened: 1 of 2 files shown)


tests/test_reuse_guardrails.py
```diff
@@ -284,3 +285,34 @@ def test_exception_usage_goes_through_the_shared_helper() -> None:
...
+    for path in sorted(SRC.rglob('*.py')):
+        rel = path.relative_to(SRC).as_posix()
+        tree = ast.parse(path.read_text(encoding='utf-8'))
+        for node in ast.walk(tree):
+            if not isinstance(node, ast.Module | ast.ClassDef | ast.FunctionDef | ast.AsyncFunctionDef):
+                continue
+            doc = ast.get_docstring(node, clean=False)
+            if doc and SPHINX_ROLE.search(doc):
+                sites.append(f'{rel}:{getattr(node, "lineno", 1)}')
+    return sites
```
currentlycodinng: This is the only AST guardrail in the file without a paired synthetic self-test (compare `test_detection_actually_fires` at line 162, `test_hand_written_properties_detector_actually_fires` at line 230). The file's own philosophy is that an unproven guardrail can silently no-op, worth adding a small parametrized test here that runs the detector against a synthetic docstring with a role, and one without, so a future edit to SPHINX_ROLE can't quietly break detection.

tests/test_reuse_guardrails.py
```diff
@@ -284,3 +285,34 @@ def test_exception_usage_goes_through_the_shared_helper() -> None:
         'the harvest rule (why getattr rather than an except clause, and when the '
         'exception carries no total so the result must not be double-counted).'
     )
+
+
+# Sphinx cross-reference roles (`:class:`Foo``) render as literal text on the
+# MkDocs site — the role prefix and the `~` shorthand both print. RES-1278.
+SPHINX_ROLE = re.compile(r':(?:class|func|meth|attr|mod|data|exc|obj):`')
```
currentlycodinng: List covers class/func/meth/attr/mod/data/exc/obj but not ref (or paramref/term). `:ref:` breaks the same way on mkdocstrings if it ever shows up in a docstring. Not blocking since it wasn't part of what #139 removed, but worth a one-line comment saying the scope is deliberately autodoc object-reference roles, not all Sphinx roles, so it doesn't read like an oversight.

currentlycodinng (commented):
Cloned and ran it. `test_docstrings_carry_no_sphinx_roles` is green on main, and I confirmed red on the actual bug: dropped a probe module with `:class:`Foo`` and `:func:`baz`` / `:func:`qux`` (two targets, same role) into src/evaluatorq, the test failed and named both path:line sites correctly, then removed it. Also checked the Module-node lineno fallback at line 306, `ast.Module` has no lineno attribute, so `getattr(node, "lineno", 1)` is load-bearing, not decorative, for a module-level hit.

Two things I'd want before calling it done:

1. Every other AST guardrail in this file (`test_detection_actually_fires`, `test_hand_written_properties_detector_actually_fires`) ships a synthetic self-test proving the detector can actually fire, the file's own docstring says a guardrail nobody proved can fail is one that silently no-ops. This one doesn't have that companion test, only the manual verification described in the PR body.

2. `SPHINX_ROLE` (line 292) covers class/func/meth/attr/mod/data/exc/obj but not ref, paramref, or term. Those render the same broken way on mkdocstrings. Worth either adding them or noting in the comment that the list is deliberately scoped to autodoc object-reference roles.

Neither is a false positive or a miss on the case this PR is meant to catch, so not blocking, but I'd like the self-test added before merge.

Baukebrenninkmeijer:
Thanks — both adopted, pushed in 4a3b0f0-ish (latest commit on the branch).

1. **Self-test added.** Split the scan into `_sphinx_role_lines(source)` so the detector runs on synthetic input like `_hand_written_properties_dict_lines` does, with `_sphinx_role_sites()` now just mapping it over the tree. `test_sphinx_role_detector_actually_fires` covers: every role in the pattern fires; the `~` shorthand and a fully qualified target fire; two docstrings report two line numbers; and a comment, a bare expression string, an assigned string and a plain code span / autoref do not. Your point about the `getattr(node, "lineno", 1)` fallback is now pinned by the first assertion — a module docstring reports line 1, and the docstring says why.

2. **Roles widened.** Added `ref`, `paramref`, `term`. The comment above the pattern now states the scope explicitly (autodoc object-reference roles plus those three) and says domain-specific roles like `:option:` / `:envvar:` are out until one actually shows up, so the next person hits a decision rather than an oversight.

Re-verified end to end: green on the tree, and red with a probe module using the two newly added roles, naming the site.

tests/test_reuse_guardrails.py
Baukebrenninkmeijer: Added in a0a44c66. Split the scan into `_sphinx_role_lines(source)` so the detector takes a source string like `_hand_written_properties_dict_lines` does — `_sphinx_role_sites()` now just maps it over the tree — and added `test_sphinx_role_detector_actually_fires` alongside it. It asserts each role fires, the `~` shorthand and a fully qualified target fire, two docstrings report two distinct line numbers, and that a comment, a bare expression string, an assigned string and a plain code span / autoref all stay clean. The module-docstring case is the first assertion, so the `getattr(node, "lineno", 1)` fallback you spotted is now pinned by a test rather than by a comment.

tests/test_reuse_guardrails.py
Baukebrenninkmeijer: Added `ref`, `paramref` and `term` in a0a44c66 — you're right that they render the same broken way, and the cost of covering them is zero. The comment above the pattern now names the scope out loud: autodoc object-reference roles plus those three doc-level ones, with domain-specific roles (`:option:`, `:envvar:`, …) deliberately out until one actually appears in a docstring. That way the next person adding one hits a stated decision instead of guessing whether the omission was intentional. Verified with a probe module using the two new roles: the guard fails and names the site.

currentlycodinng (approved):
Reviewed the latest commit, including the two earlier follow-ups. The guardrail now proves its detector fires on each scoped role, reports module and nested-docstring sites correctly, and does not flag comments or ordinary strings. The role scope is documented and covers the additional doc-level roles that fail the same way in mkdocstrings.\n\nRan tests/test_reuse_guardrails.py (19 passed), ruff, format check, and basedpyright; CI is green across Python 3.10–3.13. No actionable findings.

PR #173, opened 2026-08-24
fix: stop sending a default temperature on every LLM call

### LINKED ISSUE #168: Weekly example run failed (2026-08-24)
One or more curated examples failed to run.

Run: https://github.com/orq-ai/evaluatorq/actions/runs/32696446156

Likely either genuine drift in the examples or a flaky upstream API — check which before editing anything.

Baukebrenninkmeijer:
Closes #168 — the weekly curated-example failure.

## What broke

`examples/agent_simulation/01_basic_simulation.py` died with a 400 from the default pipeline model `openai/gpt-5.6-luna`:

```
Unsupported parameter: 'temperature' is not supported with this model.
```

Reasoning-class models reject `temperature` outright rather than clamping it. evaluatorq sent one from nine internal call sites plus a `LLMCallConfig.temperature` field defaulting to `1.0`, so every persona x scenario pair failed and `simulate()` raised `RuntimeError: first-message generation produced no datapoints`. Deterministic, not flaky — it would have failed every week until fixed.

## What changed

`LLMCallConfig.temperature` defaults to `None`, and `request_params` omits the key entirely when unset, on both endpoints. A call site passing `temperature=None` means "leave it unset", never "send null". Every internal literal is deleted: `0.8` for persona and first-message generation, `0.9` for edge-case scenarios, `0.7` for the executive summary and chat-completions agent calls, `0.3` for trace analysis, `0.0` for the judge.

`BaseAgent._resolved_temperature` loses its `fallback` argument and gates on `model_fields_set` like its three sibling resolvers, so an explicit `LLMCallConfig(temperature=None)` opts an agent out instead of deferring to a call site.

Left alone deliberately: `simulation/convert.py`'s `'temperature': 1` builds a synthetic *Response* payload, not a request, and `owasp/models.py`'s `LlmEvaluatorEntity.temperature` has no reader anywhere.

## Behaviour change worth planning around

The judge's scoring calls are no longer pinned to `temperature=0.0`, so judgments are no longer reproducible run-to-run by default. This affects every `simulate()` caller, not only those on a reasoning model. `JudgeAgentConfig(temperature=0.0)` restores it on a model that accepts the parameter. Called out under `### Notable defaults` in `CHANGELOG.md`.

## Verification

All three curated examples in the weekly job were run against the live API, not just the one that failed:

| Example | Result |
|---|---|
| `agent_simulation/01_basic_simulation.py` | 6 turns, no 400 |
| `redteam/02_static_dataset.py` | 3/3 attacks, 100% resistance |
| `redteam/01_basic_dynamic.py` | 5/5 evaluated, 2 vulnerabilities |

The four CI commands verbatim: `ruff check src`, `ruff format --check src`, bare `basedpyright` (0 errors), `pytest -m 'not integration'` (5044 passed). `mkdocs build --strict` green.

New regression tests cover the path that actually broke — `temperature` absent from the request on `_call_chat_completions`, `_call_responses`, the first-message generator and the judge — plus the explicit-`None`-opts-out case.

## Open decision, not addressed here

Four surfaces have no way to pass a temperature at all: `PersonaGenerator`, `ScenarioGenerator`, `TraceAnalysisConfig` and `populate_run_executive_summary`. They never did — the values were hardcoded and unconfigurable — but now nothing is sent where a tuned value used to be. Threading a `temperature` parameter through all four is a feature, not part of fixing a broken CI job, so it is left for a separate issue.

## Second commit

`docs:` extends the no-hard-wrap prose rule in `CLAUDE.md` and the docs-autofill skill to PR bodies, issue text and review comments, and names code comments and docstrings as the one place that still wraps.


currentlycodinng (approved):
Reviewed the latest commit. The default path now omits temperature on both endpoints, including the simulator, first-message generator, judge, reports, and trace analysis; explicit values still pass through, and an explicit None overrides a call-site literal. The public default change and the judge-determinism trade-off are documented.\n\nRan the focused regression tests, Ruff, format check, and basedpyright locally. CI is green across Python 3.10–3.13, including strict docs. No actionable findings.

PR #174, opened 2026-08-24
feat(simulation): expose an llm_config on every simulation entry point

Baukebrenninkmeijer:
## Why

Unsetting the `LLMCallConfig.temperature` default (#173) fixed the reasoning-model 400s but took the only lever with it: there was no way to pin sampling on the simulator's own calls short of hand-building each agent. Red teaming already had `red_team(llm_config=LLMConfig(attacker=..., evaluator=...))`; simulation had nothing equivalent.

## What

`llm_config: LLMCallConfig` on `simulate()`, `generate_and_simulate()`, `generate()`, `generate_personas()` / `generate_scenarios()` (and singular forms), `summarize_conversations()`, `datapoints_from_traces()`, `extend_from_traces()` and `extend_from_experiment()`.

One config drives every simulation-side call — user simulator, judge, persona / scenario / first-message generators, executive summary. It never touches the target under test; that is the thing being measured and is configured where it is constructed.

`sim_model=` stays as the shorthand for setting only the model. When both are given `llm_config.model` wins and the contradiction is logged, so a run cannot quietly use a model the caller did not pick. Only the fields you set take effect, so an unset `temperature` still means the parameter is omitted from the request.

Three fields are narrower and say so when they do not apply: `max_tokens`, `api` and `retry_count` reach the two agents, while every `generate_structured` call site — the generators, the recommendations pass, the trace helpers — owns them itself and logs which of them the config supplied in vain.

## The agents take the config directly

`UserSimulatorAgent` and `JudgeAgent` accept an `LLMCallConfig` plus their prompt data as separate keyword arguments (`system_prompt=`, and `goal=` / `criteria=` / `ground_truth=`), and `SimulationRunner` hands its config straight over. Routing it through the legacy `AgentConfig` on the way lost any field explicitly set to `None` — `temperature=None` means "send no temperature", and `AgentConfig` spells "unset" the same way, so the two collapsed together.

`AgentConfig` and its two subclasses stay accepted and deprecated; deleting them is breaking and belongs to RES-1421.

Both agents now default to the **Responses API** when the config leaves `api` unset — `BaseAgent.DEFAULT_API`, a class attribute rather than an environment variable, since it is a per-call setting. The two agent config classes supplied that default before, so a caller handing an agent a bare `LLMCallConfig` used to get chat completions, which returns a 400 on models like `gpt-5.4-mini` the moment the judge sends function tools and `reasoning_effort` together. Pass `LLMCallConfig(api='chat_completions')` to opt out.

## Supporting pieces

- `LLMCallConfig.timeout_s(default)` — one home for the `timeout_ms / 1000` conversion, which had four copies.
- `resolve_sim_llm_config` — one home for the `sim_model` / `llm_config` fold, which had five copies across the generators. `sim_model=` was silently discarded whenever an `llm_config` without an explicit model was also passed; it now applies, and only a genuine contradiction warns.
- `config=` on `common.structured_output.generate_structured` — fills in only the sampling knobs the caller did not name, and warns naming any field it will not read. `model` and `api` stay the function's own authority so a config cannot silently redirect a call to another endpoint.
- `config=` on `common.reports.executive_summary.generate_executive_summary`, replacing the six loose sampling keywords it used to rebuild into a config internally. One route per setting, so there is no precedence to guess.
- `SimulationConfig` is `extra='forbid'`, so `SimulationConfig(model=...)` raises instead of being swallowed now that `model` is a derived property.
- `simulate()` warns when `llm_config` carries settings that cannot reach an injected `judge=` / `user_simulator=`, which arrives already built.

## Docs

`docs/tuning.md` gains a "Where to put that config" subsection under Temperature, covering the single-config trade-off, the Responses default and the three narrower fields. `CHANGELOG.md` gets a New Features entry plus a Notable defaults entry for the endpoint change. `.claude/skills/docs-coverage/axes.md` gains an **own-calls LLM config** axis with its impossible combinations — a new dimension users choose along, so coverage checking needs to see it.

## Checks

`ruff check src`, `ruff format --check src`, bare `basedpyright`, `pytest -m 'not integration'` (5154 passed), `mkdocs build --strict` — all green locally.


src/evaluatorq/common/structured_output.py
```diff
@@ -704,7 +706,28 @@ async def generate_structured(
...
+        set_fields = config.model_fields_set
+        if temperature is None and 'temperature' in set_fields:
+            temperature = config.temperature
+        if extra_kwargs is None and 'extra_kwargs' in set_fields:
+            extra_kwargs = config.extra_kwargs
+        if extra_body is None and 'extra_body' in set_fields:
+            extra_body = config.extra_body
+        if reasoning_effort is None and 'reasoning_effort' in set_fields:
+            reasoning_effort = config.reasoning_effort
+        if timeout_s == _STRUCTURED_TIMEOUT_S and 'timeout_ms' in set_fields:
```
currentlycodinng: `timeout_s == _STRUCTURED_TIMEOUT_S` can't tell "caller didn't pass timeout_s" from "caller explicitly passed the default value", a caller pinning `timeout_s=_STRUCTURED_TIMEOUT_S` on purpose gets silently overridden by `config.timeout_ms`. Same shape as the `sim_model != DEFAULT_MODEL` check in api.py:118. Not a blocker, just worth a comment noting the limitation since the docstring above claims an explicit keyword always wins.

src/evaluatorq/common/structured_output.py
```diff
@@ -704,7 +706,28 @@ async def generate_structured(
...
+    ``config`` supplies the sampling knobs a caller holds as one object rather
+    than as five keywords: ``temperature``, ``extra_kwargs``, ``extra_body``,
+    ``reasoning_effort`` and ``timeout_ms``. Only the fields the caller
+    explicitly set on it are read (``model_fields_set``), and an explicit
+    keyword here still wins — a call site pinning ``temperature=0.0`` for a
+    reason keeps it. ``config.model`` and ``config.api`` are NOT read: this
+    function's own ``model`` / ``api`` arguments stay the single authority, so a
+    config cannot silently redirect a call to another endpoint.
     """
+    if config is not None:
```
currentlycodinng: This merge (5 fields keyed on `model_fields_set`) has no direct unit test anywhere in the diff or existing suite, every caller that exercises it (PersonaGenerator, ScenarioGenerator, traces.py) mocks `generate_structured` out in the new tests. Worth a small `tests/common/test_structured_output.py` case that passes a real `LLMCallConfig` and asserts the resolved kwargs, since this is the one function all the generators funnel through.

src/evaluatorq/simulation/_config.py
```diff
@@ -77,6 +77,14 @@ class SimulationConfig(BaseModel):
     """None means "unset": resolved in ``_simulate_core`` to a replayed run's
     cap when replaying, else to ``DEFAULT_MAX_TURNS``."""
     model: str = DEFAULT_MODEL
+    """Resolved simulation model. Always equals ``llm_config.model`` — both are
```
currentlycodinng: `model` is documented to always equal `llm_config.model` but that's enforced by discipline at the two SimulationConfig construction sites, not by a validator. Consider deriving it as a property instead of duplicating the field, so a future third construction site can't drift.

src/evaluatorq/simulation/api.py
```diff
@@ -99,6 +100,31 @@ async def _attach_recommendations(
...
+    silent loss — the two used to be resolved at different layers, so whichever
+    the reader looked at was the one they believed.
+
+    When ``llm_config`` is omitted the shorthand still works and produces a
+    config with only ``model`` set, so every other field stays unset and the
+    per-call-site defaults keep applying.
+    """
+    if llm_config is None:
+        return LLMCallConfig(model=sim_model)
+    if sim_model != DEFAULT_MODEL and llm_config.model != sim_model:
```
currentlycodinng: If a caller explicitly passes `sim_model=DEFAULT_MODEL` (a plausible thing to do) alongside a contradicting `llm_config`, this silently skips the warning since it can't distinguish 'not passed' from 'passed the default.' llm_config.model still wins correctly, so no behavior bug, just a suppressed diagnostic.

currentlycodinng (commented):
Traced llm_config through every entry point named in the PR body and CHANGELOG: `simulate()`, `generate_and_simulate()`, `generate()`, `generate_personas/generate_persona`, `generate_scenarios/generate_scenario`, `summarize_conversations()`, `datapoints_from_traces()`, `extend_from_traces()`, `extend_from_experiment()`. All of them thread it down to the actual call sites (PersonaGenerator/ScenarioGenerator/FirstMessageGenerator/DatapointGenerator, the judge and user simulator in `runner/simulation.py`, the executive summary). Nothing claimed is missing.

Backward compat checks out: every new `llm_config` param is keyword-only with a `None` default (`simulation/api.py:269`, `simulation/experiments.py:82`), and the internal helpers that lost a required `model` arg (`_generate_datapoints_inner`, `_resolve_or_generate_datapoints`, `_generate_personas_scenarios`) all gained defaults rather than becoming stricter. `DatapointGenerator.__init__` and the three generator classes are `*`-only so old positional-free call sites are untouched.

`SimulationConfig.model` and `SimulationConfig.llm_config.model` (`simulation/_config.py:79-83`) are documented as always-equal but that's discipline at two construction sites, not a validator. Fine today, but nothing stops a third construction site from setting one and not the other later. Consider a `model_validator` or deriving `model` as a property of `llm_config.model` instead of duplicating the field.

`generate_structured`'s `config` merge (`common/structured_output.py:719-730`) is the one function every generator funnels through, and it has no direct unit test. `test_sim_llm_config.py` mocks `generate_structured` out everywhere it's exercised, and `tests/common/test_structured_output_usage.py` doesn't touch the new param either. The merge logic itself (five separate `if x is None and 'field' in set_fields` branches) is exactly the kind of thing that silently regresses under refactor with nothing to catch it.

Same file, line 729: `if timeout_s == _STRUCTURED_TIMEOUT_S and 'timeout_ms' in set_fields` uses value equality to detect "caller didn't pass timeout_s." If a caller explicitly passes `timeout_s=_STRUCTURED_TIMEOUT_S` (the default value, which is plausible since it's a public constant), the config's timeout_ms silently wins instead of the explicit kwarg, the doc comment says "an explicit keyword here still wins" but that's only true when the explicit value happens to differ from the default. Same shape of bug, not new, mirrors `_resolve_sim_llm_config`'s `sim_model != DEFAULT_MODEL` check in `simulation/api.py:118`, a real caller passing `sim_model=DEFAULT_MODEL` explicitly alongside a contradicting `llm_config` gets no warning. Both are edge cases, not correctness bugs (the richer config still wins in both), just silently-swallowed intent.

CLI (`simulation/cli.py`) isn't touched, `evaluatorq simulate`/`generate` still only expose `--sim-model`, no way to reach `llm_config`'s temperature/reasoning_effort/extra_body from the command line. That mirrors `redteam/cli.py`, which never exposed attacker/evaluator sampling knobs either, so it's not a new gap, but worth being explicit that "every simulation entry point" in the PR body means the Python API, not the CLI.

Nothing here blocks merge. The core threading and warning-on-contradiction logic is correct and well tested where it matters (`_resolve_sim_llm_config`, `from_call_config`, the runner, the generators' config plumbing, FirstMessageGenerator's temperature forwarding). Add a direct test for `generate_structured(config=...)`'s merge and I'd be comfortable calling this done.

Baukebrenninkmeijer:
Review decisions, for the next pass over this code.

### Decisions taken

**1. `llm_config.client` is honoured on every simulation path.**
The alternative was documenting that the field does nothing on this surface. It lost because a config field that is accepted and ignored is the same bug that was just fixed for `llm_config.model`, and the precedence rule already existed — `build_simulation_client` and `resolve_llm_client` both resolve an explicit client before the environment. Nothing new was added beside them; the config's client is simply handed over instead of consulted afterwards.

**2. `generate_structured` uses an `UNSET` sentinel rather than `None` for "not passed".**
The cheaper option was to correct the docstring, which overclaimed that an explicit keyword always beats `config`. It lost because `temperature=None` means "omit the parameter" — the thing a reasoning-class model needs — and `timeout_s=_STRUCTURED_TIMEOUT_S` is a public constant a caller can pass on purpose. Both were unreachable while `None` doubled as "said nothing". The sentinel is public (`UNSET` / `Unset`) because forwarding callers, `generate_recommendations` among them, need to express the same distinction.

**3. Recommendations take the run's config.**
The alternative was narrowing the CHANGELOG claim to exclude them. It lost because remediation suggestions are a simulation-side LLM call by any reading. `extra_kwargs=dict(llm_kwargs or {})` was also passing an empty dict unconditionally, which would have beaten the config even after threading it.

**4. The CLI stays as it is — deferred, deliberately.**
`eq simulate` exposes only `--sim-model`. `eq redteam` has the identical gap, so exposing sampling knobs is one change across both surfaces, not a rider on this PR. The CHANGELOG now says that "every simulation entry point" means the Python API.

**5. `sim_model=` deprecation is tracked, not done here.**
[RES-1424](https://linear.app/orqai/issue/RES-1424/deprecate-sim-model-in-favour-of-llm-config-on-the-simulation-entry) / #176. It is blocked on a public signature change: `resolve_sim_llm_config` treats `sim_model == DEFAULT_MODEL` as "not passed", so a `DeprecationWarning` on any use of `sim_model` needs `sim_model: str | None = None` or a sentinel first. That is the same problem decision 2 solved one layer down, and the ticket says to follow that pattern.

### Also applied from the review

- `llm_config.model` was silently ignored by the four trace/experiment entry points — they passed both a raw `model=` and `config=llm_config` to `generate_structured`, which deliberately does not read `config.model`. `extend_from_traces` could summarize on one model and generate on another in a single call.
- The resolver moved to `simulation/_config.py` so every entry point reaches it instead of re-implementing the fold, and it takes the caller's name — the warning said `simulate()` no matter who raised it. Its loguru format string used `%s` placeholders, which loguru does not interpolate, so it printed the raw template.
- The "read only the fields the caller set" filter existed in three hand-rolled copies and the third was incomplete: the executive summary dropped `max_tokens`, `timeout_ms`, `extra_body`, `extra_kwargs`. Now `LLMCallConfig.set_values`.
- `AgentConfig.from_call_config` returns `Self`, not `Any`.
- First direct tests of the `generate_structured` config merge and of trace-helper model resolution — the gap that let the model bug through.

### Rejected

- **The three `model` params on the `api.py` internals are dead.** They are not; existing tests pass them.
- **`SimulationConfig.model` and `.llm_config.model` should be validated equal.** Two construction sites, both in one function; a validator here would be a guard against a caller that does not exist.

src/evaluatorq/common/structured_output.py
Baukebrenninkmeijer: Fixed properly rather than documented. `temperature`, `extra_kwargs`, `extra_body`, `reasoning_effort` and `timeout_s` now default to a public `UNSET` sentinel (`common/structured_output.py`), so `None` and `timeout_s=_STRUCTURED_TIMEOUT_S` are values a caller can mean. You were right that the docstring was overclaiming — and it mattered more than a doc fix could cover, because `temperature=None` is exactly what a reasoning-class model needs and it was unreachable while `None` doubled as 'said nothing'. The fold moved into `_fold_config`, which states the precedence once. Two tests cover exactly your case: `test_generate_structured_explicit_none_temperature_beats_the_config` and `test_generate_structured_explicit_default_timeout_beats_the_config`.

src/evaluatorq/common/structured_output.py
Baukebrenninkmeijer: Added, in `tests/common/test_structured_output_config.py` as you suggested — four cases passing a real `LLMCallConfig` and asserting the kwargs that reach the executor: config fields applied, `config.model` deliberately not read, explicit keyword wins, explicit `None` wins. This gap was not hypothetical: it is how the model bug below got through. `traces.py` was passing both a raw `model=` and `config=llm_config` to `generate_structured`, which by design ignores `config.model`, so `datapoints_from_traces(llm_config=LLMCallConfig(model='x'))` applied the temperature and silently ran on `DEFAULT_MODEL`. `extend_from_traces` could summarize on one model and generate on another in a single call. Fixed in b0c0ac6-onward; all four trace/experiment entry points now resolve first and thread the resolved model down.

src/evaluatorq/simulation/_config.py
Baukebrenninkmeijer: Done — `model` is now a property returning `self.llm_config.model`, and both construction sites stopped setting it. Nothing to keep in step, so a third site cannot drift.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Correct, and it is now the one thing here I have deliberately not fixed. Making the resolver able to tell 'not passed' from 'passed the default' means changing the public signature to `sim_model: str | None = None` (or a sentinel, as `generate_structured` now uses one layer down), which is out of scope for this PR. Tracked as [RES-1424](https://linear.app/orqai/issue/RES-1424/deprecate-sim-model-in-favour-of-llm-config-on-the-simulation-entry) / #176, where it blocks the `sim_model=` deprecation warning for the same reason. The resolver also moved to `simulation/_config.py` and now takes the caller's name — it used to say `simulate()` no matter who raised it, and its loguru format string used `%s` placeholders, which loguru does not interpolate, so it was printing the raw template.

Baukebrenninkmeijer:
All four of your inline comments are addressed and the threads are resolved — thanks, two of them found real problems rather than nits.

**The timeout/`None` ambiguity you flagged was the bigger one.** You called it "not a blocker, just worth a comment noting the limitation", but the limitation turned out to be unreachable behaviour, not just an inaccurate docstring: `temperature=None` means "omit the parameter", which is exactly what a reasoning-class model needs, and it could never override a configured temperature while `None` also meant "caller said nothing". So `temperature`, `extra_kwargs`, `extra_body`, `reasoning_effort` and `timeout_s` now default to a public `UNSET` sentinel and the fold lives in one place, `_fold_config`.

**The missing test found a live bug.** Writing the direct merge test you asked for is what surfaced that `traces.py` was passing both a raw `model=` and `config=llm_config` to `generate_structured`, which by design ignores `config.model` — so `datapoints_from_traces(llm_config=LLMCallConfig(model="x"))` applied the temperature and silently ran on `DEFAULT_MODEL`, and `extend_from_traces` could summarize on one model and generate on another in a single call. All four trace/experiment entry points now resolve first and thread the resolved model down.

`SimulationConfig.model` is a property now. The `sim_model=DEFAULT_MODEL` diagnostic gap is the one thing left open on purpose — it needs a public signature change, tracked in [RES-1424](https://linear.app/orqai/issue/RES-1424/deprecate-sim-model-in-favour-of-llm-config-on-the-simulation-entry) / #176 where it blocks the `sim_model=` deprecation for the same reason.

Two other things landed alongside, from a separate review pass — worth a look since they are behaviour changes rather than cleanups:

- `llm_config.client` was accepted and ignored everywhere. A caller supplying their own `AsyncOpenAI` still failed for want of environment credentials it was never going to use. `build_simulation_client` / `resolve_llm_client` already owned the precedence rule, so the fix is handing them the config's client rather than adding a second rule beside them.
- `generate_recommendations` takes `llm_config` and forwards it. Its `extra_kwargs=dict(llm_kwargs or {})` was also passing an empty dict unconditionally, which would have beaten the config even after threading it.

The CLI stays as it is, deliberately — `eq redteam` has the identical gap, so exposing sampling knobs is one change across both surfaces rather than a rider here. The CHANGELOG now says "every simulation entry point" means the Python API.

CI is green across all 12 checks. @currentlycodinng ready for another look when you have a moment.

currentlycodinng (approved):
Re-traced all three fixes against current code at f28ecf80, not just the commit messages.

d6c2598 is a real fix, not just cleanup. Before it, summarize_conversations/datapoints_from_traces/extend_from_traces/extend_from_experiment took model= and llm_config= as two unreconciled params. Since generate_structured ignores config.model by design, a caller passing both got the config's temperature and silently ran on DEFAULT_MODEL. extend_from_traces could summarize and generate on two different models in one call. Now everything resolves through resolve_sim_llm_config in _config.py:165 first. Good catch, good fix.

65fb6a3 fixes the exact bug shape I flagged at structured_output.py:729. UNSET sentinel replaces the timeout_s == _STRUCTURED_TIMEOUT_S value-equality check, so passing the default on purpose now actually wins over config, like the docstring always claimed. New test at tests/common/test_structured_output_config.py:106 covers it directly. llm_config.client also now reaches the client builders instead of the environment winning first, which was a real "your own client fails for missing env creds" bug.

f28ecf8 does what I suggested: SimulationConfig.model is a property now (_config.py:91-99), derived from llm_config.model, not a second field held in sync by discipline.

generate_structured's config merge also got direct tests it didn't have before (tests/common/test_structured_output_config.py), which was my main ask.

Only outstanding item: sim_model != DEFAULT_MODEL at _config.py:192 still uses the same value-equality shape I flagged in api.py:118 before. Untouched, still an edge case, still not blocking.

Ran test_sim_llm_config.py, test_structured_output_config.py, test_structured_output_usage.py, test_structured_output.py, 94 passed. Full simulation+common suite is green except 8 unrelated failures from a missing fasthtml dependency in my environment.

Approving.

PR #175, opened 2026-08-24
fix(tracing): render Responses spans as transcripts and trace deployment targets

Baukebrenninkmeijer:
## Problem

Every Responses-API call evaluatorq makes showed up in the Orq trace UI as a raw JSON tree instead of a message transcript, and some target calls left no LLM span at all.

Orq's OTLP ingest derives a span type from `gen_ai.operation.name` through a fixed table that knows `chat` but not `responses`. An unmapped operation falls through to span-name heuristics and lands on `span.generic`, which the detail pane renders as JSON. Orq's own Go emitters canonicalize `responses` → `chat` at emit time, which is why server-side calls look right and ours did not.

## Fix

A client-supplied `orq.span_type` overrides the ingest derivation — the same mechanism evaluation spans already use. All three `with_llm_span` builders (`common`, `openresponses`, `simulation`) now set `span.responses` for `responses` and `agents.responses` operations, via one shared helper.

Responses spans also emit the unprefixed `openresponses.input` / `openresponses.output` / `openresponses.instructions`. The transcript renderer understands raw Responses items; the `gen_ai.input.messages` path it falls back to keeps only role-bearing ones, so `function_call`, `function_call_output` and `reasoning` items vanish from the view.

**Verified live**, not just in tests: a real simulation run's spans arrive stored as `span.responses`.

## Two simulation target-path bugs

Both found while chasing target calls that produced no LLM span:

- **`from_orq_deployment` emitted no span.** It was the one target flavour with no tracing at all — `deployment()` has none, while the red-team deployment leg wraps its call. Now wrapped the same way.
- **`SimulationRunner(target=<AgentTarget>)` failed every turn silently.** An `AgentTarget` passed as `target=` (rather than `target_agent=`) got wrapped in `CallableTarget`, raising `TypeError: not callable`. `call_target_with_retry` caught it, retried twice and returned a synthetic `[ERROR: ...]` response — so the run *completed*, reporting `terminated_by=error`, with no span and no HTTP call ever made. It now routes to `target_agent`.

## Trace context propagation (RES-1422)

Two target legs sent no `traceparent` at all, so their server-side execution started its own root trace instead of nesting under the target-call span:

- **`ORQAgentTarget`** — now passes the headers on `agents.responses.create`.
- **The deployment legs** — `deployment()` (which `invoke()` and the simulation adapter `from_orq_deployment` both go through) and the red-team static deployment job now pass them on `deployments.invoke_async`.

RES-1422 claimed this needed an SDK affordance. It did not: the generated Orq SDK takes `http_headers` on every operation, `deployments.invoke_async` included. It was a call-site fix.

Propagation is also toggleable now: **`EVALUATORQ_PROPAGATE_TRACE_CONTEXT`** (default `true`), gated in one place — `common.tracing.get_trace_context_headers()` — so every call site inherits it. Documented in `docs/configuration.md`, `docs/tracing.md`, `CHANGELOG.md` and `CLAUDE.md`.

Tests assert the outgoing kwargs (injection and toggle-off) on each leg; the nesting itself is not verified against a live deployment.

## Tests

New `tests/simulation/test_target_span_emission.py`. Every simulation test used a plain callable target, so nothing here was guarded — a test asserting `orq.simulation.target_call` exists passes even when the inner Responses span, its attributes and the traceparent all vanish.

## Note on uv.lock

One line, catching up to 44307f64, which changed the `openai` bound in `pyproject.toml` without re-locking.


src/evaluatorq/openresponses/tracing.py
```diff
@@ -154,3 +113,9 @@ def record_openresponses_response(span: Span | None, response: Any) -> None:
             'orq.openresponses.response',
```
currentlycodinng, line 123: When model_dump() fails and we fall back to {'repr': repr(response)}, payload.get('output') is always None here, so openresponses.output silently never gets set on that path. Probably fine (it's already a best-effort diagnostic dump) but worth a comment so it doesn't look like a bug on the next pass.

src/evaluatorq/common/tracing.py
```diff
@@ -412,9 +412,32 @@ def record_llm_output(span: Span | None, output: str) -> None:
     span.set_attribute('output', serialized)
 
 
+def orq_span_type_for_operation(operation: str) -> str | None:
```
currentlycodinng, line 428: operation.endswith('.responses') is exhaustive for every call site in the repo today (grepped all operation= usages: only 'responses' and 'agents.responses'). Just flagging that this is coverage-by-audit, not a documented contract, if a new caller introduces e.g. 'responses.stream' this still matches, but 'response' (no s) or 'Responses' (case) would silently fall through to span.generic again.

currentlycodinng (approved):
Checked out 02b7c5d4 and diffed against the actual merge-base, not the stated base, the branch wasn't rebased onto 2dd2b4fe so a straight two-dot diff pulls in ~1800 lines of unrelated main history. Real PR content is the 15-file, +368/-216 diff.

Shared helper checks out: orq_span_type_for_operation (common/tracing.py:415) lives once, both openresponses.tracing.with_llm_span and simulation.tracing.with_llm_span now delegate to common.with_llm_span instead of carrying their own copy. Grepped every operation= call site in the repo, only 'responses' and 'agents.responses' are ever passed, so the matcher is exhaustive today.

from_orq_deployment (simulation/adapters.py:40) wraps the deployment call the same way redteam/runtime/jobs.py:104 does, same operation, same provider, same purpose tag. AgentTarget routing fix (simulation/runner/simulation.py:654) is sound: AgentTarget is an ABC with no __call__, so there's no real collision case for a bare callable getting misrouted. Both-supplied case is handled with a warning and target_agent wins.

Ran tests/simulation/test_target_span_emission.py, 5/5 pass, and they actually assert the fixed behavior (span type, provider, effective target identity), not just "a span exists." Ran the full suite: 4282 passed, 12 failed, and the 12 failures are identical on base 2dd2b4fe, missing fasthtml/huggingface-hub in this venv, nothing to do with this PR. uv.lock diff is exactly the one line described.

One thing not called out in the PR description: the second commit also normalizes the Azure provider name to azure.ai.openai (semconv-correct spelling) across every red-team/simulation/openresponses span, not just Responses ones. It's tested and correct, just wider blast radius than the title suggests. Worth a mention for anyone diffing behavior later, not a blocker.

Checks out. Approving.

src/evaluatorq/openresponses/tracing.py
Baukebrenninkmeijer: Right, and the repr fallback also loses `gen_ai.output.messages`, not just `openresponses.output`. Rather than a comment, 7219c168 bumps the `model_dump` failure from `logger.debug` to `logger.warning` naming both attributes it costs — the house rule is that a degraded path announces itself — and adds a test covering that branch.

src/evaluatorq/common/tracing.py
Baukebrenninkmeijer: Agreed, and leaving it as-is deliberately. All six call sites pass a literal we control (`openresponses/target.py:256`, `redteam/backends/orq.py:273`, `simulation/agents/base.py:564`, `simulation/generators/first_message_generator.py:132`, `common/judge.py:585`, `common/structured_output.py:308`), and `tests/simulation/test_target_span_emission.py:35` pins the mapping. A looser match (case-folding, prefix matching) would start accepting names we never emit, which trades a visible gap for a silent one — if a caller ever needs a new operation name the fix is to add it to that test.

PR #179, opened 2026-08-28
docs: unwrap every hard-wrapped markdown file

Baukebrenninkmeijer:
CLAUDE.md has said *do not hard-wrap prose you are writing* for a while, but 42 of the 71 tracked markdown files were still wrapped at roughly 80 columns. Every new author matched the surrounding file, so the rule never took. This unwraps all of them: one line per paragraph, list item, admonition body, and blockquote.

Left verbatim: fenced and indented code, tables, YAML frontmatter, headings, horizontal rules, reference link definitions, and any line ending in a two-space hard break. Admonition bodies are joined but keep their four-space indent.

## Why bother

The point is the next diff, not this one. Reword a sentence in the middle of a wrapped paragraph and the reflow repaints every line below it, which buries the actual change in the review and collides with parallel sessions editing the same file.

## Verification

Whitespace only, checked two ways.

1. Normalising every changed file to a single whitespace run gives text identical to `HEAD`. 53 files, zero content differences, so no character moved except a line break.
2. Building the site before and after and comparing the rendered text of all 97 pages gives zero differences.

`mkdocs build --strict`, `scripts/validate_mermaid.py`, `ruff check src`, `ruff format --check src`, `basedpyright` (0 errors) and `pytest -m 'not integration'` (5141 passed) are all green.

## Not in this PR

The docs tree names `gpt-4o-mini` in about thirty places. A companion change on #165 adds a CLAUDE.md rule requiring examples to name a current model (`orq models list --json` carries a `created` timestamp, so the latest OpenAI family is one query away), but the sweep itself is a content change and would ruin the whitespace-only verification above.

### DIFF
--- a/.claude/skills/docs-coverage/SKILL.md
+++ b/.claude/skills/docs-coverage/SKILL.md
@@ -5,21 +5,15 @@ description: Use when checking whether evaluatorq docs cover every usage path �
 
 # docs-coverage
 
-The code supports combinations of things. This skill finds combinations a user could
-plausibly hit that no page explains.
+The code supports combinations of things. This skill finds combinations a user could plausibly hit that no page explains.
 
-Not a symbol checklist. "`red_team` is mentioned somewhere" is not coverage — the
-interesting question is whether *static mode against a `deployment:` target* is
-explained anywhere.
+Not a symbol checklist. "`red_team` is mentioned somewhere" is not coverage — the interesting question is whether *static mode against a `deployment:` target* is explained anywhere.
 
-Companion skill: `docs-drift` answers the opposite question — what the docs claim
-that is no longer true.
+Companion skill: `docs-drift` answers the opposite question — what the docs claim that is no longer true.
 
 ## Axes
 
-Read `axes.md` (next to this file) first. It defines the axis *names*, the tiers, and
-the impossible-combination list. Axis *values* come from source at run time, so new
-modes and backends appear automatically.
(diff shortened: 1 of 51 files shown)


arianpasquali (approved):
Verified the whitespace-only claim independently rather than taking the PR body's word for it: across all 53 files, normalising every whitespace run gives text identical to `origin/main`; fenced code blocks, table lines, two-space hard breaks and YAML frontmatter are byte-identical; admonition and content-tab markers keep their bodies and indentation; and rendering both versions through python-markdown (fenced_code, tables, admonition, attr_list, def_list, toc, md_in_html) is equivalent on 53/53 files, with the only delta a trailing space inside the hidden `<div markdown>` in `docs/index.md`. The wrapped-file count drops from 50 to 5. Content risk is nil.

Three follow-ups, none blocking:

Raw HTML blocks were joined too, which the stated scope did not cover. The cost-calculator `<form>` in `docs/guides/red-teaming.md` now carries its structural tags on sibling lines (`</div> <label ...>`, `</div> </form>`), and the README `<p align="center">` badge row is a single ~900-character line. Rendered output is unchanged, but that source is harder to edit than it was before, which is the opposite of what this PR is for. Restoring the `<form>` block by hand would be worth it.

"Every" is a slight overstatement: `=== "tab"` content-tab bodies are still wrapped in five files, including `docs/guides/agent-simulation.md` and `docs/guides/red-teaming.md`. The script handles `!!!` but not `===`.

The one worth doing soon is a mechanical guard. Those 42 files stayed wrapped because the rule lived only as prose in CLAUDE.md, and nothing here stops that recurring.

PR #180, opened 2026-08-31
docs: document wrap_simulation_agent()

Baukebrenninkmeijer:
## The gap

`gap: entry point (wrap_simulation_agent()) × surface (Python API)`

It ranked first because it is the seam between the two largest surfaces in the package and nothing hand-written described it. `wrap_simulation_agent()` turns a simulation into an evaluatorq job, so a multi-turn conversation becomes a scored row in an ordinary `evaluatorq()` batch. `docs/guides/agent-simulation.md` is 882 lines and contained zero occurrences of the symbol or of the idea; the only coverage on the site was the generated example page for `examples/agent_simulation/05_wrap_and_experiment.py` and the generated API reference, and `axes.md` is explicit that generated pages never satisfy Tier 1.

The two other Tier 1 axes came back clean this week, and the matrix in `.context/docs-coverage-matrix.md` records why so next week does not re-derive them. Every environment variable read by `src/` is documented — five names that look undocumented to a naive grep (`ORQ_DEFAULT_HOST`, `ORQ_ROUTER_SUFFIX`, `ORQ_SCALE_AGENT`, `ORQ_SCALE_GOOD_BAD`, `ORQ_SCALE_HEAT`) are module constants rather than environment variables, reachable by a name-shaped search only through their `__all__` entries. Every CLI command and subcommand is documented too.

## Read this section first if you reviewed an earlier version

The routine opened this PR, and review since has changed it enough that the original body was describing a page that no longer exists. This section says what moved; the rest of the body is kept current rather than left as a record of the first draft.

**The routine's receipt was not trustworthy when this PR was opened.** `run_receipt.sh` never checked whether its block extractor succeeded: `index.tsv` was then never written, the loop read nothing, `FAILED` stayed `0`, and the script still printed `RESULT: OK — every executable block exited 0` and exited 0. A page that does not exist, a `uv` failure, or a page with no fenced blocks all produced a green receipt having run nothing. Since `SKILL.md` makes that file the routine's hard gate, that is the one failure mode it could not have.

Fixed in `d83d7afe`, and to be precise about how, because an earlier version of this section was vague enough to be read as claiming otherwise: the fix is a `die()` on extraction failure plus a counted `SEEN`/`EXTRACTED` assertion at `run_receipt.sh:139`, not the addition of `set -e`. The script still runs under `set -uo pipefail`. `-e` would have been the wrong fix here — line 136 is `[ "$code" -ne 0 ] && FAILED=1`, which evaluates false on every *passing* block, so under `-e` the script would exit on the first block that worked.

**Four claims on the page were wrong, and a reader would have acted on each** (`bc4685f5`): `terminated_by` was documented as three values when the enum has a fourth, `timeout`, which `scorers.py` groups with `error` as unevaluated — so a reader branching on the documented three scored a timed-out run as a normal completion. `to_open_responses()` copies neither `criteria_verified` nor `criteria_meta`, so on this path a criterion that passed because the judge audited it is indistinguishable from one that passed because the judge never mentioned it. The JSON-string coercion claim held for three of the four input shapes, not all four. And the CI workflow installed `evaluatorq[simulation]`, which ships no OpenTelemetry, while the page promised spans.

**The CI section documented two gating mechanisms side by side** (`3f41172d`): a hand-rolled `gate()` / `scores_from()` pair walking the result shape by hand, and `check_pass_failures` linked four lines later as the alternative. The hand-rolled half reimplemented a shipped helper. It is gone; the gate is `check_pass_failures(results, treat_errors_as_failure=True)`, and `criteria_scorer` sets `pass_` so the threshold lives in one place.

**Docs samples now have a CI gate of their own** (`bfb921f6`): `check_examples.py` covered `examples/` and nothing else, so a renamed evaluatorq symbol could reach a published page with no signal. It now walks every fenced `python` block under `docs/` too, taking the checked import count from 225 to 410.

**Two errors in the input-shapes section** (`89d12953`, from @currentlycodinng's review): the precedence table was in the reverse of the order the code checks, while telling the reader the first shape listed wins — so the one sentence that exists to resolve an ambiguous row pointed the wrong way for it. And the `datapoint` row named three required fields where `SimulationDatapoint` requires five; a dict built from the table passed the shape check and raised in `model_validate` instead. That section is the one part of the page no runnable block covers, which is why both survived the receipt.

## What the page covers

`docs/guides/simulation-in-evaluatorq.md`, new, under Guides → Agent Simulation → In an evaluatorq Run:

- When to reach for this instead of `simulate()`, as a table rather than prose.
- A complete runnable script: local callable target, persona and scenario with criteria, an evaluator that sets both `value` and `pass_`, and `aclose()` in a `finally`. It checks `terminated_by` before it finishes, because evaluatorq reports a 100% success rate for a run whose conversation never happened — the job returned a result dict, so nothing errored from its point of view.
- A parameter table for all seven parameters, including `max_turns`'s default of 10 and the fact that `model` sets the simulator's and judge's model rather than the agent's.
- The `metadata` contract the job returns, key by key, with the optional keys marked and `terminated_by`'s unevaluated values called out.
- Why `goal_achieved` and `criteria_results` can disagree.
- A warning that `rules_broken` is normally id-keyed while `criteria_results` is description-keyed, and that `rules_broken` degrades to free text when the judge's audit does not arrive.
- A warning that `criteria_verified` and `criteria_meta` do not survive onto this path, so an audited pass and an unmentioned one look identical.
- The deployment-backed variant, and that constructing the job issues no request, so a wrong key surfaces on the first conversation rather than at import.
- The accepted `inputs` shapes in the order the code checks them, the full required-field set for each, and which of them coerce JSON strings.
- A CI section: `check_pass_failures` with `treat_errors_as_failure=True`, what `simulate()`'s `exit_on_failure` does not cover, and a complete workflow file installing `evaluatorq[simulation,otel]`.
- What the run uploads, and that no public parameter turns it off.

**Deliberately not covered.** The `eq sim` CLI surface, which has its own reference page and its own upload behaviour; the built-in simulation scorers, which take a `SimulationResult` and so are not usable on this path (the page says so rather than implying it by omission); and a worked example against a live Orq deployment, because no deployment key was available and an unrunnable block does not go in a page.

## Receipt

Regenerated on the current head, against a fresh `EVALUATORQ_DIR`, from the fixed `run_receipt.sh`. Every block on the page is accounted for, which is now asserted by count rather than assumed.

| block | lang | exit | first output lines |
|---|---|---|---|
| 001 | bash | 0 | `Resolved 241 packages` / `Checked 227 packages` |
| 002 | python | 0 | three `on_turn_complete` turns, then `✓ Evaluation completed` |
| 003 | console | — illustrative (not run by this runner) | — |
| 004 | python | 0 | (no stdout — constructs the deployment-backed job) |
| 005 | python | 0 | `gate helper imported: check_pass_failures` |
| 006 | yaml | — illustrative (not run by this runner) | — |
| 007 | console | — illustrative (not run by this runner) | — |

`RESULT: OK — 7 blocks accounted for, every executable one exited 0`.

The illustrative rows read "not run by this runner" rather than "not executable by nature": whether a console or yaml block *could* have been executed is a claim for a reviewer to make, not one the script is in a position to assert.

Also green on the merged tree: `mkdocs build --strict`, `scripts/validate_mermaid.py`, `check_examples.py` (410 imports, 0 failed), `ruff check src`, `ruff format --check src`, `basedpyright` (0 errors), and `pytest -m 'not integration'` (5171 passed, 2 skipped).

**What the receipt does not cover.** It proves every fenced block runs and exits 0. It says nothing about the prose, and nothing at all about a section with no runnable block — which is exactly where both of the review's page findings were.

## Persona verdicts

From the routine's own review rounds, before the corrections above. A record of what those reviewers found at the time, not a claim about the page as it now stands.

|---|---|---|---|
| First-timer | Get a scored result for whether a support chatbot holds up against an impatient customer demanding a refund | COMPLETED | 1 |
| Practitioner | Fold `simulate()` conversations into an existing `evaluatorq()` batch, scoring criteria and goal separately | COMPLETED | 2 |
| Platform user | Run simulated conversations against a deployed agent, land them as an Experiment, and know what that puts in a shared workspace | COMPLETED | 3 |
| CI engineer | A GitHub Actions workflow that fails the build on a scored regression, non-interactive, no state leakage | COMPLETED WITH GUESSWORK | 3 |

The CI engineer's residual guesswork was that `CommunicationStyle` members are listed nowhere on the site — it guessed `.detailed`, got an `AttributeError`, and worked around it. The members are now named in the example.

Two caveats on reading that table. The first-timer solved its task from the existing overview guide using `simulate()` without needing this page, which is the right outcome for a single agent and no batch but means its COMPLETED is not evidence about this page. And the platform user's round-2 report was contaminated: all five reviewers shared one scratch directory, so the adversarial critic's `git diff` was visible to personas required not to read it, and that persona read part of it before catching itself and disclosing. Round 3 gave each persona an isolated directory and the COMPLETED above is from that clean run. Handing the critic a diff while the personas share its working directory was a flaw in the routine's harness, not in the reviewers.

## Code and docs inconsistencies found, and not fixed here

The routine may not touch `src/`, so these are reported rather than repaired. Each was verified against source.

1. **`upload_results=False` is documented as a "local-only" run and is not one.** `src/evaluatorq/common/model_catalogue.py:313` fetches the model catalogue for cost pricing from the Orq host whenever `ORQ_API_KEY` is set, ungated by that flag. The docstring is `src/evaluatorq/simulation/api.py:315` and :322, and it reaches users through the generated API reference. The sibling guide's admonition has been corrected on this branch; the docstring has not.
2. **A debug line claims an upload on runs that upload nothing.** The log lines are `logger.debug('Results upload host resolved from ...')` at `src/evaluatorq/common/llm_client.py:99` and `:102`, inside `resolve_results_base_url`. `src/evaluatorq/simulation/api.py:2312` is the unconditional *call* to that helper, sitting above the `_send_results=upload_results` gate at `:2321` — it computes a string and issues no request, but the helper it calls logs on every run regardless of whether anything is uploaded. (An earlier version of this entry attributed the log line itself to `api.py:2312`; corrected per review.)
3. **`docs/cli-reference/simulation.md:17` shows a command that fails in a common configuration.** `eq sim run --agent-description "..." --openai-model gpt-4o-mini`. `src/evaluatorq/common/llm_client.py:16` gives `ORQ_API_KEY` precedence over `OPENAI_API_KEY`, so with an Orq key present that bare model id goes to the router, which cannot resolve it. Re-verified directly against the router on the current head: `gpt-4o-mini` returns `AuthenticationError` **401** (`Access denied due to invalid subscription key or wrong API endpoint`), while `openai/gpt-4o-mini` succeeds on the same client. That page also names `gpt-4o-mini`, which `CLAUDE.md` rules out for new examples; it is one of roughly sixteen such occurrences across existing pages.
4. **The CLI has no upload switch and does not say so.** `eq sim run` and `eq sim simulate` upload whenever `ORQ_API_KEY` is set, and the CLI reference documents no flag to suppress it and never states the default.
5. **Enum members are invisible on the site.** `CommunicationStyle`, `StartingEmotion` and `EmotionalArc` render on the generated API pages as `Bases: StrEnum` and nothing else, because mkdocstrings is not configured to show enum values. Named inline on this page as a local workaround; the general fix is an mkdocstrings setting.

## Not addressed

- Findings 3 and 4 live on a page this PR does not otherwise touch. Editing them here would be a drive-by change that hides this diff's subject.
- `bt_sigma_aggregation()` appears in no prose, but its whole output is documented as `report.bt_sigma.*` in `pairwise-judging.md`. Tier 2.
- `API endpoint × red_team()` remains a known gap recorded in `axes.md:32`. Documenting a field the call sites render inert would document a defect as though it were a design.

## Merge

`main` moved under this branch — #179 unwrapped every hard-wrapped markdown file, which collided with this branch's edit to the "CI and local runs" note in `agent-simulation.md`. Resolved by merging `main` in (no rebase, since this branch carries other people's commits) and keeping this branch's text, which was already unwrapped and additionally corrects "local-only run" to name what `upload_results=False` actually suppresses. Nothing from `main`'s side was lost: its only change to that note was the unwrapping.

## Workspace side effects

The routine's live blocks and its reviewers created roughly a dozen Experiments in the shared `orq-research` workspace, among them `support-simulation`, `support-simulation-ci`, `eval-poc-agent-sim-bauke` and `persona-platform-r3-support-simulation`. That is within the volume disclosed before the run started, but neither the SDK nor the CLI exposes a delete, so clearing them is a manual job in the Orq UI.

---

Opened by the weekly docs-autofill routine, and corrected by review since.


docs/guides/simulation-in-evaluatorq.md
```diff
@@ -0,0 +1,326 @@
...
+`wrap_simulation_agent()` reads one conversation per `DataPoint`, and accepts four spellings in `inputs`:
+
+| `inputs` contains | Notes |
+|---|---|
+| `persona` + `scenario` | The usual form. Add `first_message` to fix the opening line instead of generating it. |
+| `datapoint` | A full `SimulationDatapoint`, which must carry `persona`, `scenario` and `first_message`. |
+| `datapoints` | A list of exactly one datapoint. More than one raises. |
+| `personas` + `scenarios` | Lists of exactly one each. More than one raises. |
+
+The first matching shape in the order listed wins; a row carrying two of them silently uses the earlier one.
```
currentlycodinng: Order is wrong. _datapoint_io.py checks datapoint at :61, datapoints at :65, persona + scenario at :76, personas + scenarios at :90. I ran a DataPoint carrying both persona/scenario and datapoint and got the datapoint value back. Reorder the table to match the source, or this sentence sends a reader the wrong way in the exact case it exists to cover.

docs/guides/simulation-in-evaluatorq.md
```diff
@@ -0,0 +1,326 @@
...
+Put the `aclose()` in a `finally`, as the complete run above does, not after the `evaluatorq()` call. An exception raised mid-batch would otherwise skip it, which is exactly the run you least want leaking.
+
+## Accepted input shapes
+
+`wrap_simulation_agent()` reads one conversation per `DataPoint`, and accepts four spellings in `inputs`:
+
+| `inputs` contains | Notes |
+|---|---|
+| `persona` + `scenario` | The usual form. Add `first_message` to fix the opening line instead of generating it. |
+| `datapoint` | A full `SimulationDatapoint`, which must carry `persona`, `scenario` and `first_message`. |
```
currentlycodinng: Incomplete. _validate_shape checks those three at :63, but SimulationDatapoint.model_validate on the next line also requires id and user_system_prompt. Building the dict as documented raises a ValidationError with 8 errors. Same applies to the datapoints row below.

.claude/skills/docs-autofill/run_receipt.sh
```diff
@@ -0,0 +1,148 @@
...
+#   everything else (text, yaml, json, console, ...) -> recorded, never run
+#
+# Blocks run against a fresh EVALUATORQ_DIR so a page cannot pass by leaning on
+# a run store some earlier probe happened to populate.
+#
+# The OK verdict is a counted assertion, never a default: the script refuses to
+# claim success unless every extracted block was accounted for. A receipt that
+# ran nothing must not be able to say "every block exited 0".
+
+set -uo pipefail
```
currentlycodinng, line 23: Still set -uo pipefail, no -e. The PR description says the receipt bug was fixed by adding it. The real fix is the SEEN/EXTRACTED assertion at :139 plus the explicit FAILED tracking, which is better than -e would have been here (line 136 would exit under -e on every passing block). No change needed, just flagging that the description does not match its own patch.

currentlycodinng (changes requested):
Verified this independently against the head commit rather than taking the description's word for it. Most of it holds up.

The tooling numbers all reproduce: mkdocs build --strict exit 0, check_examples.py 410 imports 0 failed, basedpyright 0 errors, pytest -m "not integration" 5171 passed 2 skipped. ruff check/format clean on src, which is CI's actual scope (ci.yml:27,30). I also typo'd an import in the new page to check the new gate is not decorative: exit 1, named failure. Good.

All five "found, not fixed" claims are real. model_catalogue.py:313 does fetch whenever a key exists, ungated by upload_results. api.py:2312 does sit above the _send_results gate at :2321. llm_client.py:16 confirms ORQ_API_KEY wins over OPENAI_API_KEY, so the bare gpt-4o-mini at cli-reference/simulation.md:17 really does route to a router that cannot resolve it. No upload flag on eq sim. And I built the site to check the last one: CommunicationStyle renders as Bases: StrEnum and nothing else.

Two nits on those. The example fails on a bad model id, which is a 400/404, not the 401 the description claims. And the debug lines are at common/llm_client.py:99,102, not at api.py:2312, that line is the unconditional call.

On the receipt script: it is honest now. die() on extraction failure, the SEEN/EXTRACTED assertion at :139, per-block exit capture, exit "$FAILED". But line 23 is still set -uo pipefail with no -e. The description says the fix was adding -e. It wasn't. The counted assertion is the better fix, so no objection to the code, only to the description of it.

The metadata contract, the terminated_by enum, the criteria_verified/criteria_meta drop, the rules_broken keying, the check_pass_failures signature, the lazy agent_key, the enum member names, which field lives on Persona vs Scenario, all checked, all correct.

Two things the description missed, both in "Accepted input shapes", both in the one section no runnable block covers.

The precedence order is backwards. Source checks datapoint first (_datapoint_io.py:61), then datapoints (:65), then persona + scenario (:76). The table lists them in the opposite order and line 233 tells the reader the first one listed wins. I ran a row carrying both and got the datapoint value. That sentence exists to cover exactly this case and it points the wrong way.

And the datapoint row understates what is required. _validate_shape checks persona, scenario, first_message at :63, but model_validate on the very next line also needs id and user_system_prompt. A reader who builds the three-key dict the table describes gets a ValidationError with 8 entries.

Both are two-line fixes. Happy to approve once they land.

docs/guides/simulation-in-evaluatorq.md
Baukebrenninkmeijer: Fixed in 89d12953. Table reordered to the source order — `datapoint`, `datapoints`, `persona` + `scenario`, `personas` + `scenarios` — and the sentence below it now names the concrete case: a row carrying both `datapoint` and `persona` + `scenario` uses the `datapoint`. Added a line saying the table is in check order rather than in the order a reader is most likely to use, since `persona` + `scenario` is still the common form and leading with `datapoint` otherwise reads as a recommendation.

---

docs/guides/simulation-in-evaluatorq.md
Baukebrenninkmeijer: Fixed in 89d12953. Confirmed against the model rather than the shape check — `SimulationDatapoint.model_fields` reports all five of `id`, `persona`, `scenario`, `user_system_prompt`, `first_message` as required, none optional. Both the `datapoint` and `datapoints` rows now list all five.

I also kept your diagnosis in the row, because the gap between the two checks is the reason the failure is confusing: `_validate_shape` looks for three fields, so a dict missing `id` or `user_system_prompt` passes the shape check and then fails in `model_validate`, which is a different error in a different place from the one the reader was steered toward.

---

.claude/skills/docs-autofill/run_receipt.sh
Baukebrenninkmeijer: You are right, and thank you for reading the patch against its own description rather than taking it. The description was mine and it was wrong to leave "set `-uo pipefail` without `-e`" sitting next to "Fixed" — that reads as though `-e` was added, and it was not.

Your reason why it should not be is the sharper half, so I have put it in the body rather than just the correction: line 136 is `[ "$code" -ne 0 ] && FAILED=1`, which evaluates false on every passing block, so under `-e` the script would exit on the first block that worked. The description now names the actual fix — `die()` on extraction failure plus the counted `SEEN`/`EXTRACTED` assertion at :139 — and says the script still runs under `set -uo pipefail` deliberately.

No code change, as you said.

---

Baukebrenninkmeijer:
Both page findings are fixed in 89d12953 and answered on their threads. Receipt regenerated (7 blocks, every executable one exit 0), `mkdocs build --strict`, `validate_mermaid.py` and `check_examples.py` all green. Re-requesting your review.

Two things from your review body rather than the inline threads.

**The log-line location — you are right, and the body is corrected.** `api.py:2312` is the unconditional *call*; the `logger.debug('Results upload host resolved from ...')` lines are at `common/llm_client.py:99` and `:102` inside `resolve_results_base_url`. Finding 2 now cites all three lines and says which is which.

**The status code — I think this one is the other way round, and I would rather show the check than just assert it.** You read the failure as a bad model id giving a 400/404. On the router it comes back 401, because the request never gets far enough to be a model-resolution problem. Run against the current head:

```
base_url: https://my.orq.ai/v3/router/
'gpt-4o-mini':        AuthenticationError status=401 :: Error code: 401 - {'error': {'message': 'Access denied due to invalid subscription key or wrong API endpoint. ...
'openai/gpt-4o-mini': OK
```

Same client, same key, one call each. The persona reviewer that originally hit this got the same 401 with the same "invalid subscription key or wrong API endpoint" body, which is what the message is telling you: the unprefixed id does not address a route the router recognises, so it fails as auth rather than as a missing model. I have left finding 3 saying 401 and noted it was re-verified directly rather than carried over from the earlier report.

If you have a case where it does surface as a 400 or 404 — a different base URL, or the OpenAI-direct path where `OPENAI_API_KEY` wins — that would be worth knowing, since it would mean the status depends on which client resolved and the finding should say so.

---

currentlycodinng (approved):
Re-verified against 89d1295, not the commit title.

Precedence order: fixed. Table now matches source order (datapoint, datapoints, persona+scenario, personas+scenarios). Built a DataPoint carrying both datapoint and persona/scenario keys, ran it through _extract_single_datapoint, got the datapoint value back. Matches what the doc's precedence sentence now claims.

Datapoint row required fields: fixed. Row now lists all five fields (id, persona, scenario, user_system_prompt, first_message). Built the dict exactly as documented and it validated cleanly. Re-ran the old 3-field version as a sanity check and it reproduced the original ValidationError, so this isn't a coincidence.

Datapoints row: same fix applied, also confirmed by running the 5-field dict through the list form.

Diff is scoped to the one section, +6/-4 in the doc, nothing else touched. mkdocs build --strict, check_examples.py, and CI all green.

Approving.

Issue #167, opened 2026-08-24
Follow-ups from docs-autofill PR #165 (red-team replay): 5 open decisions


Baukebrenninkmeijer:
Filed so these survive [#165](https://github.com/orq-ai/evaluatorq/pull/165) closing — @currentlycodinng flagged in the PR review that its five "Open decisions" read like separate tickets and would get lost once the PR merges. The doc content in #165 is accurate either way (independently verified against source in that review); none of these block it.

Each keeps the PR body's **Lean**. Two need a human call by nature (#1 is a code change, #3 is a judgement call); the other three are mechanical once someone says go.

### 1. `mode` sentinel in `red_team()` silently swallows `mode="dynamic"` on replay
`previous_run=` rejects eight data-selection args by name but not the ninth. `mode` defaults to `Pipeline.DYNAMIC` (`runner.py:490`), so `Pipeline(mode) != Pipeline.DYNAMIC` (`runner.py:698`) can't tell "passed dynamic" from "passed nothing"; the value is then overwritten by the stored pipeline. `max_turns` two lines away uses exactly the sentinel that would fix it. `--mode dynamic` on the CLI has the identical hole.
**Lean:** `mode: Pipeline | str | None = None`, resolve to `DYNAMIC` after the conflict check, same for `cli.py`; add a test (`test_replay.py` currently asserts only the eight that raise). Non-breaking. Documented in a `!!! note` for now because the routine may not touch `src/`.

### 2. Replay documented in two places that will drift
`src/evaluatorq/redteam/README.md:101-128` already documents replay, and #165 had to backfill two facts from it (8+ char run-id prefix, newer-format rejection).
**Lean:** one source — move the shared fact-set into `docs/_snippets/` and transclude into both, or cut the guide's mechanics to a link. Needs `src/` scope.

### 3. Should `replay` be a peer bullet under `## Modes`?
Its own first clause says it isn't reached by `mode=`, unlike the other three bullets. `axes.md` models replay as a mode value for coverage bookkeeping; #165 carried that into reader-facing prose.
**Lean:** kept the bullet (discoverability was why the gap ranked first, and every review persona found the section) — but this is a judgement call about how the repo models replay.

### 4. `bt_sigma_aggregation` is a false-positive Tier 1 gap
[...]

PR #182, opened 2026-08-31
docs: make the red-team guide the single source for replay mechanics

Baukebrenninkmeijer:
Closes the last two open items on #167.

**Replay documented in two places (item 2).** `src/evaluatorq/redteam/README.md` and `docs/guides/red-teaming.md` both carried the reference forms `--from-run` accepts, what replay restores from the stored run, and the list of arguments it refuses. The guide has the fuller version, so the README now keeps the two invocations and links out for the mechanics.

**`bt_sigma_aggregation` false positive (item 4).** `.claude/skills/docs-coverage/axes.md` now says that a symbol whose *output* is documented under the name users actually type is not a Tier-1 gap, with `bt_sigma_aggregation` / `report.bt_sigma` as the standing example. Without this the coverage run re-derives it every time.

The other three items on #167 already landed: the `mode` sentinel is fixed in `runner.py` and `cli.py` with a test in `tests/redteam/test_replay.py`, replay is no longer a peer bullet under `## Modes`, and the docs-autofill receipt is a hard gate with the venv-on-`PATH` and fresh-run-store rules spelled out.

`uv run --group docs mkdocs build --strict` passes. No `src/` behaviour change.

### DIFF
--- a/.claude/skills/docs-coverage/axes.md
+++ b/.claude/skills/docs-coverage/axes.md
@@ -81,4 +81,6 @@ Marked `N/A` in the matrix, never reported as a gap.
 
 Tier-1 items are **not** matrix cells — there is no `env var` axis, and `surface` has exactly the three values above. Record a Tier-1 gap in the `docs-autofill` ledger as `tier 1: <kind> (<name>, …)` with no second axis, e.g. `tier 1: env var (ORQ_OTEL_MAX_QUEUE_SIZE, ORQ_OTEL_MAX_BATCH_SIZE)`. The ledger dedupe compares that cell as free text, so naming the items is what stops the same gap being re-derived next week; an invented axis pair never matches.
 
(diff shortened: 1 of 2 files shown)


currentlycodinng (approved):
Checked the diff against the actual replay code, not just against itself.

The reference forms, restore list, and refused-args list in docs/guides/red-teaming.md match resolve_run_path() in src/evaluatorq/common/replay.py and the conflicting check in runner.py:699-711 exactly. Nothing drifted. The README link at line 108 resolves and the anchor exists (red-teaming.md:245).

bt_sigma_aggregation / report.bt_sigma checks out too: build_report() assigns the result to report.bt_sigma in pairwise.py:654,671, and pairwise-judging.md:135-138 documents it under that name. One nit, not blocking: axes.md calls the symbol "internal," but it's in evaluatorq.__all__ (__init__.py:115), so it's actually a public export. Doesn't change the conclusion, just worth tightening if you touch this file again.

mkdocs build --strict passes clean, ran it myself. CI is green. Mode sentinel test, the Modes-section rewrite, and the docs-autofill hard gate are all present on this branch as claimed.

Approving.

PR #183, opened 2026-09-01
fix(redteam): refuse deployment targets in the adaptive pipelines, and rebuild the setup docs (RES-1492)

Baukebrenninkmeijer:
Closes RES-1492. Also picks up RES-1493 (the deployment-target guard). Real deployment support in the adaptive pipelines is RES-1494.

## Why

`docs/configuration.md` opened with a 25-row environment variable table, so a first-time reader met the reference before any setup instructions and got no signal about which few variables matter.

`docs/orq-deployment.md` was three unrelated topics stapled together — calling a deployment from a job, replaying a past experiment, and an API-signature dump duplicating the generated reference — and none of it was reachable from the pages people actually read.

The docs also promised something the code could not do: `red_team(target="deployment:<key>")` on the default pipeline 404s during context retrieval, before a single attack runs.

## What changed

**Code — deployment targets are refused, not 404ed.** `red_team()` with a `deployment:<key>` target now raises a `ValueError` naming `mode="static"` and `simulate()` as the two things that do work. The guard runs before any client, credential or paid call, in both `red_team()` and `_run_dynamic_or_hybrid()`, with a backstop arm in `_prepare_target`. A table in `redteam/contracts.py` is the single source of which target kinds each pipeline drives, so the refusal and the docs cannot drift apart. `eq redteam run --target` help now names the `--mode static` requirement.

**Configuration.** Opens with a two-step setup: a tabbed `.env` sample per backend (Orq / OpenAI / self-hosted), then how to get those variables into the process. Then a four-row table of the decisions that matter — backend, run store, tracing, dashboard links. The full reference follows, split into Backend, Runs and output, Dashboard, Tracing, and Simulation defaults. Deprecated `ROUTER_BASE_URL` row dropped.

**orq-deployment.md deleted, content redistributed.** Orq-hosted target material to `docs/guides/targets.md`, which named `agent:<key>` and `deployment:<key>` in a table and never explained either. The `deployment()`/`invoke()`-from-a-job material and the experiment-replay material to `docs/evaluation-reference.md`, beside the `@job` decorator and the data sources they belong with. The API-signature table and the duplicate env-var table are dropped — the generated reference and the configuration page already own them. `mkdocs.yml` drops the page from `nav` and `plugins.llmstxt.sections` and adds a redirect, so `/evaluatorq/orq-deployment/` keeps working.

**Two new Setup pages.** `docs/installation.md` and `docs/components.md` open the Setup section, so a reader arriving at the docs root learns what the extras install and what the pieces are before hitting a reference page.

**Two agent skills, harmonized.** `docs-writing` gains the reviewer loop's shared parts — the four reader personas and their verdicts, the artifacts-not-conclusions rule, and a plain-language table format for reporting a review in a PR. `docs-autofill` now invokes `docs-writing` for drafting and defers to it for the loop, keeping only what is specific to an unattended run (the adversarial `hate` track, the receipt check, the Slack DM). No behaviour outside `.claude/skills/` depends on either.

## Four wrong claims fixed

Three were inherited from the pages being reorganised, not introduced here:

- `report.summary.pass_rate` does not exist. The field is `resistance_rate` (`redteam/contracts.py:1773`), so the sample raised `AttributeError` after a completed run.
- `EVALUATORQ_DASHBOARD_ROOTS` was documented as colon-separated. It is a JSON array parsed with `json.loads` (`dashboard/launch.py:87`) — the documented format throws.
- `OPENAI_BASE_URL` was documented as honoured by red teaming *and* simulation. Simulation's conversation path resolves its client with `honor_openai_base_url=False` (`openresponses/client.py:50`), so a self-hosted simulation run silently went to OpenAI. Its post-run calls do honour it, and the row now says which is which.
- The first-run instructions called `load_dotenv()`, but `python-dotenv` is not a dependency — a clean install raised `ModuleNotFoundError`.

Also dropped the `uv add "evaluatorq[orq]"` instruction: `orq-ai-sdk>=4.12` is already a runtime dependency, so the extra is a no-op and the `ImportError` branch it described is unreachable in a normal install.

## Review

|---|---|---|
| `deployment:<key>` is not static-mode-only in the docs, but the default pipeline dies on it | @currentlycodinng | Fixed — the row says it, and the code now refuses with a clear error instead of 404ing |
| "Both string forms work identically" is the sentence that will bite someone | @currentlycodinng | Fixed — the sentence now says the `mode="static"` restriction applies in Python and on the CLI alike |
| The `OPENAI_BASE_URL` row overstates the exclusion; three post-run simulation paths do honour it | @currentlycodinng | Fixed — the row now says *conversation path*, and names the post-run calls that honour it |
| The `gpt-4o-mini` → `gpt-5.6-luna` swap leaves the rest of the site inconsistent | @currentlycodinng | Fixed — the whole site moved in #185; no `gpt-4o-mini` remains under `docs/` |
| The old page's anchors cannot survive a single redirect | @currentlycodinng | Left out — one stub cannot serve three destinations, and the page URL does survive |
| `ROUTER_BASE_URL` warning is live but now undocumented | @currentlycodinng | Left out — the value has been dead a while; deleting the warning is the real fix, and it is not this PR |

## Verification

`uv run --group docs mkdocs build --strict`, `ruff check src`, `ruff format --check src`, `basedpyright` (whole repo) and `pytest -m 'not integration'` (5247 passed) all clean locally. New tests in `tests/redteam/test_deployment_target_guard.py` cover the refusal on every pipeline.


docs/guides/targets.md
```diff
@@ -22,6 +22,21 @@ The CLI only accepts the string form — an object target is constructed in Pyth
...
 Context discovery matters more than it looks. Attack strategies are selected and written against the target's declared tools, memory and system prompt: a target that reports no tools never gets a tool-misuse attack, because those strategies are gated on `requires_tools=True`. See [Writing your own target](#writing-your-own-target) below.
 
+## Orq-hosted: agents and deployments
+
+If the thing under test already runs on Orq, you do not write a target class. How you name it depends on which entry point you are using:
+
+| You want to | Use |
+|---|---|
+| Red-team or simulate a hosted **agent** | `target="agent:<key>"` — the platform supplies tools, memory stores and knowledge bases to the attack planner |
+| Red-team or simulate a **deployment** (prompt + model) | `target="deployment:<key>"` |
```
currentlycodinng: This row is only true for simulate(). For red_team() it holds in static mode only, and static is not the default.

mode defaults to Pipeline.DYNAMIC (redteam/runner.py:733). The dynamic context prefetch loops over every string target with no kind check and calls resolve_context on the Orq agents backend (runner.py:2008-2013), which hits agents.retrieve(agent_key=<deployment key>) (backends/orq.py:494-501) and 404s. resolve_context catches only ValidationError and deliberately lets a 404 propagate (backends/base.py:81-90), so the run dies during context retrieval before any attack runs. deployments.invoke_async() is only reachable from the static leg (redteam/runtime/jobs.py:135 via runner.py:1461-1462).

Suggest splitting the row, roughly:

| Simulate a deployment | target="deployment:<key>" |
| Red-team a deployment | target="deployment:<key>" with mode="static". Dynamic and hybrid mode route deployment targets through the agents endpoint and fail. |

I will file the code-side ticket separately.

docs/guides/targets.md
```diff
@@ -22,6 +22,21 @@ The CLI only accepts the string form — an object target is constructed in Pyth
...
+| Call a deployment from an `evaluatorq()` **job** | `invoke()` or `deployment()` inside your `@job` — see [Evaluation Reference › Calling an Orq deployment from a job](../evaluation-reference.md#calling-an-orq-deployment-from-a-job) |
+| Pin the exact call parameters instead of discovering them | [`OrqResponsesTarget`](#orqresponsestarget) |
+
+Both string forms work identically in Python and on the CLI (`eq redteam run --target agent:my-key`).
```
currentlycodinng: "Both string forms work identically" is the specific sentence that will bite someone. On the CLI it is worse than in Python, because eq redteam run also defaults to dynamic while redteam/cli.py:253 advertises deployment:<key> in --target help. Please drop or qualify this line alongside the table row above.

docs/configuration.md
```diff
@@ -1,24 +1,101 @@
...
-| `OPENAI_BASE_URL` | No | OpenAI default | Redirect OpenAI-compatible calls to a different host (vLLM, OpenRouter, Azure, local). Honoured by the red teaming and simulation LLM client. |
-| `ORQ_DISABLE_TRACING` | No | unset | Set to `1` or `true` to suppress all OpenTelemetry spans even when `ORQ_API_KEY` or `OTEL_EXPORTER_OTLP_ENDPOINT` is present. |
-| `ORQ_DEBUG` | No | unset | Set to any non-empty value to print tracing setup diagnostics to stdout (endpoint, auth headers, initialization errors). |
-| `EQ_DEBUG` | No | unset | Set to any non-empty value to show the full traceback on CLI errors instead of the one-line message. CLI-wide; distinct from `ORQ_DEBUG`, which only affects tracing diagnostics. |
+| `OPENAI_BASE_URL` | No | OpenAI default | Redirect OpenAI-compatible calls to a different host (vLLM, OpenRouter, Azure, local). Honoured by the red teaming LLM client. **Not** honoured on the simulation path, which resolves its client with `honor_openai_base_url=False` — inject a pre-built client there instead. |
```
currentlycodinng: Nit, and only if you want the row exact. "Not honoured on the simulation path" is true for every conversational and generation call, but three post-run simulation paths call resolve_llm_client directly and so do pick up OPENAI_BASE_URL: simulation/api.py:93 and simulation/cli.py:255 (recommendations) and simulation/reports/executive_summary.py:106. Something like "not on the simulation conversation path" would cover it. Not worth holding the PR for.

docs/guides/targets.md
```diff
@@ -34,7 +49,7 @@ from evaluatorq.redteam import OpenAIModelTarget, red_team
 
 async def main() -> None:
     target = OpenAIModelTarget(
-        model="gpt-4o-mini",
+        model="gpt-5.6-luna",
```
currentlycodinng, line 52: This gpt-4o-mini -> gpt-5.6-luna swap (also lines 83 and 263) is not in the PR description, and it leaves this page inconsistent with tuning.md, faq.md, guides/red-teaming.md, guides/agent-simulation.md, cli-reference/simulation.md and llm-as-a-jury.md, which all still say gpt-4o-mini. Either revert it here or do the sweep in a separate PR so the whole site moves together.

currentlycodinng (changes requested):
Answering your three open questions, since that is what was asked.

deployment:<key> in dynamic mode: confirmed broken, and the docs need to say so. Your reviewing agent was right and the picture is a bit worse than "never reaches invoke_async()". mode defaults to Pipeline.DYNAMIC (redteam/runner.py:733). _run_dynamic_or_hybrid prefetches context for every string target with no kind check at all (runner.py:2008-2013), which lands in ORQAgentTarget.get_agent_context() calling agents.retrieve(agent_key=<deployment key>) (backends/orq.py:494-501). Backend.resolve_context catches only ValidationError, and the comment there says a 404 must fail loudly (backends/base.py:81-90), so the run dies during context retrieval before a single attack. Past that, _prepare_target sends everything that is not AGENT to the Orq agents backend (runner.py:1621-1626). deployments.invoke_async() has one call site in the package, redteam/runtime/jobs.py:135, reachable only from the static leg. runner.py:1875-1877 already admits this in a comment.

One correction to the framing: this is red teaming only. simulate() has a real TargetKind.DEPLOYMENT branch (simulation/api.py:1795-1798 -> from_orq_deployment, simulation/adapters.py:20-64) that reaches deployment() properly. So the accurate statement is: deployment targets work in simulation, and in red teaming only with mode="static". Hybrid is broken the same way, since it dies at the shared prefetch. No test covers a deployment target in dynamic or hybrid mode.

That makes the row at targets.md:32 and the sentence at :36 wrong as written. Fix the text and I will approve.

Old deep links: the single redirect is the right call. Three destinations cannot be preserved by one stub, and a compatibility page just adds something to rot. site/orq-deployment/index.html emits url=../guides/targets/, so the page URL survives, and anyone who lands mid-page can use search.

ROUTER_BASE_URL: confirmed live at redteam/backends/registry.py:55-56 and now absent from docs/ entirely. Not blocking. The cleaner fix is deleting the warning, since the value has been dead for a while and the docs row was the only thing keeping it discoverable. Follow-up ticket, not this PR.

Everything else checks out. I verified all four claimed fixes against source: ReportSummary.resistance_rate at contracts.py:1773 with no pass_rate in the redteam package; json.loads at dashboard/launch.py:88; honor_openai_base_url=False at openresponses/client.py:50 against True at redteam/backends/registry.py:62; no dotenv anywhere in pyproject.toml, and orq-ai-sdk>=4.12 is a plain runtime dep at :261 so dropping the extra install line is right. The moved material also holds up against source rather than just having been carried over: fetch_data.py:196-197 for blank rows failing loudly, evaluatorq.py:216,245 for jobs being optional under inference=False, types.py:201-215 for ExperimentInput. I cross-checked the config page against every os.environ read in src/ and found no invented variable and nothing missing from the samples.

mkdocs build --strict clean on my side, anchors present in the built HTML, CI 13/13.

The page reads much better than what it replaces. The 25-row wall was genuinely hard to start from.

Baukebrenninkmeijer:
All four inline comments are addressed and the threads are resolved. Two needed an edit — the "both string forms work identically" sentence at `targets.md:36`, and the `OPENAI_BASE_URL` row, which now says *conversation path* and names the three post-run calls that do honour it. The `gpt-4o-mini` inconsistency resolved itself when this branch rebased onto main, since #185 moved the whole site; `grep -rn gpt-4o-mini docs/` is empty.

Two things beyond what you raised. `eq redteam run --target` help now names the `--mode static` requirement, so the CLI stops advertising a form the default pipeline refuses. And the PR grew: it also carries the new `docs/installation.md` and `docs/components.md` pages, and a harmonization of the `docs-writing` / `docs-autofill` agent skills so the docs review loop lives in one place. Title and body updated to match — `ROUTER_BASE_URL` stays as you called it, a follow-up rather than this PR.

Ready for another look.

docs/guides/targets.md
Baukebrenninkmeijer: Fixed, and the code moved too rather than only the row. `red_team()` with a `deployment:<key>` target now raises a `ValueError` naming `mode="static"` and `simulate()`, before any client, credential or paid call — in `red_team()`, in `_run_dynamic_or_hybrid()`, and with a backstop arm in `_prepare_target`. The row says the same thing, and a table in `redteam/contracts.py` is now the single source for which target kinds each pipeline drives, so the two cannot drift. Tests in `tests/redteam/test_deployment_target_guard.py`. Real deployment support in the adaptive pipelines is RES-1494.

docs/guides/targets.md
Baukebrenninkmeijer: Fixed. The sentence now reads: both string forms are parsed the same way in Python and on the CLI, and the `mode="static"` restriction on deployments applies to both — on the CLI it is `--mode static`. The `eq redteam run --target` help string now carries the caveat too, so the advertisement and the restriction arrive together.

docs/configuration.md
Baukebrenninkmeijer: Taken — you are right, and it was the default that gave it away: `resolve_llm_client` has `honor_openai_base_url=True` by default, so the bare calls at `simulation/api.py:93`, `simulation/cli.py:255` and `simulation/reports/executive_summary.py:106` do pick it up. Only `openresponses/client.py:50` passes `False`. The row now says *conversation path* and names the post-run calls that honour it.

docs/guides/targets.md
Baukebrenninkmeijer: Resolved by rebasing onto main rather than by an edit here: #185 moved the whole site, and `grep -rn gpt-4o-mini docs/` is now empty. This page is consistent with `tuning.md`, `faq.md`, `guides/red-teaming.md`, `guides/agent-simulation.md`, `cli-reference/simulation.md` and `llm-as-a-jury.md`. No revert needed.

PR #188, opened 2026-09-03
feat: consolidate env-var readers on common/env_config (RES-1297)

KarinaKKarinaK:
Retire the private env-var readers and route the numeric/bool overrides through one shared, validated reader.

## Framing (corrected after review)
An earlier version of this description called the ticket wrong for assuming `common/env_config.py` existed. That was my error. A shared reader with `env_int`/`env_float` did exist briefly on the RES-1286 branch (`ff0e4049`) and was removed before merge (`bf3e88ee`) in favour of pydantic `Field` bounds on the recommendations config, which reject a meaningless value outright rather than warning and falling back. So this PR is a deliberate reintroduction of that reader for the process-global tuning knobs, where a warn-and-continue contract is wanted over a hard failure, not a correction to a mistaken ticket. Both spellings are the right call in their place: `Field` bounds for a config a caller constructs, a warn-and-fallback reader for an env knob you do not want to crash the process.

## What this does
- Adds `common/env_config.py` with `env_int` / `env_float` (optional `min_value`/`max_value`) and `env_bool`, on one contract: unset -> default silently; set-but-empty/whitespace, unparseable, out-of-range, or non-finite -> `WARNING` + default; never raises. Truthy `1/true/yes/on`, falsy `0/false/no/off`, case-insensitive.
- Routes `tracing/setup.py` (the four `ORQ_OTEL_*` ints via `env_int(min_value=1)`, `ORQ_DISABLE_TRACING` via `env_bool`) and `simulation/agents/base.py` (`EVALUATORQ_LLM_TIMEOUT_S`, `EVALUATORQ_LLM_MAX_TOKENS`, both `min_value=1`) through it, and deletes the private readers.

## Scope
Numeric/bool parsers only. String reads (`ORQ_API_KEY`, `ORQ_BASE_URL`) are left as-is. Left deliberately with a reason each: `EVALUATORQ_SPAN_MAX_TEXT_CHARS` (Optional int with a `<=0` = capture-all sentinel), `EVALUATORQ_CAPTURE_MESSAGE_CONTENT` (PII gate, bespoke opt-out), `EVALUATORQ_REASONING_EFFORT` (string enum), `ORQ_DEBUG` (truthy-any toggle), `COLUMNS` (terminal probe).

## Behavior change (documented in CHANGELOG)
A misconfigured `EVALUATORQ_LLM_TIMEOUT_S` / `EVALUATORQ_LLM_MAX_TOKENS` now warns and falls back instead of raising and crashing at import (`DEFAULT_MAX_TOKENS` is computed at module scope). Both are bounded `min_value=1`, so `0`, a negative, or `nan`/`inf` fall back with a warning rather than reaching the provider. `ORQ_DISABLE_TRACING` also accepts `yes`/`on` and is case-insensitive. The tracing knobs' empty-string warning and positive-only checks are preserved.

## Tests
`tests/common/test_env_config.py` covers the contract and asserts a `WARNING` is emitted on every invalid case (empty, whitespace, unparseable, out-of-range, non-finite, unrecognised bool), with unset kept silent. Removed-private-reader tests are rewritten against the new contract. `basedpyright` clean (whole repo), `ruff check src` / `ruff format --check src` clean, `pytest -m 'not integration'` green.

Closes RES-1297


src/evaluatorq/common/env_config.py
```diff
@@ -0,0 +1,75 @@
...
+    if max_value is not None and value > max_value:
+        logger.warning('{} must be <= {} (got {}); using default {}.', name, max_value, value, default)
+        return default
+    return value
+
+
+def env_int(name: str, default: int, *, min_value: int | None = None, max_value: int | None = None) -> int:
+    """Read an int override. Unset/empty -> default; invalid or out-of-range -> WARNING + default."""
+    raw = os.environ.get(name)
+    if raw is None or raw == '':
```
currentlycodinng: Empty string returns the default with no warning here, and the same short-circuit is at :51 and :67. CHANGELOG.md:20 says empty warns. Verified it does not:

    empty string ''  -> 7  warn=''
    whitespace '   ' -> 7  warn="X is not an integer ('   ')..."

On main the old tracing reader gave `ORQ_OTEL_MAX_QUEUE_SIZE is set but empty; using default 4096.` This reverts CHANGELOG.md:17, which documents that fix and the unresolved-CI-variable case it exists for. Split the empty branch out and warn on it before falling back.

src/evaluatorq/simulation/agents/base.py
```diff
@@ -90,7 +71,7 @@ def _default_max_tokens() -> int:
     as "no text and no tool calls". Raise via EVALUATORQ_LLM_MAX_TOKENS, or
     per-agent via ``LLMCallConfig.max_tokens``.
     """
-    return _env_int('EVALUATORQ_LLM_MAX_TOKENS', DEFAULT_TARGET_MAX_TOKENS)
+    return env_int('EVALUATORQ_LLM_MAX_TOKENS', DEFAULT_TARGET_MAX_TOKENS)
```
currentlycodinng: No min_value here or on the timeout at :63, so EVALUATORQ_LLM_MAX_TOKENS=0 is accepted and forwarded to the provider. Previously the raise at least stopped a nonsense value; now nothing does. You built min_value for exactly this and used it only on the tracing knobs. min_value=1 on both, please.

src/evaluatorq/simulation/agents/base.py
```diff
@@ -79,7 +60,7 @@ def _default_timeout_s() -> float:
     tailscale box under parallel load) can exceed the default; raise via
     EVALUATORQ_LLM_TIMEOUT_S, or per-agent via ``LLMCallConfig.timeout_ms``.
     """
-    return _env_float('EVALUATORQ_LLM_TIMEOUT_S', 60.0)
+    return env_float('EVALUATORQ_LLM_TIMEOUT_S', 60.0)
```
currentlycodinng: Same here, plus env_float parses 'nan' and 'inf' with no warning (verified, both returned as-is), and -5 is accepted as a timeout. min_value=1 (or some floor) would cover all three.

tests/common/test_env_config.py
```diff
@@ -0,0 +1,60 @@
...
+    monkeypatch.setenv('X_INT', '')
+    assert env_int('X_INT', 7) == 7
+
+
+def test_env_int_valid(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setenv('X_INT', '42')
+    assert env_int('X_INT', 7) == 42
+
+
+def test_env_int_invalid_warns_and_defaults(monkeypatch: pytest.MonkeyPatch) -> None:
```
currentlycodinng: Named ..._warns_and_defaults but nothing asserts a warning. Same across the file, no test in this PR checks that a warning is emitted at all. The deleted test_env_int_warns_on_set_but_invalid_value did (warning.assert_called_once() over ['', '   ', 'invalid', '0', '-1']), and that lost assertion is why the empty-string regression got through.

tests/common/test_env_config.py
```diff
@@ -0,0 +1,60 @@
...
+@pytest.fixture(autouse=True)
+def _clear(monkeypatch: pytest.MonkeyPatch) -> None:
+    for name in ('X_INT', 'X_FLOAT', 'X_BOOL'):
+        monkeypatch.delenv(name, raising=False)
+
+
+# --- env_int ---
+def test_env_int_unset_and_empty_use_default(monkeypatch: pytest.MonkeyPatch) -> None:
+    assert env_int('X_INT', 7) == 7
+    monkeypatch.setenv('X_INT', '')
```
currentlycodinng, line 42: This pins the silent-empty behaviour as intended, which is the opposite of what CHANGELOG.md:20 claims for this contract. Once empty warns, split this into an unset case and an empty case that asserts the warning.

tests/common/test_tracing_lifecycle.py
```diff
@@ -131,32 +131,9 @@ async def capture_processing(*args: object) -> list[object]:  # noqa: RUF029
...
-
-@pytest.mark.parametrize('raw', ['', '   ', 'invalid', '0', '-1'])
-def test_env_int_warns_on_set_but_invalid_value(monkeypatch: pytest.MonkeyPatch, raw: str) -> None:
-    warning = Mock()
-    monkeypatch.setenv('X', raw)
-    monkeypatch.setattr(tracing_setup.logger, 'warning', warning)
-
-    assert tracing_setup._env_int('X', 4096) == 4096
-    warning.assert_called_once()
+# The ORQ_OTEL_* int knobs read through common.env_config.env_int (min_value=1); the reader's
```
currentlycodinng, line 134: "the reader's contract (unset/empty/invalid/non-positive -> default + warning) is covered in tests/common/test_env_config.py" - the values are covered, the warning is not asserted anywhere in that file, and the empty case does not warn at all. Either the comment or the new tests need to change.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ All notable changes to `evaluatorq` are documented here.
...
+- **Env-var overrides now share one reader, and a misconfigured `EVALUATORQ_LLM_TIMEOUT_S` / `EVALUATORQ_LLM_MAX_TOKENS` warns and falls back to the default instead of raising.** `common.env_config` (`env_int` / `env_float` / `env_bool`) is now the single place env overrides are parsed and validated: unset falls back to the default, and an empty/whitespace, unparseable, or out-of-range value logs a `WARNING` and falls back to the default — it never raises. The private readers in `tracing/setup.py` and `simulation/agents/base.py` route through it. This changes the two simulation knobs above, which previously raised a `ValueError` on a non-numeric value and crashed the process where the fallback resolved; they now warn and use the default, matching the non-fatal contract the `ORQ_OTEL_*` tracing knobs already followed. `ORQ_DISABLE_TRACING` also now recognises `yes` / `on` and is case-insensitive, in addition to the previous `1` / `true`.
```
currentlycodinng: "an empty/whitespace, unparseable, or out-of-range value logs a WARNING" is not accurate for empty. It also contradicts the entry two lines above at :17 in the same unreleased section, which says a set-but-empty ORQ_OTEL_* variable warns instead of falling back silently. Whichever way the code lands, these two entries need to agree.

currentlycodinng (changes requested):
The refactor itself is right and the mechanics are clean: all four ORQ_OTEL_* ints route through env_int(min_value=1), which is equivalent to the old value <= 0 reject for ints, ORQ_DISABLE_TRACING goes through env_bool, and grep -rn '_env_int|_env_float' src/ confirms both private readers are actually gone rather than orphaned. ruff, basedpyright and 5239 tests are green for me too, and I could not reproduce the test_framework_targets_exposed.py flake in two runs (once with -p no:randomly), so that caveat may be stale.

Three things before this goes in.

The empty-string warning is not preserved, and CHANGELOG.md contradicts itself about it in the same unreleased section. Line 17 (a prior, unrelated entry) documents that a set-but-empty ORQ_OTEL_* variable now warns instead of falling back silently, specifically for the CI ${{ vars.X }}-expands-to-empty case. Line 20, this PR's own entry, claims the new shared reader preserves that: 'an empty/whitespace, unparseable, or out-of-range value logs a WARNING'. It does not, for empty. env_config.py:38 (and :51, :67) short-circuit on raw == '' before any logging:

    empty string ''  -> 7  warn=''
    whitespace '   ' -> 7  warn="X is not an integer ('   ')..."

On main the old tracing reader gave 'ORQ_OTEL_MAX_QUEUE_SIZE is set but empty; using default 4096.'; on this PR it's silent. That's a straight revert of the fix line 17 describes. Split the empty branch out and warn on it, or fix the CHANGELOG line to say what the code does and drop the now-false claim.

No test asserts a warning is emitted anywhere in this PR. test_env_int_invalid_warns_and_defaults only checks the return value, same pattern across the file. The test you deleted, test_env_int_warns_on_set_but_invalid_value, did assert warning.assert_called_once() over ['', '   ', 'invalid', '0', '-1'], and the comment replacing it at test_tracing_lifecycle.py:134 claims that contract is 'covered in tests/common/test_env_config.py'. It is not, and that lost assertion is exactly why the empty-string regression above got through untested.

Bound the two LLM knobs. Making them non-fatal is fine by me, and the old failure was worse than the description says: DEFAULT_MAX_TOKENS is computed at module scope (base.py:103), so a bad value raised at import time, not call time, in the old code. But you built min_value for exactly this kind of guard and used it only on the tracing knobs. Neither LLM call site passes one, so EVALUATORQ_LLM_MAX_TOKENS=0 and EVALUATORQ_LLM_TIMEOUT_S=-5 now sail through to the provider silently, and env_float parses 'nan' and 'inf' without complaint too. Removing the hard stop is only safe if the range check replaces it. min_value=1 on both.

Separately, on the framing. git log --all -S'def env_int' finds ff0e4049 on the #141 branch, which added common/env_config.py with env_int/env_float, and bf3e88ee, later in that same PR, which deleted it. The ticket was not wrong, env_config existed there and was consolidated away before merge. That commit's message says why: pydantic Field bounds reject values that cannot mean anything (a cap of 0 pays for the call, drops every result) instead of falling back with a warning, and the rationale still lives in common/recommendations.py:31. So this reintroduces a module a previous review removed and flips its contract back. A shared reader for these knobs still seems like the right call to me, but say that in the description as a reversal you're making on purpose, not as a correction to a mistaken ticket. Someone will read the two commits side by side eventually.

Scope calls all check out. EVALUATORQ_SPAN_MAX_TEXT_CHARS really does use value if value > 0 else None as a capture-all sentinel (common/tracing.py:63), which env_int(min_value=...) would have destroyed, correctly left alone, same for the other three. EVALUATORQ_REASONING_EFFORT (base.py:96) is a fourth bare read in a file you're editing, worth a line in the scope list.

KarinaKKarinaK:
Thanks, all five landed in ea22cda1.

1. Empty-string warning: you were right, the committed reader short-circuited on empty before logging (my warn-on-empty fix was sitting unstaged, so what I reported as verified was my working tree, not the PR). `env_int`/`env_float` now warn "set but empty" on empty/whitespace and fall back, restoring the tracing contract and the unresolved-CI-variable signal. CHANGELOG line now matches, and no longer contradicts the entry above it.

2. Bounds: `EVALUATORQ_LLM_TIMEOUT_S` and `EVALUATORQ_LLM_MAX_TOKENS` now pass `min_value=1`, and `env_float` rejects `nan`/`inf` (float() accepts them and the min/max checks can't catch a nan since every comparison is False). So 0, negatives and non-finite fall back with a warning instead of reaching the provider. Good catch on the import-time raise too, noted in the CHANGELOG.

3. Warning assertions: added. Every invalid case now asserts a WARNING fired (empty, whitespace, unparseable, out-of-range, non-finite, unrecognised bool), unset is asserted silent, and the empty and nan/inf cases are covered. That was the lost assertion that let the regression through.

4. Framing: corrected. You're right that env_config existed on the RES-1286 branch (ff0e4049) and was removed before merge (bf3e88ee) for the Field-bounds reason. The PR body and CHANGELOG now describe this as a deliberate reintroduction of a warn-and-fallback reader for the process-global knobs, not a ticket correction. I also fixed the same wrong framing on the Linear ticket.

5. EVALUATORQ_REASONING_EFFORT added to the left-alone scope list (base.py:96, a string enum).

basedpyright/ruff/tests green. Ready for another look.

src/evaluatorq/common/env_config.py
```diff
@@ -0,0 +1,94 @@
...
+
+def env_bool(name: str, *, default: bool) -> bool:
+    """Read a bool override. Unset/empty -> default; unrecognised -> WARNING + default.
+
+    Truthy: 1/true/yes/on. Falsy: 0/false/no/off (case-insensitive).
+    """
+    raw = os.environ.get(name)
+    if raw is None or raw == '':
+        return default
+    value = raw.strip().lower()
```
Baukebrenninkmeijer, line 88: **P2** — this tests for empty before it strips, so a whitespace-only value warns `is not a boolean`. `_raw_number` strips first and warns `is set but empty` for the same input. Strip here too.

tests/simulation/test_agents_base_env.py
```diff
@@ -1,38 +1,41 @@
...
+def test_max_tokens_parses_valid(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setenv("EVALUATORQ_LLM_MAX_TOKENS", "4096")
+    assert _default_max_tokens() == 4096
 
 
-def test_env_int_raises_named_error_on_garbage(monkeypatch: pytest.MonkeyPatch) -> None:
-    monkeypatch.setenv("EVALUATORQ_TEST_INT", "abc")
-    with pytest.raises(ValueError, match="EVALUATORQ_TEST_INT"):
-        _env_int("EVALUATORQ_TEST_INT", 8192)
+def test_max_tokens_garbage_warns_and_defaults(monkeypatch: pytest.MonkeyPatch) -> None:
```
Baukebrenninkmeijer: **P2** — the name promises a warning that the test does not assert. Add the warning assertion, or drop `warns` from the name. The same applies to `test_timeout_garbage_warns_and_defaults`.

Baukebrenninkmeijer (commented):

**3.5/5** — one shared reader, a real contract, and a test file that asserts the warning on every invalid case; the ticket's inventory step misses one env read.

I read the whole diff and checked the contract against each input class: unset, empty, whitespace, `1e3`, a float given to `env_int`, `nan`, `inf`, a negative, a value at `min_value`, and an unrecognised or mixed-case bool. Every invalid path logs a `WARNING` and returns the default, and none of them raises. All 14 CI checks pass: `Lint`, `Typecheck + test` on Python 3.10 to 3.13, and SDK compatibility on three versions.

### P1

1. **One numeric env read stays ad hoc and is not on the list of deliberate exceptions** (`src/evaluatorq/common/model_catalogue.py:122`) — `_CATALOGUE_TIMEOUT_S = float(os.environ.get('EVALUATORQ_CATALOGUE_TIMEOUT_S', '30'))` runs at module scope. A non-numeric value raises `ValueError` at import, which is the failure this PR removes from `simulation/agents/base.py`. The `Scope` section of the PR body names five bespoke reads and does not name this one. Route it through `env_float`, or add it to the list with a reason.

### P2

- **`env_bool` does not strip before it tests for empty** (`src/evaluatorq/common/env_config.py:86`) — `ORQ_DISABLE_TRACING='   '` warns `is not a boolean`. `env_int` and `env_float` call `_raw_number`, which strips first and warns `is set but empty`. Strip in `env_bool` too.
- **Two tests promise a warning that they do not assert** (`tests/simulation/test_agents_base_env.py:27`) — `test_max_tokens_garbage_warns_and_defaults` and `test_timeout_garbage_warns_and_defaults` assert the default only. Add the warning assertion, or drop `warns` from the name.
- **The module docstring lists two truthy words where the code now accepts four** (`src/evaluatorq/tracing/setup.py:56`) — `ORQ_DISABLE_TRACING` also accepts `yes` and `on`, and the match is case-insensitive.
- **No test covers the new `min_value=1` bound on the two simulation knobs** (`tests/simulation/test_agents_base_env.py:41`) — `EVALUATORQ_LLM_MAX_TOKENS=0` and `EVALUATORQ_LLM_TIMEOUT_S=-5` are a documented behaviour change with no case in this file.

Baukebrenninkmeijer:
## Against the ticket

RES-1297 asks for one thing: a single contract for how env overrides are parsed and validated. The scope names three steps — inventory all ad hoc `os.getenv` parsing in the package, replace it with `env_int` / `env_float` (plus a string or bool variant if needed), and keep the behaviour identical or document the change where the old readers were stricter or looser.

Steps 2 and 3 are done well for what the PR touches. `common/env_config.py` is a real contract rather than a helper: every invalid input class — unset, empty, whitespace, a float given to `env_int`, `nan`, `inf`, below `min_value` — logs a `WARNING` and returns the default, and none of them raises. The four `ORQ_OTEL_*` ints route through `env_int(min_value=1)`, which is equivalent to the old `value <= 0` reject; `ORQ_DISABLE_TRACING` goes through the new `env_bool`; the two simulation knobs stop raising `ValueError` on garbage, and the CHANGELOG says so. `tests/common/test_env_config.py` asserts the warning on each invalid case rather than only the fallback value, which is the part that usually gets skipped. Both private readers are gone rather than orphaned.

Step 1 — the inventory — is where it falls short, and that is the step the rest of the ticket depends on. Five typed env reads still parse by hand, and three of them are in `common/` next to the new module:

- `common/tracing.py:51` — `EVALUATORQ_SPAN_MAX_TEXT_CHARS` has its own private int reader with its own warning text. This is a third copy of exactly the thing the ticket asks to retire, and it survived because the PR body's scope list does not mention it. Its `None` default and its "non-positive disables truncation" rule are genuinely different from `env_int`, so it either needs an `env_int` variant that returns `None`, or a line in the PR saying it is a deliberate exception and why.
- `common/tracing.py:108` and `common/tracing.py:624` — `EVALUATORQ_CAPTURE_MESSAGE_CONTENT` and `EVALUATORQ_PROPAGATE_TRACE_CONTEXT` both do `flag.lower() == 'true' or flag == '1'`. That is now a *second* boolean contract in the same package: `EVALUATORQ_PROPAGATE_TRACE_CONTEXT=yes` silently turns propagation off, where `ORQ_DISABLE_TRACING=yes` turns disabling on. Neither warns on an unrecognised value. Two boolean dialects one import apart is the drift the ticket exists to prevent, and `env_bool` already covers both call sites.
- `common/model_catalogue.py:122` — `_CATALOGUE_TIMEOUT_S = float(os.environ.get('EVALUATORQ_CATALOGUE_TIMEOUT_S', '30'))` at module scope. A non-numeric value raises `ValueError` at *import* time, which is the failure mode this PR just removed from `simulation/agents/base.py`.
- `simulation/agents/base.py:96` — `EVALUATORQ_REASONING_EFFORT` is read and normalised by hand. The ticket anticipated this with "plus a string variant if needed", and no `env_str` was added. Fine to skip, but say so.

`common/cli_width.py:18` parses `COLUMNS` inside a `try`, and the plain string reads (`ORQ_API_KEY`, `ORQ_BASE_URL`, and friends) are out of scope — nothing is being parsed there.

**Verdict: the contract is right, the coverage is partial.** As it stands the PR replaces "several private readers" with "one shared reader plus four surviving private ones", which is an improvement but not the single contract the ticket asks for. Either finish the inventory in this PR — the three `common/` ones are a handful of lines each — or list the survivors with a reason in the PR body and open a follow-up, so the next person does not have to re-derive that they were considered.

## Still open on this PR

`currentlycodinng`'s `CHANGES_REQUESTED` predates `ea22cda1`, which addressed all five of its points; it is stale and only needs a re-review click to unblock the merge.

My own five findings against `ea22cda1` are still open, and are all small:

1. `common/model_catalogue.py:122` — route through `env_float` (also listed above).
2. `common/env_config.py:86` — `env_bool` tests `raw == ''` before stripping, so `'   '` warns "is not a boolean" while `env_int` and `env_float` warn "is set but empty" for the same input. Strip first.
3. `tests/simulation/test_agents_base_env.py` — `test_max_tokens_garbage_warns_and_defaults` and `test_timeout_garbage_warns_and_defaults` assert the default but not the warning. Add the assertion or drop `warns` from the name.
4. `src/evaluatorq/tracing/setup.py:9` — the module docstring says `ORQ_DISABLE_TRACING=1 or true`; `env_bool` accepts `1` / `true` / `yes` / `on`, case-insensitive.
5. No test covers the new `min_value=1` bound on `EVALUATORQ_LLM_MAX_TOKENS` or `EVALUATORQ_LLM_TIMEOUT_S`, which is a documented behaviour change.

## Tickets

RES-1297 is In Review with this PR attached. RES-1286, which spawned it, is Done. No other ticket covers the remaining readers, so if they are not finished here they need a new one — otherwise the third and fourth copies of the parsing contract stay in `common/` with nothing pointing at them.

src/evaluatorq/common/env_config.py
Baukebrenninkmeijer: Fixed in ea22cda1. `_raw_number` now strips first and warns `is set but empty` before falling back, so the empty and whitespace cases both signal and CHANGELOG.md:20 is accurate again. `tests/common/test_env_config.py` asserts the warning on both.

src/evaluatorq/simulation/agents/base.py
Baukebrenninkmeijer: Fixed in ea22cda1 — `_default_max_tokens` passes `min_value=1`. d9dfa372 adds the missing coverage: `EVALUATORQ_LLM_MAX_TOKENS=0` and `-5` both warn and fall back.

src/evaluatorq/simulation/agents/base.py
Baukebrenninkmeijer: Fixed in ea22cda1: `_default_timeout_s` passes `min_value=1.0`, and `env_float` rejects `nan`/`inf` with a `not a finite number` warning before the bounds check. d9dfa372 covers `0`, `-5` and `nan` at this call site.

tests/common/test_env_config.py
Baukebrenninkmeijer: Fixed in ea22cda1 — the file now captures loguru WARNINGs through a sink fixture and asserts one on every invalid case (empty, whitespace, unparseable, out-of-range, non-finite, unrecognised bool), plus asserts silence on the unset and valid paths. You were right that the lost assertion is what let the empty-string regression through.

tests/common/test_env_config.py
Baukebrenninkmeijer: Fixed in ea22cda1 — split into `test_env_int_unset_is_silent` and `test_env_int_empty_warns`, with `test_env_int_whitespace_only_warns` beside them. The unset case is the only silent one now.

tests/common/test_tracing_lifecycle.py
Baukebrenninkmeijer: Fixed in ea22cda1 — the warning is now asserted throughout `tests/common/test_env_config.py`, so the comment describes what is actually covered. The empty case warns too.

CHANGELOG.md
Baukebrenninkmeijer: Fixed in ea22cda1 — the code changed rather than the entry: empty and whitespace now warn, so :20 and :17 agree. d9dfa372 extends the same entry with the two bool toggles and the catalogue timeout that have since been folded in.

src/evaluatorq/common/env_config.py
Baukebrenninkmeijer: Adopted in d9dfa372 — `env_bool` strips first, then treats an empty result as unset. It stays a silent default rather than a warning: unlike a number, a bool has no "unresolved CI variable" case worth signalling, since the default is a real value either way. Covered by `test_env_bool_empty_and_whitespace_are_silent_defaults`.

tests/simulation/test_agents_base_env.py
Baukebrenninkmeijer: Adopted in d9dfa372 — both tests take a loguru sink fixture and assert `not an integer` / `not a number`. The same commit adds `min_value=1` cases for both knobs, which had no coverage.

Baukebrenninkmeijer:
Processed the feedback in d9dfa372. All nine inline threads are replied to and resolved.

**Adopted**

- `env_bool` strips before it tests for empty, so a whitespace-only value reads as unset rather than warning `is not a boolean`. It stays a *silent* default rather than matching the numeric readers' `is set but empty` warning: a bool has no unresolved-CI-variable case worth signalling, because the default is a real value either way. Tested.
- `EVALUATORQ_CAPTURE_MESSAGE_CONTENT` and `EVALUATORQ_PROPAGATE_TRACE_CONTEXT` now route through `env_bool`. Both parsed `flag.lower() == 'true' or flag == '1'` inline, which was a second boolean dialect one import from the shared one — an unrecognised value silently turned the toggle off, where `ORQ_DISABLE_TRACING` warns and keeps its default. They now accept `yes`/`on`/`no`/`off` too, and warn instead of reading garbage as false.
- `EVALUATORQ_CATALOGUE_TIMEOUT_S` routes through `env_float(min_value=0.1)`. It was parsed at module scope, so a non-numeric value raised `ValueError` at import — the exact failure this branch removed from `simulation/agents/base.py`.
- `EVALUATORQ_LLM_MAX_TOKENS` and `EVALUATORQ_LLM_TIMEOUT_S` gain tests for the new `min_value=1` bound, and the two garbage-value tests assert the warning their names promise.
- The `tracing/setup.py` module docstring names all four truthy words `ORQ_DISABLE_TRACING` accepts.
- The CHANGELOG entry is updated for all of the above, including the bool-toggle behaviour change.

**Deliberately not adopted**

`EVALUATORQ_SPAN_MAX_TEXT_CHARS` keeps its bespoke reader. Its default is `None` rather than a number and any non-positive value means "capture all", which `env_int` cannot express without a variant that returns `None` — more surface than the one call site earns. `EVALUATORQ_REASONING_EFFORT` is a string enum with no reader to route to, `ORQ_DEBUG` is a debug-only truthy-any check, and `COLUMNS` is a guarded terminal probe. All four are named in the CHANGELOG as exceptions with a reason, which was the ask.

That closes the inventory step: every typed env read in the package now goes through `common/env_config` or is listed as a deliberate exception.

`uv run ruff check src`, `ruff format --check src`, `basedpyright` (whole repo) and `pytest -m 'not integration'` (5268 passed, 2 skipped) are green locally.

@currentlycodinng — your `CHANGES_REQUESTED` predates ea22cda1 and all seven of its points are addressed; a re-review would unblock the merge.

currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 4.5/5** — all 6 original findings genuinely fixed, two knob-design calls flagged separately

### Verification
All 6 findings from the first round confirmed fixed, checked with a loguru warning sink attached, not by reading the diff. `env_int`/`env_float` on empty and whitespace now warn where they silently defaulted before (verified against the old commit side by side: old gave `warns=[]`, new gives "is set but empty; using default N"). Both LLM knobs are bounded now — 0/-1/nan/inf on the timeout all warn and fall back instead of reaching the provider silently. The warns-and-defaults tests actually assert a warning now instead of just the return value. CHANGELOG and the tracing docstring agree with the code.

**Suite:** ruff, ruff format, basedpyright, `pytest -m 'not integration'` all clean (5252 passed, 4 skipped).

Today's additions (`env_bool` for the two trace-context flags, `env_float` for the catalogue timeout, new `min_value=1` bound tests) don't reintroduce any of the three original bug classes — no unbounded knob left in the touched set, new bound tests assert their warnings.

### 🟡 Two judgment calls, not folded into the approval
1. **`EVALUATORQ_CAPTURE_MESSAGE_CONTENT` now fails open.** `common/tracing.py:109` defaults to `True`, so an unrecognized value used to give silent OFF and now gives ON with a warning — message text starts landing on spans. Every other knob here is right to warn-and-keep-default, but this one gates PII, and a WARNING in a log is a weak backstop against a typo'd deploy value turning capture on. Worth deciding if this one knob should fail closed instead of following the shared contract.
2. **The bool/number asymmetry is documented and deliberate** (`env_config.py:11`), so not an oversight, but still worth a second look: an unresolved CI var expands to empty for a bool exactly as much as for a number, so "a bool has no unresolved-variable signal" doesn't fully hold as the reason to skip the warning there.

### Two small gaps, non-blocking
No test covers the unrecognized-value path for the two newly-routed flags (`CAPTURE_MESSAGE_CONTENT`/`PROPAGATE_TRACE_CONTEXT`) — a regression there is invisible today. `redteam/README.md:215` and `simulation/README.md:159` still document only false/0 for the bool toggles; `tracing/setup.py`'s docstring got the update, those two didn't.

PR #189, opened 2026-09-03
feat: LLM-as-a-jury presets (RES-1171)

currentlycodinng:
Closes RES-1171.

Five named panels seatable with `llm_jury(preset="Balanced Trio")`, ported from the preset library derived in the research repo (RES-1346 / RES-1347, now merged).

| Preset | Seats | $/1k | Use when |
|---|---|---|---|
| Balanced Trio | deepseek-v4-pro, gpt-5.6-luna, gemini-3.6-flash | 4.64 | Default subjective eval |
| Strong Jury | claude-opus-5, gpt-5.6-sol, gemini-3.6-flash | 23.63 | Customer-facing benchmarks, preference data |
| Open-Weight / Portable | deepseek-v4-pro, kimi-k3, glm-5.2 | 10.29 | No dependence on a closed frontier vendor |
| EU Region | eu.claude-haiku-4-5, eu.gemini-3.5-flash, eu.gpt-5.6-luna | 6.56 | Data residency |
| Single-Provider Trio | gpt-5.6-sol, gpt-5.6-terra, gpt-5.6-luna | 14.28 | Workspaces locked to one provider contract |

$/1k is USD per 1,000 pointwise items at 1,500 input / 150 output tokens, uncached.

## What is here

- `jury_presets.py`: the five panels, and the two registers that owe a caller an explanation. `DROPPED_PRESETS` records a preset that shipped and was retired, `WITHHELD_PRESETS` one that is derived and costed but cannot be seated yet. `get_preset` hands either reason back by name instead of a bare "unknown preset".
- `common/data/jury_judge_rates.json`, 2.4KB: what each seated judge bills, the reasoning rung it was ranked at, and when the rates were captured. Committed rather than fetched, because a preset ships to customers on their own contracts and must not change because our gateway is down or a model is switched off in our workspace.
- `llm_jury(preset=...)`. A preset is a whole panel, so `judges=`/`model=` alongside it raises rather than overriding. It defaults aggregation and quorum, which the caller can still overrule.
- Published costs recomputed from those captured rates rather than typed in, so a repricing fails CI instead of quietly making the table wrong. `tests/integration/test_preset_pricing_drift.py` (opt-in) compares the captured rates against the live catalogue.

## What is deliberately not here

**The derivation.** It reads a capture of the whole model garden, and it is rerun, reviewed and argued in the research repo that owns the capture. An earlier revision of this PR carried `common/model_frontier.py`, a 46KB garden snapshot and a 593-line freshness suite into the customer wheel, where nothing ran them: the runtime touched two functions out of the module, and every other symbol had exactly one consumer, the test file. Worse, its staleness gate was a wall-clock assertion in the default unit suite whose remediation script lives in another repo, so it would have turned CI red for every contributor around 2026-10-24 with no fix available here. What ships is the derivation's output.

**Cheap Aggregate**, recorded in `WITHHELD_PRESETS` rather than shipped. `minimax/MiniMax-M2.7` answers a `json_schema` response format with prose, so the panel votes with four of its five seats, and four is an even panel, the one shape `_panel_is_well_formed` exists to reject. It passed the validator only because the validator counts declared seats and not voting ones. Seating its named reserve does not rescue it: `zai/glm-5.2` costs $2.76 per 1k for that seat alone and takes the panel to $4.69, past Balanced Trio, which is the whole reason the panel exists gone.

The seat itself is fine. The model produces schema-conformant JSON when the schema arrives in the prompt, and only fails when it arrives as a `json_schema` response format, which its model card says it does not support. `common/judge.py` already has the working path but degrades to it only inside `except BadRequestError`; MiniMax rejects nothing, it answers 200 with prose. Fixing that retry condition is its own change, since it affects every model rather than just presets. Cheap Aggregate ships the day it lands, with no seat changed.

## Two fixes that reach beyond presets

**`provider_family` matched markers in table order.** `('llama', 'meta')` is listed above `('deepseek', 'deepseek')`, so `deepseek/deepseek-r1-distill-llama-70b` came back `meta` while its qwen-distilled sibling came back `deepseek`, purely from where each marker sits in the tuple. Tokens are the outer loop now, so the first lineage token in the id decides and the table stops being order-sensitive. This function drives `_panel_composition_messages` and `strict_panel`, so the old behaviour was suppressing a single-family warning on a real single-family panel and aborting on a judge that did not share the target's lineage. Covered in `tests/redteam/test_panel_of_judges.py`.

**`model_catalogue` collapsed every host of a model onto one entry.** Hosts republish each other's models and do not always agree on price, so an exact-key lookup returned tensorix's $1.75/$3.50 for `deepseek/deepseek-v4-pro`, which bills $0.435/$0.87. That seat is the first seat of the default preset, so the SDK was reporting its own users a fourfold overstate on it. Entries are keyed on `provider/model_id` as well as bare, and a pinned id now prices at its own host. The drift test asks that module instead of routing around it with raw `httpx`.

## Verified live

Every seat was exercised against `my.orq.ai`, not just unit-tested: all preset router IDs bill exactly what the rate table captured, all seats and reserves were probed for structured output (`minimax/MiniMax-M2.7` is the one failure, above), and every shipped panel seats, runs, aggregates and reports agreement end to end.

## Notes

**Quorum.** `min_successful_judges` defaults to a majority of the seats under a preset. It is a floor on how many judges must answer, not the threshold the aggregator then applies: `majority` takes more than half of the votes that arrived, so a trio answering 1-1 clears a quorum of two and still returns nothing. An earlier revision of this PR described the two as the same number in four places. It defaulted to the full panel until a live run showed the cost, one unreachable seat voiding items the rest of the panel agreed on.

**Costing.** Strong Jury (23.625) and EU Region (6.555) land exactly on a half-cent, so binary floats and banker's rounding were deciding two published figures by representation rather than by a price. Costing is `Decimal` with `ROUND_HALF_UP`; those two figures move a cent, to 23.63 and 6.56.

## Open on the ticket, not in this PR

RES-1171 lists three rules this does not implement: `temperature: 0` for non-reasoning judges (which `llm_jury` already warns against, since most current seats reason by default), a `DEGRADED` verdict (`JuryResult` already carries `tie`, `inconclusive` and `judges_configured` vs `judges_succeeded`, which is the same information), and "disagreement routes to human review" (not built, and a real gap rather than a deviation). Raising all three on the ticket.


src/evaluatorq/llm_jury.py
```diff
@@ -359,6 +404,28 @@ def llm_jury(
...
+    ``preset`` seats a named panel from ``evaluatorq.jury_presets`` instead of
+    listing judges by hand — ``llm_jury(name='helpfulness', preset='Balanced Trio')``.
+    The seats are derived from the committed model-garden snapshot and carry a
+    published $/1k that a test recomputes, so a preset cannot quietly become a
+    different or a differently-priced panel.
+
+    It fills in three things and contradicts none: the judges (so ``judges``/
+    ``model`` alongside it is an error, not an override), ``aggregator``
+    (``"majority"``, or the preset's numeric rule) and ``min_successful_judges``
+    (the full panel — a preset's cost and agreement story are about all its
```
Baukebrenninkmeijer: **P0** — this says the quorum is the full panel. `_apply_preset` sets `quorum = len(panel) // 2 + 1` at line 356, which is a majority. Change the docstring to say a majority of the seats.

docs/llm-as-a-jury.md
```diff
@@ -89,12 +89,13 @@ helpfulness = llm_jury(
 
 | Argument | Default | What it does |
 | --- | --- | --- |
+| `preset` | `None` | Name of a ready-made panel from [Jury Presets](jury-presets.md), such as `"Balanced Trio"`. Seats the judges, the aggregation rule and a full-panel quorum. Mutually exclusive with `judges` and `model`. |
```
Baukebrenninkmeijer: **P0** — "a full-panel quorum" contradicts the `min_successful_judges` row six lines below, which says a majority of the seats. The code does the majority. Change this row.

docs/jury-presets.md
```diff
@@ -0,0 +1,82 @@
+# Jury Presets
+
+A jury is only as good as its panel, and assembling one by hand means picking judges, checking they do not share a training lineage, keeping an odd number of them, and re-checking all of it every time a provider ships something new. A preset is that decision made once and kept current: `preset="Balanced Trio"` seats three judges from three families, sets the aggregation rule, and requires the whole panel to vote.
```
Baukebrenninkmeijer: **P0** — "requires the whole panel to vote" contradicts line 47 of this same page and the code. Change it to a majority quorum.

docs/jury-presets.md
```diff
@@ -0,0 +1,82 @@
...
+
+## The presets
+
+Costs are USD per 1,000 pointwise items at 1,500 input and 150 output tokens, uncached, one call per judge per item. Every figure is recomputed from the committed model garden snapshot in the test suite, so the table below cannot drift away from what the code seats.
+
+| Preset | Judges | Aggregation | $ / 1k | Reserve |
+| --- | --- | --- | --- | --- |
+| **Balanced Trio** (default) | `deepseek/deepseek-v4-pro`<br>`openai/gpt-5.6-luna`<br>`google/gemini-3.6-flash` | majority | 4.64 | `deepseek/deepseek-v4-flash` |
+| **Strong Jury** | `anthropic/claude-opus-5`<br>`openai/gpt-5.6-sol`<br>`google/gemini-3.6-flash` | majority | 23.62 | `deepseek/deepseek-v4-pro` |
+| **Cheap Aggregate** | `openai/gpt-5.6-luna`<br>`google/gemini-3.5-flash-lite`<br>`xai/grok-4-1-fast`<br>`deepseek/deepseek-v4-flash`<br>`minimax/MiniMax-M2.7` | 3 of 5 majority | 2.56 | `zai/glm-5.2` |
```
Baukebrenninkmeijer: **P0** — this row shows five seats concluding on 3 of 5 for $2.56. `minimax/MiniMax-M2.7` returns no usable verdict today, so four seats vote. Add that to the limits section below.

src/evaluatorq/jury_presets.py
```diff
@@ -0,0 +1,595 @@
...
+        from evaluatorq.common.model_frontier import load_garden
+
+        garden = load_garden()
+        return tuple(j for j in self.judges if (m := garden.get(j)) and not m.priced_at_ceiling)
+
+    def duplicated_lineages(self) -> dict[str, tuple[str, ...]]:
+        """Lineages seated more than once, whose errors correlate.
+
+        Not an error. A panel may deliberately repeat a lineage on cost grounds,
+        as Cheap Aggregate does with two OpenAI judges, but the diversity claim
```
Baukebrenninkmeijer: **P2** — Cheap Aggregate seats one OpenAI judge, `openai/gpt-5.6-luna`, not two. Line 506 repeats the claim.

src/evaluatorq/jury_presets.py
```diff
@@ -0,0 +1,595 @@
...
+    'google': 'inside',
+    'groq': 'inside',
+    'moonshotai': 'inside',
+    'openai': 'inside',
+    'xai': 'alongside',
+    'zai': 'inside',
+}
+
+
+def billable_completion(usage: dict[str, int], model: str) -> int:
```
Baukebrenninkmeijer: **P2** — `billable_completion` has no caller in the package or in its tests. `REASONING_ACCOUNTING` and `JuryPreset.priced_below_seated_effort` have none either. Delete them, or move them to the probe script that uses them.

Baukebrenninkmeijer (changes requested):

**2/5** — the code is careful and well tested, but the shipped documentation states a quorum the code does not use, and it does not disclose a seat that cannot vote.

I read `_apply_preset`, the preset table, the freshness suite and the committed snapshot, and I checked RES-1171. CI is green on every check: `Lint`, `Typecheck + test` on 3.10 to 3.13, `SDK compatibility` on three SDK versions, `Examples (static)` and `Build (strict)`.

### P0

1. **The docs and the `llm_jury` docstring state a full-panel quorum, but the code uses a majority** (`src/evaluatorq/llm_jury.py:418`) — `_apply_preset` sets `quorum = len(panel) // 2 + 1` at line 356. A caller who reads `help(llm_jury)` believes `preset="Cheap Aggregate"` needs 5 decisive judges. It needs 3. Two more places say the same wrong thing: the `preset` row in the argument table of `docs/llm-as-a-jury.md` says "a full-panel quorum", and the `min_successful_judges` row six lines below it says "a majority of the seats". `docs/jury-presets.md` opens with "requires the whole panel to vote" and says the opposite at line 47.

2. **`docs/jury-presets.md` does not say that the `minimax/MiniMax-M2.7` seat returns no usable verdict** (`docs/jury-presets.md:23`) — the table sells Cheap Aggregate as five seats that conclude on 3 of 5 for $2.56. Today four seats vote. The section "Two more limits worth knowing before you quote a number" covers reasoning overspend and the missing human-agreement baseline, but not the dead seat. The PR body discloses it; the page a customer reads does not. Add a line there. The retry fix stays deferred, as you decided.

### P2

- The comment at `src/evaluatorq/jury_presets.py:174` says Cheap Aggregate seats "two OpenAI judges". It seats one, `openai/gpt-5.6-luna`. The same claim is at line 506.
- `billable_completion`, `REASONING_ACCOUNTING` and `JuryPreset.priced_below_seated_effort` have no caller in the package or in its tests. They ship to customers as dead code.

### Open Questions

- RES-1171 says to port the five presets from RES-996 verbatim. This PR ships six from the newer RES-1346 / RES-1347 derivation, with different seats under the same names, and retires Value Trio. Should RES-1171 point at the newer derivation instead?
- RES-1171 lists full-panel quorum, `TIE` and `DEGRADED` verdicts, and an automatic reserve swap-in on a generator-family conflict as rules that are baked in and not knobs. This PR defaults quorum to a majority and keeps it caller-overridable, and a family conflict warns instead of swapping in the reserve. Are those rules superseded?
- Should the derivation tooling — `common/model_frontier.py`, the disposition registers and the 593-line freshness suite — ship inside the customer package, or stay in the research repo that regenerates the snapshot?

currentlycodinng:
Both P0s and both P2s are fixed, and the branch is rebased onto main (`52f43ff`), so it merges cleanly again.

**P0-1, the quorum.** Fixed in all three places: the `llm_jury` docstring, the `preset` row of `docs/llm-as-a-jury.md`, and the intro of `docs/jury-presets.md`. All now say a majority of the seats, two of three or three of five, which is the count the aggregation rule needs to be decisive. Good catch: the docstring was written when the default was the full panel, and live testing changed the code without changing the prose.

**P0-2, the dead seat.** `docs/jury-presets.md` now opens its limits section with it: `minimax/MiniMax-M2.7` returns prose instead of a structured verdict through the router, so Cheap Aggregate concludes on the four judges that answer and bills for four rather than the five its figure assumes. The line also says the seat itself is fine, since the model complies when the schema reaches it as instructions and the fallback that would do so only fires when a provider rejects the request outright. That retry fix stays its own ticket.

**P2, the comments.** Both fixed. The `duplicated_lineages` docstring now points at Single-Provider Trio, which is the panel that actually repeats a lineage, three times.

**P2, the dead code.** `billable_completion` and `REASONING_ACCOUNTING` are gone. They recost probe artifacts, the probe and the artifact test both live in the research repo, and nothing here calls them. `JuryPreset.priced_below_seated_effort` I kept and wired up instead of deleting: it names a real understatement in the published table, five of the six panels return a non-empty list, and it now has a docs bullet and two tests. Deleting it would have removed the disclosure rather than the dead weight.

### Open questions

**RES-996 or RES-1346 as the source of the presets.** RES-1346 and RES-1347, and I will update RES-1171 to say so. The ticket's table is the RES-996 shape and predates the repricing: it still lists Value Trio, which RES-1346 retired on measurement (83% over its own table on the 2026-08-18 probe, and not the cheapest panel anyway), and it lists seats that are no longer the best buy in their lineage. Porting it verbatim would ship a table two derivations stale on day one. The names are the same, the seats are the current derivation, and every difference is on record in the registers.

**Superseded rules.** Yes for the quorum, no for the rest.

- Quorum: superseded, and by live testing rather than preference. Defaulting to the full panel made Cheap Aggregate return `inconclusive` on every item the moment one seat did not answer, which is the state that seat is in today. A majority is also what the `majority` aggregator needs to be decisive, so the two now agree. It stays caller-overridable because `min_successful_judges` is an existing public argument and a preset silently ignoring an argument the caller passed is worse than a preset that lets them raise it.
- `TIE` and `DEGRADED`: not dropped, already there. `JuryResult` carries `tie` when decisive votes split evenly and a caller tie policy broke them, and `inconclusive` when too few judges answered, with `judges_configured` against `judges_succeeded` showing a short panel. A preset that added verdict states of its own would fork the jury runner's reporting. I have documented the mapping in `docs/jury-presets.md` so it is not folklore.
- Reserve swap-in: deliberately a warning. A reserve is the model that takes a seat when its occupant is retired, which is a maintenance decision with a written reason behind it. Swapping one in automatically on a same-day failure changes the panel's lineage mix, and therefore the cost and agreement figures the preset publishes, without anyone seeing it happen. The docs now say this in as many words.

**Where the derivation tooling belongs.** `common/model_frontier.py` here is already the jury-relevant subset: `derive_trio`, `pareto_frontier`, `dominated_by` and the tier-ratio machinery stayed in the research repo, and what shipped is the snapshot reader plus the two accessors the presets and their tests need. The freshness suite has to ship with it, since its whole job is to fail when a committed seat stops being defensible, and a test that runs only in another repo cannot do that. Happy to move both if you would rather the customer package carried the presets and nothing else, but then the staleness checks stop guarding what customers actually receive.

Everything is green locally: 5,440 tests, `ruff check src`, `ruff format --check src`, and full-repo `basedpyright` clean.

currentlycodinng:
Reworked after a destructive review of the branch. The diff is now +388 / -2,986 against the previous revision, and most of what went is machinery that never ran here.

**Two real defects, both reaching past the presets.**

`provider_family` matched markers in `_FAMILY_MARKERS` order, and `('llama', 'meta')` sits above `('deepseek', 'deepseek')`. So `deepseek/deepseek-r1-distill-llama-70b` classified as meta while `deepseek-r1-distill-qwen-32b` classified as deepseek, from table position alone. That drives `_panel_composition_messages` and `strict_panel`, so it was suppressing the single-family warning on a genuinely single-family panel and aborting on a judge that did not share the target's lineage. Tokens are the outer loop now, so the first lineage token in the id decides. Tested in `tests/redteam/test_panel_of_judges.py` where the function lives, including both distill siblings.

`model_catalogue` keyed prices on the bare model id and collapsed every host, so `deepseek/deepseek-v4-pro` resolved to tensorix's $1.75/$3.50 against the $0.435/$0.87 that seat actually bills. That is the first seat of the default preset, so `price_usage` was reporting a fourfold overstate to our own users on the live judge path. I had found this, written it up in a test docstring and then routed around it with a private price table, which was the wrong call. Entries are keyed on `provider/model_id` as well as bare, the bare key still resolves to the developer's own entry as before, and the drift test asks that module now instead of hand-rolling `httpx`.

**The derivation left the wheel.** `common/model_frontier.py`, the 46KB garden snapshot and the 593-line freshness suite are gone. The runtime touched two functions out of that module; everything else had exactly one consumer, the test file. What ships is a 2.4KB table of what each seated judge bills, the rung it was ranked at and the capture date. Published costs are still recomputed from it on every run and the opt-in drift test still checks it against the live catalogue. That also removes three wall-clock assertions, one of which would have turned the whole unit suite red around 2026-10-24 for every contributor, with the remediation script sitting in a different repo.

**Cheap Aggregate is held back rather than shipped**, in a new `WITHHELD_PRESETS` register that `get_preset` reads. Four voting seats is an even panel, which is the one shape the validator exists to reject, and it only passed because the validator counts declared seats and not voting ones. Seating the named reserve is not a fix: `zai/glm-5.2` is $2.76 per 1k for that seat alone and takes the panel to $4.69, past Balanced Trio. It ships when the judge path falls back on an unparseable 200.

**The quorum claim.** You were right that the docs and the code disagreed, and the fix I pushed for that was still wrong: I described the majority quorum as "the same count the aggregation rule needs to be decisive", in four places. It is not. `_strict_majority` takes more than half of the votes that arrived, not of the seats, so a trio answering 1-1 clears a quorum of two and still returns nothing. All four now say it is a floor on how many judges must answer, and the difference is spelled out.

**Smaller ones.** `Aggregation` never reached the aggregator, so it is gone and the panel-size check reads `len(judges)`. `DEFAULT_PRESET` had no reader. `PRESETS` is a `MappingProxyType` with a trimmed lookup, matching `strategy_registry` and `backends/registry`; the rate table is frozen pydantic behind an `lru_cache`, so a caller can no longer reprice every preset process-wide by mutating a returned dict. Costing moved to `Decimal` with `ROUND_HALF_UP`, since Strong Jury (23.625) and EU Region (6.555) land exactly on a half-cent and were being decided by float representation; those two published figures move a cent. The tautological tests are gone, and the four dangling references to things this repo does not contain (`router_pools`, "Anthropic Only pool", "Balanced Pareto Trio", a `probes/archive/...` path) are out of the shipped source.

RES-1346 is merged, so the derivation this ports from is now the reviewed one.

Three ticket rules are still unimplemented and I am raising them on RES-1171 rather than deciding them here: `temperature: 0` for non-reasoning judges (which `llm_jury` already warns against), a `DEGRADED` verdict (`JuryResult` already carries the same information), and "disagreement routes to human review", which is a genuine gap rather than a deviation.

5,388 tests pass, `ruff check src`, `ruff format --check src` and full-repo `basedpyright` are clean.

Baukebrenninkmeijer (changes requested):
Reviewed the branch as it stands (through `eeef84ee`, plus the uncommitted working tree). The core of this is good and I want to say what specifically: seating derived from a committed snapshot rather than a live fetch is the right call, the decision registers (`NEVER_SEAT`, `UNROUTABLE`, `AGED_SEATS`, `REVIEWED_SUCCESSORS`, `DROPPED_PRESETS`) are a genuinely better pattern than a comment nobody diffs, and `ceiling_index` vs `intelligence_index` is a real distinction that most model-selection code gets wrong silently.

Requesting changes on three groups: one strategic (there is now a server-side version of this logic), one correctness (the documented flagship usage sends values providers reject), and one arithmetic (a set of numbers in comments and docs that do not reconcile with the snapshot they claim to describe).

---

## 1. Reuse: `orq/auto` and `orq/frontier` now do this server-side

Arian shipped `orq/auto` / `orq/frontier` on 2026-09-05 — two days before the reseats on this branch — and it implements the same selection arithmetic this PR derives locally:

- Pareto frontier of intelligence vs price (this PR's `dominates()`)
- a quality floor on the intelligence index (this PR's `MIN_SEAT_INDEX`)
- constraint variants: `orq/auto-open` for open-weight only, `orq/auto-eu` for EU-hosted endpoints — which map onto the **Open-Weight / Portable** and **EU Region** presets almost exactly
- `orq/frontier-cheapest` currently resolves to `deepseek-v4-flash`, the same model this branch just reseated Balanced Trio onto

The part that matters most for this PR: it re-runs on every catalog sync, model toggle and key change. That is the live version of the staleness problem this PR solves with a committed snapshot plus a 60-day expiry.

RES-1528 already tracks this and is assigned to you, so I am not asking for it here. What I would ask is that the module docstring say the relationship out loud — that the seating arithmetic is a local reimplementation of a policy the platform now also computes, that the snapshot exists so a customer-facing preset does not change under them mid-run, and that RES-1528 is where the two get reconciled. Right now a reader of `model_frontier.py` has no way to know a server-side equivalent exists, which is how the two drift apart.

One thing genuinely does not transfer, and it is worth writing down: a jury deliberately seats **three lineages** to decorrelate errors, whereas `orq/auto` picks the best models per complexity band and would happily return three cards from one vendor. So the diversity constraint stays local even if the frontier maths moves. That is the argument for keeping this code, and it should be in the docstring rather than inferred.

---

## 2. `seated_efforts()` returns values no provider accepts

`docs/jury-presets.md`, the `seated_efforts()` docstring and the CHANGELOG all tell the reader to pass the mapping straight into `llm_jury(reasoning_effort=...)`. Three seats break on that:

```
Cheap Aggregate  xai/grok-4-1-fast                     -> 'reasoning'
EU Region        aws/eu.anthropic.claude-haiku-4-5...  -> 'reasoning'
Cheap Aggregate  minimax/MiniMax-M3                    -> 'none'
```

`'reasoning'` is the capture's binary-dialect label for "thinking on" — a fact about the card, not an argument. `EFFORT_ORDER`'s own comment says it is deliberately excluded from the graded scale. `'none'` has the same problem and reaches the wire too, because `llm_call.py:218` gates on `if reasoning_effort:` and the string `'none'` is truthy.

Confirmed independently against the live catalogue: `supported_parameters` on all three of those router IDs does not list `reasoning_effort` at all.

The failure is silent rather than loud, which is what makes it worth blocking on. `llm_call.py:119` is:

```python
return 'reasoning_effort' in params and 'reasoning' in err_body
```

An error reading *"invalid value 'reasoning'"* satisfies `'reasoning' in err_body`, so a bad **value** is diagnosed as an unsupported **parameter**. The setting is dropped, the call is retried, and the model is memoized in `_REASONING_EFFORT_REJECTORS` for the rest of the process — so every later call to that judge silently runs at the provider default while `seated_efforts()` keeps reporting the operating point the panel was costed at. One wasted round-trip, then a panel that is not running where its published price says it is.

Two ways out, and I would take the first:

- Add a translation seam. A seat's *ranked* rung is not always a *sendable* value: map `reasoning` and `none` to `None` at the `llm_jury` / `PairwiseComparator` boundary, so `seated_efforts()` stays a truthful report of what a seat was ranked at while the request degrades cleanly.
- Or retract the documented usage and let `seated_efforts()` be reporting-only.

Related, and the reason I would rather fix this structurally than delete a docs sentence: **`EFFORT_ORDER` has no completeness guarantee.** It is a hand-maintained 7-tuple with one use site and no test, and providers add levels between captures — `xhigh` arrived that way. `ceiling_effort` breaks ties with `EFFORT_ORDER.index(e) if e in EFFORT_ORDER else len(EFFORT_ORDER)`, so an unrecognised rung sorts *last*, loses every cheapest-rung tie-break, and gets its seat ranked at an operating point picked by a fallback. Nothing raises. The snapshot already contains one value outside the tuple (`reasoning`, on 26 cards), so this is not hypothetical.

Worth noting the authoritative source already exists in this repo: `common/model_catalogue.py:204` parses accepted effort values from `parameters[reasoningEffort].config.options[].value`, and its own docstring explains that this is the field that grows a new level without a schema change (the `metadata.supports_reasoning_effort_*` booleans disagree with it on 50 of 148 entries). `model_frontier` just never sees it.

---

## 3. There is no way to refresh the snapshot from this repository

`tests/unit/test_preset_freshness.py` fails when the capture passes `MAX_SNAPSHOT_AGE` (60 days). Against `captured_at = 2026-08-25`, that is **around 24 October**, on every PR, including ones that touch none of this.

All four references to the refresh tooling name `projects/hitl-evaluators/scripts/refresh_model_garden_snapshot.py`, which is a path in the research repo. So the failure message points at something a reader of this repository cannot run, and the only way to satisfy the test is access that does not exist here.

This needs a script in-tree and something that runs it on a schedule. A monthly job that re-captures and opens a PR fits the 60-day ceiling with room to spare, and keeps a model change on its own reviewable diff instead of bundled into whatever was merging that week. The capture is mechanical but the consequences are not — a repriced seat invalidates a published `estimated_cost_per_1k`, a retired seat means promoting a reserve — so the job should report those as decisions in the PR body rather than acting on them.

On the endpoint: it has to be `/v2/models`, not `/v2/model-catalog`. The public catalogue is unauthenticated and lists everything, but publishes no reasoning-effort ladder and no autorouter intelligence index, and carries `pricing: {}` on canonical entries (the rates live on offering rows such as `alibaba/deepseek-v4-flash` with `offering_of: deepseek/deepseek-v4-flash`). It is still useful for exactly one thing the project-scoped endpoint cannot do: telling a **retired** seat apart from one this project merely has **switched off**.

---

## 4. `orq-ai-sdk` requirement was not bumped with the lock

`uv.lock` moved to 4.14.7 and `pyproject.toml` still says `orq-ai-sdk>=4.12` in four places (lines 261, 268, 288, 363). `model_catalog` does not exist in 4.12.4 and does exist in 4.14.0 — I checked both wheels — so a fresh resolve or a cold CI cache can legally install a version where `client.model_catalog` raises `AttributeError`. Please bump to `>=4.14`.

Worth adding a consumer in the same PR, or the bump has no stated reason. The natural one is below.

---

## 5. The drift test cannot see the two rows that most need checking

`tests/integration/test_preset_pricing_drift.py` calls `/v2/models` with raw `httpx`. That endpoint is project-scoped — and project scoping is precisely why `minimax/MiniMax-M3` and `openai/gpt-5.4-mini` had to be hand-added to the snapshot. The test ends with:

```python
if unlisted:
    pytest.skip(f'not listed for this key: {unlisted}; every other seat matched')
```

So the only two rows that did not come from the capture, and are therefore the only two nobody has cross-checked, are exactly the rows this endpoint cannot see — and the test **skips** rather than fails on them. The drift check has a hole shaped like the risk it exists to cover. Pointing it at `/v2/model-catalog` closes it (mind the `.model` wrapper on `.get()`, and that pricing lives on offering rows), and gives the SDK bump its consumer.

---

## 6. Numbers that do not reconcile with the snapshot

Each of these I recomputed against the committed `model_garden.json` rather than taking the comment's word for it. None changes a seating decision; all of them are load-bearing for the next person deciding whether a seat is still right.

**Ceiling-vs-headline mixups.** The module is explicit that seating ranks on `ceiling_index` and that `intelligence_index` is provenance only, but four comments compare one against the other:

- `jury_presets.py` Balanced Trio gemini seat: "50.1 against 45.4" is 3.6-flash's ceiling against 3.5-flash's *headline*. At the ceilings, it is **50.1 vs 50.2** — the successor is a tenth of a point *worse*, `dominates()` is False in both directions, and `in_family_upgrades` proposes neither. The reseat is still right on price ($3.00 vs $3.375); it is a trade, not an upgrade, and the module asks that a trade be justified in words.
- Strong Jury's gemini seat repeats "stronger at default effort and cheaper" for the same move.
- The haiku comparison, "23.7 at a $2.00 blend beside luna at 38.1 for $0.45", is headline-vs-headline in both `jury_presets.py` and the `AGED_SEATS` entry. At ceilings it is 29.6 vs 51.2 — same conclusion, and the honest version is stronger.
- The luna seat's "8.3 index points stronger" is likewise the headline gap; at ceilings it is 51.2 vs 29.8.

**Unit mixups.**

- Cheap Aggregate's reserve comment: "glm-5.2 ... at $2.15 against the panel's $2.56 for all five." $2.15 is a per-1M-token blend and $2.56 is per-1000-items. In the panel's own unit, glm-5.2 is **$2.76 per 1k for a single seat** — more than all five current judges together. The conclusion ("it stands in once, it does not run every night") survives and is actually better supported by the real numbers, but as written the figures read as the opposite of what they say.
- Balanced Trio's deepseek comment reads "deepseek satisfies while taking the panel from $6.11 to $4.11", then immediately describes the later pro-to-flash reseat. $4.11 is the flash total; the haiku-to-pro move was $6.11 to **$4.64**. Two moves, one figure.

**Counts that reconcile with nothing.**

- "19 of 24 look dominated" appears in both `jury_presets.py:13` and `model_frontier.py`, and it is the load-bearing justification for the whole in-family seating rule. Computed against this snapshot: 9 of 20 seat instances, 8 of 14 unique seated models, 10 of 17 counting reserves. No denominator here is 24 — I think it is a fossil from the research repo's larger library, which makes it the one number a reviewer most needs to be able to check.
- `model_frontier.py`: "monotone on 90 of 92 cards". Actual is **31 of the 33 cards that publish more than one graded rung** (both exceptions already in `NON_MONOTONE_LADDERS`). The 90/92 framing also implies the property was checked across all cards, which it cannot be — most publish fewer than two graded rungs.

**Registers describing a repo this is not.** `model_frontier.py`'s docstring says this package ships no router pools, but three register entries still reference them:

- `AGED_SEATS['claude-haiku-4-5-20251001']` says the seat is held in "the Anthropic Only pool and the cheap seat of EU Region", and then reasons about "both places" and "those two presets". Haiku holds exactly **one** seat here, in EU Region.
- `REVIEWED_SUCCESSORS['grok-4.5']` says it "already holds the middle seat of the Balanced Pareto Trio router pool".
- `UNROUTABLE['wafer/MiniMax-M3']` says the model is "seated in two pools".

**One register entry is factually wrong about the garden.** `REVIEWED_SUCCESSORS['gemini-3.7-flash']` says the garden lists it EU-only. It is served on all three Google hosts including the plain `google/` one — it is simply absent from `scored_models` and `preset_judge_pricing`, so there is no index to rank it on and no price to cost it at. Right conclusion, wrong reason, and it matters because 3.7-flash is a generation ahead of both seated Gemini cards and is the most likely thing to move these seats on the next capture.

---

## 7. Docs

- **`docs/jury-presets.md` line 70 goes stale the moment the reseat is committed.** The bullet says `minimax/MiniMax-M2.7` does not vote and to "read Cheap Aggregate as a four-seat panel". It is consistent at `eeef84ee`, but the uncommitted reseat replaces that seat with M3 and the uncommitted CHANGELOG says M3 was probed and answers — so as soon as those land, the page tells users to discount a panel over a judge that is not on it. `TestDocsTable` parses table columns and never reads prose, so nothing will catch it. Please delete the bullet in the same commit as the reseat (and renumber "Four more limits" to three).
- **Three ratios broke on the $4.64 → $4.11 reseat.** "five times as much" is 5.75×, "half the price" is 62%, "twice the default trio" is 2.50×. All three were roughly right before this branch.

---

## 8. Tests that pass without asserting the thing

- **`TestDocsTable._rows()` asserts two of five columns.** The regex captures Aggregation and discards it, and never reaches Reserve. This branch reseated three reserves; the table happens to be correct today (I checked by hand) but nothing would have caught it, and it is how the M2.7 prose survived.
- **`_unreviewed_successors` is blind to 37% of the catalogue.** `_first_listed` was fixed to treat `0` as "no date" rather than 1970; its twin forty lines up still says `if stamp is None`. **74 of 198** rows carry `0`, and a `0`-stamped candidate can never exceed `newest_seated`, so every undated model is structurally exempt from the successor check. The fixed helper is at least incidentally covered by a snapshot row; this one is not covered at all.
- **The per-judge effort test cannot distinguish keyed from positional assignment.** `{"m1": "high", "m2": None}` against `["m1", "m2", "m3"]` passes identically under an implementation that zipped efforts onto judges by position. Two distinct non-`None` efforts in a non-panel order fixes it.
- **The pairwise half of `reasoning_effort` has no tests at all**, including the unknown-key `ValueError`. And the flagship documented usage — `preset=` together with `seated_efforts()` — is never executed anywhere, which is how three seats came to be ranked at a value no provider accepts.
- **Snapshot completeness is guarded in two of four sections.** `served_model_ids` and `preset_judge_pricing` hard-assert; `scored_models` and `model_created_at` are guarded by five separate `continue`s. A hand-added row missing from the latter two takes a seat while opting out of seatability, ceiling-effort, dominance, successor and age checking — and the suite stays green.

---

## Offer

I have working fixes for most of §2 through §8 on a local branch — the requirement bump, an in-tree refresh script plus a monthly workflow, a `NON_GRADED_EFFORTS` / `known_effort` / `sendable_effort` seam with tests that fail on an unclassified rung, the corrected figures, the register fixes, the docs edits, and the strengthened tests. Full suite green (5476 passed), ruff and basedpyright clean, `mkdocs build --strict` builds.

Happy to push it to this branch, open it as a follow-up on top, or just hand you the diff — whichever you prefer. Say the word and I will not sit on it.

Baukebrenninkmeijer (approved):
Findings from the earlier review are addressed: `seated_efforts()` is reporting-only, the reasoning-effort retry is guarded at the root in `_is_reasoning_effort_rejection`, the seat comparisons are corrected and the docs disclose the dead seat and the price floor.

The derivation stays in the research repo, which is right. What was missing is that nothing here noticed a repricing until someone happened to push, so `scripts/refresh_jury_judge_rates.py` plus a monthly workflow now re-read the rates from the live catalog and open a PR when anything moves. Rates only — `seated_effort` and `priced_at_ceiling` are carried over untouched, so a reseating stays a decision made against the whole garden rather than something a cron job applies. Verified live against the catalog: 15 of 15 judges resolve and no rate has drifted since the capture.

RES-1528 replaces this with the server-side `orq/auto` routers; the script is deliberately the interim.

PR #192, opened 2026-09-06
fix(simulation): deliver evaluator failures through lifecycle hooks

Baukebrenninkmeijer:
## Problem

The first implementation of simulation evaluator results made important failures hard to observe and lifecycle behavior inconsistent. With fail-fast enabled, an evaluator could fail without its completion hook receiving the failure before `SimulationDroppedError` escaped. A result could also contain a numeric score without the evaluator's explanation, malformed criteria metadata could be reported as verified, and JSONL export could discard structured evaluator detail.

The practical failure looked like this:

```python
def on_evaluator_complete(event):
    record(event.result.error, event.result.raw_output)

try:
    await simulate(dataset, evaluators=[evaluator], exit_on_failure=True, on_evaluator_complete=on_evaluator_complete)
except SimulationDroppedError:
    pass
```

Before this follow-up, the callback could be skipped or receive incomplete detail, so callers could not reliably explain why a datapoint failed or reconstruct the evaluator output from a saved JSONL run.

## Solution

This follow-up makes evaluator failures part of the normal simulation result and lifecycle contract.

- Evaluator-completion callbacks run before fail-fast `SimulationDroppedError` propagation, so observers see the evaluator outcome.
- The primary simulation or drop error is captured explicitly, remains primary when lifecycle cleanup also raises, and cannot be replaced by an unrelated exception already being handled by the caller.
- Evaluator result contracts are enforced: duplicate evaluator names are rejected, malformed or otherwise unverified criteria evidence clears top-level `criteria_verified`, and non-numeric scores produce a derived error instead of silently looking valid.
- `SimulationResult.evaluator_details` preserves each evaluator's structured `raw_output`, including criteria evidence and conversation-quality component breakdowns.
- Existing numeric scores remain in `metadata["evaluator_scores"]` for report and dashboard readers.
- JSONL export retains the evaluator detail so a persisted run contains the same explanation as the in-memory result.
- JSONL export stringifies non-JSON-native custom evaluator detail instead of dropping the entire export.

## Verification

```text
uv run ruff check src
uv run basedpyright
```

All checks pass: 5,326 tests passed, 2 skipped, and 9 deselected; formatting, type checking, and strict documentation build also pass.

## Scope

Reports and dashboard projections for the new structured evaluator detail are intentionally deferred; existing consumers continue reading numeric scores from `metadata["evaluator_scores"]`.


Baukebrenninkmeijer:
## Review result

No remaining critical or important findings. The follow-up fixes the real issues found in the post-merge review of PR #191.

Fixed:

- Evaluator-completion hooks now run before fail-fast `SimulationDroppedError` propagation.
- `on_stage_end()` still runs when `on_run_complete()` raises, while the original exception is preserved.
- Malformed criteria metadata is marked unverified instead of appearing verified.
- Non-numeric evaluator values receive a derived error and are reported as failures.
- Returned `SimulationResult` objects now retain structured evaluator evidence and breakdowns in `evaluator_details`, while existing numeric score metadata remains unchanged.

Verification passed:

- `uv run ruff check src`
- `uv run basedpyright`
- `uv run pytest -m 'not integration'` — 5,318 passed, 2 skipped, 9 deselected.

The previously open API decision is resolved: evaluator detail is now part of the returned `SimulationResult` contract.

currentlycodinng (commented):
Ran the full verification suite myself against 05b3539: ruff check, ruff format --check, basedpyright all clean, mkdocs build --strict clean, and pytest -m 'not integration' gives 5318 passed, 2 skipped, 9 deselected, exact match to the PR body. The three headline claims (hook-before-raise, exception preservation, evaluator_details) are all real and reachable, backward compat on metadata['evaluator_scores'] holds, and the docs match what the code does.

One thing worth fixing before this merges. api.py:1636 fires on_evaluator_complete before the SimulationDroppedError raise at :1637, which is the point of the PR, but if that hook itself raises, the raise on :1637 never executes and the drop survives only as __context__ on the hook's exception. Repro:

    RAISED TYPE: RuntimeError | hook blew up
    IS DROPPED: False
    CONTEXT: SimulationDroppedError

On main the hook never fired on this path, so the drop always surfaced cleanly. Now a caller doing except SimulationDroppedError to detect an aborted run misses it if a hook also raises, and on_stage_end reports the hook's error as the stage error instead of the drop. The drop is control flow, the hook is observation, so I'd guard just this call site (log and chain, still raise the drop) rather than change the general unguarded-hook contract.

Smaller, not blocking: :2377's `not invalid and reason is None` can't do anything `reason is None` doesn't already do, since `invalid` forces `reason` non-None two lines above. :2623 keys evaluator_details by name only, but the docstring two lines above it says two JobResults can share a name, so the second silently overwrites the first, verified with two same-named scores producing one surviving entry. Same line stores score.score.raw_output by reference rather than copying it, so mutating the returned SimulationResult's evaluator_details mutates the EvaluatorScore object hooks already saw. :2620 mutates score.error on evaluatorq's own model after evaluatorq() has returned, which is safe today only because _stamp_evaluator_scores runs after upload and check_pass_failures, worth a line in the docstring since it's an ordering dependency someone could break later without noticing.

On tests: test_hooks.py:723 asserts isinstance + str(...) rather than identity for 'preserves the original exception', `hooks.stage_end_error is exc_info.value` would actually pin that claim. Nothing exercises the combined case (body raises a drop and on_run_complete also raises), which is exactly where stage_error gets overwritten at :1649, and nothing covers the log-and-swallow branch at :1657-1660. test_stamp_scores.py compares evaluator_details against the same dict object it was populated from, so it can't catch either the collapse or the aliasing above.

Things I could not settle by reading code or running the suite, flagging rather than assuming fine: every hook test here uses sync stubs (the run logs a DeprecationWarning about sync hooks), so the await_maybe async branch and datapoint_parallelism > 1 aren't exercised for this new ordering, needs a real run with async hooks. The except BaseException at :1647 catches CancelledError and does another await afterward; it re-raises so it reads correct, but a genuinely cancelled run would need to confirm that doesn't hang. And two product calls rather than correctness ones: evaluator_details now carries full raw_output (criteria evidence, conversation text) into the local run store and dataset exports, a wider footprint than the upload path, which deliberately strips it, worth confirming that's intended; and nothing in src/ consumes evaluator_details yet, so keying by name is unconstrained by any real reader, which is also what causes the collapse above.

src/evaluatorq/simulation/api.py
```diff
@@ -1630,19 +1630,36 @@ async def _simulate_core(
             # exit_on_failure aborted the run, but the rows that succeeded are real
             # results — hand them to on_run_complete (via the finally) instead of [].
             results = dropped.partial_results
+            # The drop is raised after evaluator scores have been stamped, so the
+            # buffered callbacks still belong to this run and must be delivered
+            # before the drop propagates.
+            await _notify_evaluator_complete(evaluator_events, resolved_hooks)
```
currentlycodinng: If this hook raises, the raise on the next line never runs and the SimulationDroppedError survives only as __context__ on the hook's exception. Repro against simulate(exit_on_failure=True) with a raising on_evaluator_complete: RAISED TYPE RuntimeError, IS DROPPED False, CONTEXT SimulationDroppedError. A caller catching SimulationDroppedError to detect the abort misses it. Worth guarding this one call (log + chain) even though the general hook contract is unguarded.

src/evaluatorq/simulation/api.py
```diff
@@ -2352,9 +2369,12 @@ def _sim_evaluation_raw_output(name: str, result: SimulationResult, value: objec
                 return None
 
         reason = unverified_reason(result)
+        if invalid:
+            invalid_reason = f'criteria_meta_invalid={len(invalid)}'
+            reason = f'{reason}; {invalid_reason}' if reason is not None else invalid_reason
         if 'criteria_results' in raw and reason is None:
             reason = 'criteria_meta missing (lossy dict)'
-        raw['criteria_verified'] = reason is None
+        raw['criteria_verified'] = not invalid and reason is None
```
currentlycodinng: `not invalid and reason is None` can't be anything other than `reason is None`: when invalid is truthy, reason was just forced non-None two lines above. Simplify to `reason is None`.

src/evaluatorq/simulation/api.py
```diff
@@ -2588,13 +2612,17 @@ def _stamp_evaluator_scores(
...
-                    events_out.append((sim_result, score))
                 value = score.score.value
                 numeric = isinstance(value, (int, float)) and not isinstance(value, bool)
                 reason = (
                     score.error if score.error is not None else (None if numeric else f'non-numeric value {value!r}')
                 )
+                if reason is not None and score.error is None:
+                    score.error = reason
+                if score.score.raw_output is not None:
+                    sim_result.evaluator_details[score.evaluator_name] = score.score.raw_output
```
currentlycodinng: Two things here. The docstring at :2568-2570 says two JobResults can carry the same evaluator name, but this keys by name only, so the second silently overwrites the first, verified with two same-named scores collapsing to one entry. And this stores raw_output by reference rather than copying, so mutating the returned SimulationResult.evaluator_details mutates the EvaluatorScore object hooks already received.

src/evaluatorq/simulation/api.py
```diff
@@ -2588,13 +2612,17 @@ def _stamp_evaluator_scores(
...
         for job_result in dp_result.job_results:
             for score in job_result.evaluator_scores or []:
-                if events_out is not None:
-                    events_out.append((sim_result, score))
                 value = score.score.value
                 numeric = isinstance(value, (int, float)) and not isinstance(value, bool)
                 reason = (
                     score.error if score.error is not None else (None if numeric else f'non-numeric value {value!r}')
                 )
+                if reason is not None and score.error is None:
```
currentlycodinng, line 2650: Mutates score.error on evaluatorq's own model after evaluatorq() has returned. Confirmed safe today, _stamp_evaluator_scores runs after the upload and after check_pass_failures reads score.error, but that's an ordering dependency, not a structural guarantee. Worth a line in the docstring.

tests/simulation/test_hooks.py
```diff
@@ -696,6 +696,34 @@ def on_evaluator_complete(self, datapoint_id, name, score, result) -> None:
...
+        await simulate(
+            datapoints=[datapoint_factory('dp-1')],
+            target=_ok_target,
+            user_simulator=_StubUserSim(),  # pyright: ignore[reportArgumentType]
+            judge=_StubJudge(terminate=True),  # pyright: ignore[reportArgumentType]
+            max_turns=1,
+            evaluator_names=['goal_achieved'],
+            hooks=hooks,
+        )
+    assert isinstance(hooks.stage_end_error, RuntimeError)
```
currentlycodinng: This checks isinstance and str(...), not identity. The claim being tested is 'preserves the original exception', `assert hooks.stage_end_error is exc_info.value` would actually pin that. Also no test covers the combined case (a drop while on_run_complete also raises), which is where stage_error gets overwritten at api.py:1649.

Baukebrenninkmeijer:
## Historian review

### Findings

1. The terminal-hook exception policy is duplicated and inconsistent across entry points. The new simulate() path preserves the original on_run_complete() failure and logs a later on_stage_end() failure at src/evaluatorq/simulation/api.py:1643-1662, but the neighboring generation paths still call on_stage_end() directly from finally at src/evaluatorq/simulation/api.py:995-1002 and src/evaluatorq/simulation/api.py:1154-1159. A stage-end hook failure can therefore mask the original generation failure, while the same hook failure is suppressed during simulation. Extract one shared terminal-hook helper or explicitly document that these entry points intentionally differ.

2. The new evaluator-detail field stops at SimulationResult, while the repository's canonical per-result projection drops it. src/evaluatorq/simulation/types.py:752-782 defines SimulationEntry as the report view model, and src/evaluatorq/simulation/reports/sections.py:513-545 constructs it without evaluator_details. The PR therefore retains detail for direct SimulationResult callers and saved model dumps, but silently omits the same evidence from generated reports and dashboard data. Decide whether this is intentionally API-only; if not, extend the typed projection and its consumers.

### Mechanical-fix pass

I found no safe mechanical fix that does not choose one of those two product/contract boundaries. The repository is clean and all PR checks are green.

Rating: MINOR ISSUES

Baukebrenninkmeijer:
## Full execution review

Rating: NEEDS REWORK

### Blocking correctness finding

- src/evaluatorq/simulation/api.py:1636-1637: an on_evaluator_complete hook exception can replace the intended SimulationDroppedError. The fail-fast path catches the drop, calls the hook, and only then reaches the bare re-raise. If the hook raises, that re-raise is never reached, so callers receive the hook exception instead of the control-flow error that signals dropped simulations. on_stage_end also reports the hook exception as the stage error. The existing test at tests/simulation/test_hooks.py:728-779 covers only a successful callback. Add the combined case and choose which error wins; I recommend preserving SimulationDroppedError and logging or chaining the observational hook failure.

### Important contract finding

- src/evaluatorq/simulation/api.py:2371-2378: malformed criteria metadata makes the new raw_output criteria_verified false, but SimulationResult.criteria_verified remains true. A result can therefore say it is verified through the typed public field while its evaluator detail says it is not. The new test at tests/simulation/test_evaluator_explanation.py:218-235 checks only raw output. Decide whether one malformed record makes the run-level typed flag unverified; the PR body says it does, so the scorer or validation path should update the typed field and test the report-facing value.

### Minor data-integrity findings

- src/evaluatorq/simulation/api.py:2623: evaluator_details stores raw_output by reference. Mutating the returned SimulationResult.evaluator_details mutates the EvaluatorScore object retained by hooks. Store an independent copy.
- The same line silently overwrites detail when two scores share an evaluator name. The input accepts an arbitrary evaluator_names list, so duplicate names are possible. Decide whether duplicates should be rejected, preserved as a list, or remain last-write-wins. tests/simulation/test_stamp_scores.py:79-99 does not exercise either case.

### Test gaps

- tests/simulation/test_hooks.py:713-724 checks only exception type and text, not identity. It does not prove that the original on_run_complete exception object is the one re-raised.
- No test covers the on_stage_end-also-raises branch at src/evaluatorq/simulation/api.py:1657-1660, async lifecycle hooks, or cancellation while the new BaseException handler performs another await.

### Aggregated decisions

1. Preserve SimulationDroppedError when an evaluator-completion observer fails, or let the observer exception win?
2. Should malformed criteria metadata set SimulationResult.criteria_verified=False?
3. Should duplicate evaluator names be rejected, preserved, or remain last-write-wins?
4. Should evaluator detail also be added to report and dataset-export view models, or is the direct SimulationResult and saved-run contract sufficient?

Checks run locally: targeted simulation tests 86 passed; full non-integration suite 5,318 passed, 2 skipped, 9 deselected; ruff, formatting, basedpyright, and strict docs build passed. No files were edited in this review because the repository rule explicitly forbids edits.

Baukebrenninkmeijer:
## Full review update

I ran the complete review lenses: strategy, execution and error paths, reuse and convention consistency, goal alignment, plus the adversarial `/hate --apply` pass.

Mechanical fix applied in `498d609a`: `results_to_jsonl()` now preserves `SimulationResult.evaluator_details`, with a regression test. The change is pushed in this PR.

Verification remains green: 5,319 non-integration tests passed, 2 skipped, 9 deselected; ruff check passed; ruff format check passed; basedpyright passed; strict docs build passed.

### Decisions for the owner

1. **Which error wins when a lifecycle observer fails?** A dropped run can currently raise `SimulationDroppedError`, then an unguarded `on_evaluator_complete()` or `on_stage_end()` hook can replace it. The current behavior matches the documented unguarded-hook policy. Recommended choice: preserve the root simulation/drop error and log observer-hook failures, so callers know the run failed for the real reason.

2. **Can evaluator names repeat?** `evaluator_details` is keyed by evaluator name, so repeated names overwrite earlier detail; existing numeric score metadata has the same shape. Recommended choice: reject duplicate evaluator names at configuration time, unless repeated names are an intended use case that needs a list-valued representation.

3. **Should malformed criteria metadata invalidate `SimulationResult.criteria_verified`?** The new raw detail reports `criteria_verified=False` and the scorer returns `0.0`, but the underlying result field can remain `True` when the judge audit flag was true and one metadata entry was malformed. Recommended choice: set the top-level field false too, unless it is deliberately reserved for judge-audit status only.

4. **Should evaluator detail appear in reports and dashboard view models?** The returned `SimulationResult`, saved `SimulationRun`, and JSONL export preserve it, but `SimulationEntry` and report sections currently omit it. Recommended choice: keep the current API/export scope unless the UI must display evidence; adding it to reports would be a separate surface change.

5. **Should `on_evaluator_complete` stream during scoring?** It currently delivers buffered events after evaluatorq returns, including partial events for a dropped run. Recommended choice: keep batch-end delivery for this PR and document it, unless progress callbacks are required.

6. **What is the privacy/size contract for custom evaluator `raw_output`?** The implementation retains raw output from every evaluator, not only built-ins, and saved runs therefore persist it. Recommended choice: explicitly accept this as the local-result contract, or narrow/redact it before making the field public.

No additional mechanical fixes were identified.

Baukebrenninkmeijer:
## Review decisions applied

All six owner decisions are now implemented:

- Primary simulation and drop errors survive failures in `on_evaluator_complete()`, `on_run_complete()`, and `on_stage_end()`; observer failures are logged when a primary run error already exists.
- Duplicate evaluator names are rejected at simulation configuration time because scores and detail are keyed by evaluator name.
- Malformed criteria metadata sets the top-level `SimulationResult.criteria_verified` to `False`, matching the scorer's `0.0` result and raw detail.
- Report and dashboard projections remain unchanged for now.
- `on_evaluator_complete()` remains a batch-end callback, with successful partial-row events delivered before a fail-fast drop is re-raised.
- Custom evaluator `raw_output` persistence is accepted and documented for returned results, saved `SimulationRun` JSON, and JSONL exports; Orq experiment uploads still strip it.

Added regression coverage for primary-error preservation, duplicate-name rejection, malformed criteria invalidation, and JSONL detail retention.

Final verification: 5,322 passed, 2 skipped, 9 deselected; ruff check and format check passed; basedpyright passed; strict docs build passed.

src/evaluatorq/simulation/api.py
```diff
@@ -1630,19 +1630,45 @@ async def _simulate_core(
...
             # render pairing) but reports the in-flight exception via
             # sys.exc_info()[1] — None on the success path (stage → 'completed'),
             # the live exception on failure (stage → 'error' in the manifest).
-            await await_maybe(resolved_hooks.on_run_complete(results))
-            await await_maybe(
-                resolved_hooks.on_stage_end(
-                    SimStage.SIMULATE,
-                    {'num_results': len(results), 'error': sys.exc_info()[1]},
+            stage_error = sys.exc_info()[1]
+            primary_error = stage_error
```
currentlycodinng: sys.exc_info()[1] on the line above returns whatever exception is live in the calling frame, not necessarily one raised here. Reproduced: calling `await simulate(...)` from inside a caller's own `except ValueError:` block, on a fully successful run, now raises that caller's ValueError at :1671 instead of returning results. The identical repro against 05b35397 (previously reviewed) returns correctly. primary_error needs its own None-initialized variable, independent of stage_error, which should only feed the on_stage_end payload below.

src/evaluatorq/simulation/api.py
```diff
@@ -2352,9 +2378,13 @@ def _sim_evaluation_raw_output(name: str, result: SimulationResult, value: objec
                 return None
 
         reason = unverified_reason(result)
+        if invalid:
+            invalid_reason = f'criteria_meta_invalid={len(invalid)}'
+            reason = f'{reason}; {invalid_reason}' if reason is not None else invalid_reason
+            result.criteria_verified = False
```
currentlycodinng: Non-idempotent: this sets criteria_verified on the result, but unverified_reason(result) two lines up already reads that same flag, so a second call to this function on the same result produces a different reason string. It's also one-sided, the other unverified branches below (terminated_by=error, lossy dict) don't touch the top-level flag, so raw['criteria_verified'] and result.criteria_verified disagree outside this specific malformed-metadata case.

src/evaluatorq/simulation/api.py
```diff
@@ -2352,9 +2378,13 @@ def _sim_evaluation_raw_output(name: str, result: SimulationResult, value: objec
...
         if 'criteria_results' in raw and reason is None:
             reason = 'criteria_meta missing (lossy dict)'
-        raw['criteria_verified'] = reason is None
+        raw['criteria_verified'] = not invalid and reason is None
```
currentlycodinng: Still dead from the last pass: invalid forces reason non-None two lines above whenever it's truthy, so `not invalid` can never be the deciding term here. `reason is None` is equivalent and says what's meant.

src/evaluatorq/simulation/api.py
```diff
@@ -2588,13 +2622,17 @@ def _stamp_evaluator_scores(
...
-                    events_out.append((sim_result, score))
                 value = score.score.value
                 numeric = isinstance(value, (int, float)) and not isinstance(value, bool)
                 reason = (
                     score.error if score.error is not None else (None if numeric else f'non-numeric value {value!r}')
                 )
+                if reason is not None and score.error is None:
+                    score.error = reason
+                if score.score.raw_output is not None:
+                    sim_result.evaluator_details[score.evaluator_name] = score.score.raw_output
```
currentlycodinng: Still stores raw_output by reference and still silently drops the second of two same-named scores here specifically. The construction-time duplicate check in _config.py closes the case where evaluator_names itself repeats, but two distinct EvaluatorScore objects that happen to share a name at this point (e.g. from separate evaluator runs) still collide at this assignment, which is the scenario this function's own docstring above describes.

src/evaluatorq/simulation/utils/dataset_export.py
```diff
@@ -209,6 +209,7 @@ def results_to_jsonl(
                 'token_usage': result.token_usage.model_dump(mode='json'),
                 'turn_metrics': [tm.model_dump(mode='json') for tm in result.turn_metrics],
                 'metadata': result.metadata,
+                'evaluator_details': result.evaluator_details,
```
currentlycodinng: Every other field in this dict goes through model_dump(mode='json'); this one is passed straight into json.dumps two lines up with no default=. A custom evaluator's raw_output carrying anything non-JSON-native (datetime, a numpy scalar) now raises mid-comprehension and drops the whole export batch, not just the offending row.

tests/simulation/test_hooks.py
```diff
@@ -696,6 +696,59 @@ def on_evaluator_complete(self, datapoint_id, name, score, result) -> None:
...
```
currentlycodinng: Still isinstance + str() rather than identity, same as last pass. Also still no test for the combined case, a body-level drop while on_run_complete also raises, which is exactly what the new primary_error rework above changes the behavior of.

currentlycodinng (changes requested):
Re-reviewed at 5919b22d against my prior COMMENT at 05b35397. ruff/format/basedpyright clean, pytest -m 'not integration' gives 5306 passed / 4 skipped (close enough to the claimed 5322/2, env drift), mkdocs build --strict clean.

Of the 5 prior points: #1 (hook-raise masking the drop) is properly fixed, api.py:1638-1643 now guards that call site and logs instead of losing the drop, new test covers it. #3's duplicate-name half is genuinely closed too, _config.py:170 rejects duplicate evaluator_names at both SimulationConfig construction sites, verified with a real ValidationError. Good.

But this grew by two commits and introduced a real regression, plus the rest of the original list is still open.

api.py:1649-1650 sets `primary_error = stage_error` where `stage_error = sys.exc_info()[1]`. That reads whatever exception is currently being handled in the calling frame, not necessarily one this function raised. If `simulate()` is awaited from inside a caller's own `except:` block (the ordinary `try: ... except: results = await simulate(...)` fallback pattern), a fully successful run now raises the caller's unrelated exception at :1671. I reproduced this: a successful simulate() called from inside `except ValueError: ...` returns the caller's ValueError instead of the results. Running the identical repro against the previously-reviewed 05b35397 returns the results correctly, so this is new. primary_error needs to start at None, independent of stage_error, which should stay scoped to the on_stage_end payload only.

The jsonl export widening (dataset_export.py:212) makes the still-open aliasing bug more consequential. Every other field there goes through model_dump(mode='json'); evaluator_details is dumped raw via a bare json.dumps two lines up with no default=. A custom evaluator's raw_output carrying a datetime or similar now takes down the whole export batch, not just that row, since export_to_jsonl builds one string per call inside a comprehension with no per-row guard.

api.py:2384's new `result.criteria_verified = False` write is non-idempotent: it's set inside a branch that also feeds unverified_reason(result), which itself reads criteria_verified, so calling this twice on the same result produces a different reason string the second time. It's also inconsistent with the other unverified paths a few lines below (terminated_by=error, lossy dict) which leave the top-level flag untouched, so raw['criteria_verified'] and result.criteria_verified can and do disagree outside the malformed-metadata case specifically.

Still open from before: the dead condition at api.py:2387 (`not invalid and reason is None`, invalid always forces reason non-None first); the evaluator_details aliasing at api.py:2633 (still stores raw_output by reference, still silently drops the second of two same-named scores at the point of assignment even though construction-time duplicates are now rejected, meaning two different evaluator score objects sharing a name post-construction still collide); the undocumented score.error mutation at api.py:2630-2631; and test_hooks.py:723 still asserts isinstance+str() rather than exception identity, with no test yet for the combined case (a body-level drop while on_run_complete also raises), which is exactly the class of case the new primary_error rework touches.

One more, not a bug, just worth naming since it's untested and uncatalogued: a body drop plus a raising on_run_complete now surfaces SimulationDroppedError instead of the on_run_complete error, a reversal from 05b35397's behavior. Feels like the right precedence, but it's a real behavior change to already-reviewed code with no test and no CHANGELOG line.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Already addressed in 5919b22d. The fail-fast branch now wraps `_notify_evaluator_complete` in `try/except BaseException`, logs an observer failure, and re-raises the original `SimulationDroppedError`; `test_drop_error_survives_evaluator_hook_failure` covers it.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Applied in 3b7b86d2: after invalid metadata is folded into `reason`, `criteria_verified` now uses `reason is None`; the old compound check was redundant.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Applied in 3b7b86d2: `raw_output` is deep-copied before storage, so returned details no longer alias `EvaluatorScore`. Duplicate configured evaluator names are rejected by `SimulationConfig` (already in 5919b22d); per-score hook events remain one-per-occurrence for defensive malformed upstream results.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Documented in the `_stamp_evaluator_scores` docstring in 3b7b86d2: `score.error` is filled in place for non-numeric values so the same score delivered to hooks carries the derived reason, while returned raw detail is copied. The ordering is intentional.

tests/simulation/test_hooks.py
Baukebrenninkmeijer: Applied in 3b7b86d2: the existing stage-end test now asserts exception identity, and `test_drop_error_survives_on_run_complete_failure` covers a drop plus a failing `on_run_complete` while preserving the original drop in stage metadata. The existing evaluator-observer test covers `on_evaluator_complete`.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Fixed in 4f1f79cf. `primary_error` is initialized to `None` before the inner run, set only by an exception raised by the simulation or lifecycle callbacks, and `stage_error` is derived from it; successful runs no longer consult caller state via `sys.exc_info()`. `test_successful_simulation_does_not_inherit_caller_exception` covers the repro.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Fixed in 4f1f79cf. Invalid metadata now determines a stable reason before the top-level flag is updated, and every unverified reason sets `SimulationResult.criteria_verified` to `False`. Regression coverage checks repeated extraction and terminated runs.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Already fixed in 3b7b86d2: the expression is now `raw['criteria_verified'] = reason is None`, after the invalid reason has been established.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Fixed across 3b7b86d2 and 6aeec212. Stored `raw_output` is deep-copied, and `_stamp_evaluator_scores` now rejects duplicate names at runtime as well as at configuration time, so distinct malformed upstream scores cannot silently overwrite one another. Added a regression test.

src/evaluatorq/simulation/utils/dataset_export.py
Baukebrenninkmeijer: Fixed in 4f1f79cf. JSONL export now uses `default=str` for the complete row, so non-JSON-native custom evaluator detail is serialized instead of aborting the whole export. Added a datetime regression test.

tests/simulation/test_hooks.py
Baukebrenninkmeijer: Already addressed in 3b7b86d2. The stage-end test asserts exception identity, and `test_drop_error_survives_on_run_complete_failure` covers a drop plus a failing `on_run_complete` while preserving the drop in stage metadata.

src/evaluatorq/simulation/api.py
```diff
@@ -2351,9 +2384,11 @@ def _sim_evaluation_raw_output(name: str, result: SimulationResult, value: objec
             else:
                 return None
 
-        reason = unverified_reason(result)
+        reason = f'criteria_meta_invalid={len(invalid)}' if invalid else unverified_reason(result)
         if 'criteria_results' in raw and reason is None:
             reason = 'criteria_meta missing (lossy dict)'
+        if reason is not None:
+            result.criteria_verified = False
```
currentlycodinng, line 2391: This also fires on the lossy-dict branch (raw['criteria_results'] present, no criteria_meta, reason set two lines above), which previously left the top-level flag alone. Since unverified_reason(result) at :2387 reads this same flag, repeated calls on the same result now flip the reason string, verified: first call gives 'criteria_meta missing (lossy dict)', second gives 'criteria_verified=False'. Worse than a string flip: a criteria_met score of 1.0/PASS on the first pass reads 0.0/unverified on a second pass over the same object, since the flag this line sets is what the next pass's reason computation reads. Needs the same fix already applied to the invalid-metadata branch, snapshot reason before the write, or scope this write to only the invalid case.

src/evaluatorq/simulation/api.py
```diff
@@ -2576,25 +2611,48 @@ def _stamp_evaluator_scores(
...
+                seen_names.add(score.evaluator_name)
                 value = score.score.value
                 numeric = isinstance(value, (int, float)) and not isinstance(value, bool)
                 reason = (
                     score.error if score.error is not None else (None if numeric else f'non-numeric value {value!r}')
                 )
+                if reason is not None and score.error is None:
+                    score.error = reason
+                if score.score.raw_output is not None:
+                    sim_result.evaluator_details[score.evaluator_name] = deepcopy(score.score.raw_output)
```
currentlycodinng, line 2653: deepcopy can raise on raw_output shapes the previous by-reference store tolerated (locks, anything unpicklable), which would abort here before results is assembled two lines below. Narrow today (only a custom evaluator in the public SIMULATION_EVALUATORS export could hit this), but it's the same population the export's new default=str fix was written to handle safely, worth picking one failure mode for that input rather than two commits disagreeing on it.

src/evaluatorq/simulation/api.py
```diff
@@ -2576,25 +2611,48 @@ def _stamp_evaluator_scores(
...
         sim_result = result_cache.get(id(dp_result.data_point))
         if sim_result is None or not dp_result.job_results:
             continue
+        seen_names = seen_names_by_result.setdefault(id(sim_result), set())
         for job_result in dp_result.job_results:
             for score in job_result.evaluator_scores or []:
-                if events_out is not None:
-                    events_out.append((sim_result, score))
+                if score.evaluator_name in seen_names:
+                    raise ValueError(f'duplicate evaluator name for simulation result: {score.evaluator_name!r}')
```
currentlycodinng, line 2643: This raises before results is built at :2537, so a duplicate name discards an otherwise-completed run rather than logging and keeping the first score. Not reachable today since only one job is passed in, but it now contradicts _notify_evaluator_complete's docstring a few lines below, which still describes two same-named JobResults as an expected case that produces two events. Worth reconciling the doc and the guard either way.

currentlycodinng (changes requested):
Re-reviewed at 6aeec212. ruff/format/basedpyright clean, pytest -m 'not integration' gives 5310 passed / 4 skipped, matching this run's own claims within normal drift.

Of the 6 open items: the primary_error regression is properly fixed, walked every set/read of it through the finally block and it no longer picks up a caller's unrelated exception. The dead condition is genuinely simplified. The aliasing and duplicate-name-collision pair from evaluator_details is fixed on both halves, deep copy is real (verified by mutating a nested dict and confirming the source EvaluatorScore is untouched), and same-name collision now raises instead of silently overwriting. The jsonl export handles datetime/Decimal/set/custom-repr without dropping the batch. The identity assertion and the combined-case test are both real and both pass.

One of the six moved rather than closed, and it's the one that matters most since it changes recorded scores.

api.py:2390-2391 now sets `result.criteria_verified = False` whenever `reason is not None`, which also fires on the lossy-dict branch two lines above it (`raw['criteria_results']` present, no criteria_meta). That branch used to leave the top-level flag alone. Now the first call sets it False, and since unverified_reason(result) two lines up reads that same flag, a second call on the same result produces a different reason string, exactly the non-idempotence this thread already flagged once for the invalid-metadata branch. I reproduced it end to end: repeated calls on a lossy-dict result now read `'criteria_meta missing (lossy dict)'` then `'criteria_verified=False'`, where 5919b22d was stable across repeats. And it isn't just a string, a re-scored criteria_met result that returned 1.0 / PASS on the first pass returns 0.0 / "unverified, not met" on a second, since the score computation before this function runs and the recorded flag after it now disagree with each other across passes on the same object. Reports, the dashboard, and sim_views.py all read this flag downstream. This needs the same treatment the invalid branch got, snapshot reason before writing the flag, or don't let the lossy branch feed back into unverified_reason after the write.

Two smaller ones from testing the same seam. The new deepcopy at api.py:2653 can now raise on a raw_output shape the old by-reference store tolerated fine (a lock, anything unpicklable), which would kill an otherwise-successful run at the point of stamping, before results is even assembled two lines below at :2537, the same one-error-hides-the-work shape as the primary_error bug. Only reachable through a custom evaluator registered into the public export today, but the export widening a few lines away was written for exactly that population, so the two fixes are pulling in different directions on the same input. And the new duplicate-name ValueError at :2643 fires before results is built at :2537, so a completed, paid-for run gets discarded on an anomaly in score shape rather than logged and the first score kept, not unreachable today since only one job is ever passed in, but the guard now directly contradicts _notify_evaluator_complete's own docstring three lines below still describing two same-named JobResults as an expected, handled case.

Not flagging as a bug, but worth a decision: the None -> False widening for criteria_verified on lossy-dict and terminated runs is a semantic change to what that field means for existing saved runs, not purely a correctness fix, want to confirm that's intended before it ships.

PR #193, opened 2026-09-07
docs: document CallableTarget on the targets guide

Baukebrenninkmeijer:
## The gap

`gap: entry point (red_team()) × target kind (CallableTarget)`

`CallableTarget` — the wrapper that turns a plain Python function into an `AgentTarget` — was named in no page under `docs/`. A reader whose agent is already a function was routed by the "Choosing a kind" table to *"Your own `AgentTarget` subclass"* and its ~150-line worked example, when a one-argument wrapper already ships.

It ranked first because "my agent is a Python function, how do I test it?" is a first-day question, and because the entry points disagree in a way nothing documented: `simulate()` auto-wraps a bare callable (`simulation/runner/simulation.py:679`), while `red_team()` rejects one outright — `TypeError: Invalid target type: function. Expected str or AgentTarget.` (`redteam/runner.py:762`). Verified by running it, not by reading it.

## What the page covers

A new section on `docs/guides/targets.md`, plus a row in the "Choosing a kind" table and a pointer under "Writing your own target":

- The prerequisite — the red-team extra and a key — and a warning that a red-team run uploads an Experiment with no way to opt out.
- Wrapping a function and passing it to `red_team()`, with a runnable example.
- The `red_team()` / `simulate()` asymmetry, demonstrated by a block that actually raises the `TypeError`.
- The silent failure mode: a callable declares no tools, so strategies gated on `requires_tools=True` are **skipped, not failed** — a run can report a high resistance rate meaning *those attacks were never attempted*.
- Declaring tools through `agent_context=` so the planner writes against them.
- What each return type does, including the unwarned `str()` coercion of anything that is neither `str` nor `AgentResponse`.
- `reset_fn` and `usage_fn`, including that an `async def usage_fn` is never awaited and kills the call with a pydantic `ValidationError` rather than degrading.

**Deliberately not covered:** the full `simulate()` walkthrough. An earlier draft carried one, but it duplicated the persona and scenario fixture from `guides/agent-simulation.md` character for character; two reviewers flagged it independently and it is now a cross-link.

## Receipt

Every code block this branch adds to the page, run in document order against a fresh `EVALUATORQ_DIR`. `run_receipt.sh` gained an optional base-ref argument for this: the page's pre-existing blocks were **not** re-run, because re-billing live API calls for prose nobody changed proves nothing about this branch.

| block | exit | first line of output |
|---|---|---|
| `targets.md:160` — the `uv add` / `export` prerequisite | — | illustrative (environment setup for the reader's project; not executed against this repo — see below) |
| `targets.md:168` — `red_team()` + `CallableTarget`, static, 1 datapoint | 0 | `Downloading dataset from HuggingFace: orq/redteam-vulnerabilities…` (prints `Resistance: 100%`) |
| `targets.md:212` — bare function handed to `red_team()` | 0 | `Invalid target type: function. Expected str or AgentTarget.` |
| `targets.md:240` — `agent_context=` with a declared tool | 0 | `True` |

`RESULT: OK — 4 added blocks accounted for, every executable one exited 0`

**The one exemption, stated plainly so it can be checked.** The prerequisite block is recorded, not executed. `uv add "evaluatorq[redteam]"` cannot resolve inside this package's own checkout (`self-dependencies are not permitted`), and a placeholder `export ORQ_API_KEY=...` would overwrite the real key with the literal `...` for every later block in the shared shell. Those lines are instructions for the reader's own project, not claims about this repo. The runner now applies that rule by pattern rather than on my say-so.

`mkdocs build --strict` and `scripts/validate_mermaid.py` both pass. Nothing under `src/` changed.

## Persona verdicts

|---|---|---|---|
| First-timer | Red-team a plain async function, starting from the docs root | COMPLETED | 1 |
| Platform user | Find out whether an undeployed Python-function prototype can be tested at all, then do it | COMPLETED | 1 |
| Practitioner | Make the tool-misuse strategies actually fire against a local function | COMPLETED | 1 |
| CI engineer | Run the check non-interactively in a GitHub Action — env vars, exit codes, no TTY | COMPLETED | 1 |

**Caveat, found in review and worth stating rather than burying.** All four personas ran in an environment that already had the red-team extra and a key installed, so none of them met the `ImportError` or `CredentialError` a genuinely cold reader hits. The First-timer verdict was reproducible only under that assumption. The prerequisite block added in `0fa88a02` is what closes the gap between the verdict and a real first run.

The practitioner verified the page's central claim rather than taking it on trust: with no `agent_context`, all four `tool_misuse` strategies were filtered out and the run raised "Filter selection produced zero datapoints"; with a declared `ToolInfo`, `multi_tool_chain_escalation` was selected and executed.

## Review

Round 1 was the automated reviewer loop; round 2 is @currentlycodinng's review on this PR.

|---|---|---|
| The flagship example fails cold — no key gives `CredentialError`, no extra gives `ImportError` on `huggingface-hub` | currentlycodinng | Fixed in `0fa88a02` — the extra and the key are now stated before the block, as the sibling guides do |
| Running that block uploads a real Orq Experiment with no warning | currentlycodinng | Fixed — a `!!! warning` above the block, cross-linked to `simulation-in-evaluatorq.md#what-gets-uploaded` |
| `run_receipt.sh` reads the page from the working tree but diffs the committed version, so a dirty tree silently selects the wrong blocks and still reports OK | currentlycodinng | Fixed — reads via `git show HEAD:<page>` when a base ref is set; the reviewer's 115-line dirty-tree repro now selects the right blocks |
| The bash epilogue captured only the last command's status, so a block whose real work errored reported `exit: 0` | found while fixing the above | Fixed — bash blocks run under `set -e`. This had already produced a false green in my own receipt |
| Line 192 covered `str` and `AgentResponse` but not the silent `str()` coercion of anything else | currentlycodinng | Fixed — folded into the same sentence and tied to the `content_to_text` trap below it |
| The `TypeError` is raised at `runner.py:762`, not `:492` | currentlycodinng | Fixed — `:492` was the function signature line; corrected above |
| The new table row said "you do not need to declare tools", but the section it links to shows how to declare them | conformance | Fixed — dropped the clause |
| The `simulate()` example copied the persona and scenario from `agent-simulation.md` word for word | cutter, historian | Fixed — replaced with two sentences and a cross-link |
| "This asymmetry is the one thing to remember" announces the point instead of making it | cutter | Fixed — cut |
| Readers guess `from evaluatorq.redteam import CallableTarget` and hit an ImportError | first-timer, platform user, historian | Fixed — the page now says it is **not** exported from `evaluatorq.redteam`, unlike its neighbours |
| One paragraph stacked three separate ideas | conformance | Fixed — split into three |
| The re-export sentence was phrased differently from the identical sentence about `OrqResponsesTarget` | historian | Fixed — matched the sibling's wording |
| The new row was the only one in the table to hyperlink itself | historian | Fixed — link dropped |
| An `async def usage_fn` crashes instead of degrading to `usage=None` | pedant | Fixed — verified by running it; the page now names the `ValidationError` |
| The `TypeError` example's function signature was untyped | pedant | Fixed — typed it |
| `CallableTarget` is already documented in `src/evaluatorq/redteam/README.md` | skeptic | Open — the routine may not edit `src/`, so it cannot consolidate them |
| The section re-derives mechanics the generated API reference already publishes | historian | Left out — `axes.md`'s Tier-1 rule is explicit that the generated reference never satisfies a Tier-1 gap |
| The page is over the ≤2,500 Guide guidance | conformance, currentlycodinng | Left out — the cap is guidance and the page was already over it before this change; noted as a decision below, no objection raised to the lean |
| `run_receipt.sh` re-implements the fence parser in `scripts/check_examples.py` | historian | Left out — consolidating two parsers is a refactor outside a docs PR. Noted below |
| The routine never opened a PR and wrote no ledger row | realist | Left out — checked, an artefact of the reviewer inspecting the branch at step 6, before steps 8 and 9 run |
| The `VercelAISdkTarget` gap was skipped as needing "a live Node service" | realist | Left out of the page, but the reason was wrong and is corrected below rather than recorded in the ledger |

## Open decisions

**`CallableTarget` is now documented in two places, and this routine can only maintain one of them.** `src/evaluatorq/redteam/README.md:174-191` already has a section called "Custom callable (escape hatch)" covering the `red_team()` half. That README is not published to the docs site — it is in no `mkdocs.yml` entry and nothing under `docs/` — so a reader on the site genuinely could not find it, and the gap was real. But there are now two hand-written descriptions of the same class, including two copies of the `reset_fn` caveat, and nothing ties them together. The routine is forbidden from editing anything under `src/`, so it cannot fix this. **Lean:** make the docs page canonical and cut the README section down to a link to it, in a follow-up a human owns.

**The coverage skill only greps `docs/`, so it cannot see prior art like that README.** This is how the duplicate above got written. Teaching `docs-coverage` to also search `src/**/README.md` and `examples/**/README.md` before declaring a symbol undocumented is cheap and stays outside `src/`. It is not in this PR on purpose: it changes what the routine will and will not do every future week, and an agent narrowing its own future workload is exactly the edit a human should approve rather than find. **Lean:** do it, as its own small PR.

**`CallableTarget` is not exported from `evaluatorq.redteam`.** Two of four reader personas and one critic independently wrote `from evaluatorq.redteam import CallableTarget` and hit an `ImportError`, because every other target on this page does come from there. The page now warns about it, which is a signpost around the pothole. Adding the export is a one-line `feat:` change to a shared `__init__`, which a docs PR should not carry. **Lean:** add the export, then shorten the page's imports.

**The `VercelAISdkTarget` gap was skipped for the wrong reason, and I have not put it in the ledger.** The coverage matrix said it needs "a live Vercel AI SDK HTTP endpoint (a Node service), which this environment cannot stand up". That is not true: the target is an `httpx` client that POSTs `{"messages": [...]}` and parses a documented text protocol, so a small Python stub could exercise it offline. The honest objection is different and weaker — an example verified only against a hand-rolled stub proves the client parses the protocol, not that it works against a real AI SDK route. Recording it as `skipped-no-key` would have taken it off the board permanently on a false premise, so no row was written. **Lean:** leave it eligible.

**Whether `targets.md` should be split.** It is ~3,554 words across six target kinds, over the 2,500 guideline before this PR too. **Lean:** leave it; a reader choosing between kinds wants them side by side, and the table still partitions them.

**There is no guardrail that this routine reaches its own last step.** The ledger row and the Slack message are the final two items in a long instruction list, and a run that stalls midway leaves no trace and no notification. **Lean:** worth a supervisor that posts a "run ended without reaching step 9" message if the marker is missing. Out of scope here.

**The experiments this run created.** Reviewing this PR created at least one real Experiment in `orq-research`, and this session's own runs created several more. I have not deleted any of them — removing workspace records is destructive and is a human's call. Say the word and I will list what this session created so someone can confirm before anything is removed.

## Code and docs defects found, not fixed

Reported rather than patched — the routine writes prose about code and never writes code. Each was verified by running it or by reading the source, not taken from a reviewer's word. @currentlycodinng independently re-verified all six against current HEAD.

- **`src/evaluatorq/integrations/callable_integration/target.py:116-118`** — the `agent_context` docstring says that without it "all strategies (including nonsensical ones) will be applied". That is backwards. `strategy_registry.py:108` skips any strategy with `requires_tools=True` when the context reports no tools. The docstring tells a user the opposite of what the code does.
- **`src/evaluatorq/integrations/callable_integration/target.py:154-164`** — an `async def usage_fn` is never awaited; the coroutine reaches `AgentResponse(usage=...)` and raises `ValidationError` outside the guarding `try`. Reproduced. The docstring promises degradation to `usage=None`, which only holds for a sync function that raises.
- **`src/evaluatorq/integrations/callable_integration/target.py:156`** — a return value that is neither `str` nor `AgentResponse` is silently `str()`-coerced, which `CLAUDE.md`'s own reuse table names as the anti-pattern that "renders a Python repr that a judge then scores".
- **`examples/redteam/09_custom_hooks.py:62`** — declares `on_complete(self, report, *, output_dir=None)`, but `redteam/runner.py:1193` calls it with `auto_save_path=` as well. The shipped example raises `TypeError` **after** the attacks have run and billed.
- **`ConfirmPayload.filtering_metadata`** is documented as carrying "strategy filtering metadata from datapoint generation" but is hardcoded `None` at both construction sites. The skip reasons exist only in DEBUG-level log lines.
- **`red_team()` has no `upload_results` parameter**, though `simulate()` does (`simulation/api.py:222`, default `True`). With `ORQ_API_KEY` set, every red-team run creates an Orq experiment record with no way to opt out. The page now warns readers; the missing parameter is still a `src/` gap.

## Reviewers

The `content-evaluator` opus editor that `docs-writing` names as the fourth lens **did not run** — no such agent type exists in this environment. The three narrow lenses (reader across four personas, conformance, cutter) and all four `hate` critics ran. `hate` was run in `--apply --sonnet` mode with its `architecture-reviewer` / `code-reviewer` subagent types mapped to `general-purpose`.

`routine.json` was not changed.

## Cost and workspace

`ORQ_WORKSPACE` is unset, and it would not matter if it were — it is a display setting for dashboard deep-links and routes nothing. Which workspace these records land in is decided by `ORQ_API_KEY` alone, so **this run's red-team records landed in the routine's ordinary workspace, alongside human runs.** Giving `routine.json` a key for a workspace kept for this routine would isolate it.

Spend: three receipt generations plus four reader personas running their own small red-team and simulation runs — on the order of 60 to 80 small LLM calls, all at `max_static_datapoints=1` / `max_turns=2` scope. The step-0 estimate said personas would not execute blocks; that was wrong, since `docs-writing` requires them to run commands rather than assume, and the disclosure was corrected mid-run rather than after the fact.

---

Opened by the weekly docs-autofill routine (`.claude/skills/docs-autofill/SKILL.md`).

https://claude.ai/code/session_01PHUnthpDctTs861K9BAqCt

### DIFF
--- a/.claude/skills/docs-autofill/run_receipt.sh
+++ b/.claude/skills/docs-autofill/run_receipt.sh
@@ -6,7 +6,13 @@
 # .context/ could not be reviewed. A reviewer reads this file to know what
 # "exit=0" actually meant.
 #
-# Usage: .claude/skills/docs-autofill/run_receipt.sh docs/guides/<page>.md
+# Usage: .claude/skills/docs-autofill/run_receipt.sh docs/guides/<page>.md [base-ref]
+#
+# With a base-ref, only the blocks this branch ADDED to the page are run, and the
+# receipt says so in its header. That is the "one section on an existing page"
+# case: re-running a page's pre-existing blocks bills live API calls for prose
+# nobody changed, and their pass or failure says nothing about this branch.
+# Without it, every block on the page runs — the right default for a new page.
 #
 # Languages:
 #   bash / sh     -> one logical shell: exports carry forward between blocks
@@ -22,7 +28,8 @@
 
 set -uo pipefail
 
-PAGE="${1:?usage: run_receipt.sh <page.md>}"
+PAGE="${1:?usage: run_receipt.sh <page.md> [base-ref]}"
+BASE="${2:-}"
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
 cd "$ROOT" || exit 1
 
@@ -40,6 +47,11 @@ trap 'rm -rf "$WORK" "$EVALUATORQ_DIR"' EXIT
   echo "receipt for $PAGE"
   echo "generated $(date -u +%Y-%m-%dT%H:%M:%SZ)"
   echo "EVALUATORQ_DIR=$EVALUATORQ_DIR (fresh)"
+  if [ -n "$BASE" ]; then
+    echo "scope: blocks this branch ADDED vs $BASE (pre-existing blocks on the page were not run)"
+  else
+    echo "scope: every block on the page"
+  fi
   echo
 } >>"$RECEIPT"
 
@@ -51,18 +63,37 @@ die() {
 }
 
 # Split the page into blocks: <index>\t<lang> index files land in $WORK/NNN.body
-if ! uv run python - "$PAGE" "$WORK" <<'PY'
+if ! uv run python - "$PAGE" "$WORK" "$BASE" <<'PY'
 import pathlib
+import re
+import subprocess
 import sys
 
 page, work = pathlib.Path(sys.argv[1]), pathlib.Path(sys.argv[2])
+base = sys.argv[3] if len(sys.argv) > 3 else ''
 lines = page.read_text().splitlines()
+
+# With a base ref, keep only blocks whose lines this branch added. Hunk headers
+# from a zero-context diff give those line numbers directly.
+added: set[int] = set()
+if base:
+    diff = subprocess.run(
+        ['git', 'diff', '-U0', f'{base}...HEAD', '--', str(page)],
+        capture_output=True, text=True, check=False,
+    ).stdout
+    for hunk in re.finditer(r'^@@ -\S+ \+(\d+)(?:,(\d+))? @@', diff, re.M):
+        start = int(hunk.group(1))
+        added.update(range(start, start + int(hunk.group(2) or 1)))
+    if not added:
+        print('base ref given but the diff added no lines to this page', file=sys.stderr)
+        sys.exit(1)
 
 # Track the opening fence's backtick count and close only on a run at least that
 # long — CommonMark's own rule. A 3-backtick fence inside a 4-backtick block is
 # then content, which is how this repo writes docstring samples.
-blocks, lang, buf, fence, ticks = [], None, [], None, 0
-for line in lines:
+blocks, lang, buf, fence, ticks, opened = [], None, [], None, 0, 0
+seen_total = 0
+for lineno, line in enumerate(lines, start=1):
     stripped = line.lstrip()
     run = len(stripped) - len(stripped.lstrip("`"))
     if fence is None:
@@ -71,20 +102,25 @@ for line in lines:
             ticks = run
             lang = stripped[run:].strip() or "text"
             buf = []
+            opened = lineno
         continue
     if run >= ticks and not stripped[run:].strip():
-        blocks.append((lang, "\n".join(buf)))
+        seen_total += 1
+        # An added block is one the branch touched anywhere between its fences.
+        if not base or (added & set(range(opened, lineno + 1))):
+            blocks.append((lang, "\n".join(buf), opened))
         fence, lang, buf, ticks = None, None, [], 0
         continue
     buf.append(line[fence:] if line[:fence].strip() == "" else line)
 
 index = work / "index.tsv"
 with index.open("w") as fh:
-    for i, (block_lang, body) in enumerate(blocks, start=1):
+    for i, (block_lang, body, start_line) in enumerate(blocks, start=1):
         (work / f"{i:03d}.body").write_text(body + "\n")
-        fh.write(f"{i:03d}\t{block_lang}\n")
+        fh.write(f"{i:03d}\t{block_lang}\t{start_line}\n")
 (work / "count").write_text(str(len(blocks)))
-print(f"{len(blocks)} blocks extracted", file=sys.stderr)
+scope = f' of {seen_total} on the page' if base else ''
+print(f'{len(blocks)} blocks extracted{scope}', file=sys.stderr)
 PY
 then
   die "block extraction failed — no blocks were run"
@@ -99,10 +135,10 @@ STATE="$WORK/state.sh"   # carries exports between bash blocks (one logical shel
 FAILED=0
 SEEN=0
 
-while IFS=$'\t' read -r idx lang; do
+while IFS=$'\t' read -r idx lang start_line; do
   body="$WORK/$idx.body"
   SEEN=$((SEEN + 1))
-  printf -- '--- block %s (%s)\n' "$idx" "$lang" >>"$RECEIPT"
+  printf -- '--- block %s (%s, %s:%s)\n' "$idx" "$lang" "$PAGE" "$start_line" >>"$RECEIPT"
 
   case "$lang" in
     python)
@@ -141,7 +177,11 @@ done <"$WORK/index.tsv"
 if [ "$FAILED" -ne 0 ]; then
   echo "RESULT: FAILED — at least one block exited nonzero" >>"$RECEIPT"
 else
-  echo "RESULT: OK — $EXTRACTED blocks accounted for, every executable one exited 0" >>"$RECEIPT"
+  if [ -n "$BASE" ]; then
+    echo "RESULT: OK — $EXTRACTED added blocks accounted for, every executable one exited 0" >>"$RECEIPT"
+  else
+    echo "RESULT: OK — $EXTRACTED blocks accounted for, every executable one exited 0" >>"$RECEIPT"
+  fi
 fi
 
 cat "$RECEIPT"
--- a/.claude/skills/docs-autofill/run_receipt.sh
+++ b/.claude/skills/docs-autofill/run_receipt.sh
@@ -49,6 +49,7 @@ trap 'rm -rf "$WORK" "$EVALUATORQ_DIR"' EXIT
   echo "EVALUATORQ_DIR=$EVALUATORQ_DIR (fresh)"
   if [ -n "$BASE" ]; then
     echo "scope: blocks this branch ADDED vs $BASE (pre-existing blocks on the page were not run)"
+    echo "page read from: HEAD (not the working tree — the diff's line numbers describe the committed page)"
   else
     echo "scope: every block on the page"
   fi
@@ -71,10 +72,23 @@ import sys
 
 page, work = pathlib.Path(sys.argv[1]), pathlib.Path(sys.argv[2])
 base = sys.argv[3] if len(sys.argv) > 3 else ''
-lines = page.read_text().splitlines()
 
-# With a base ref, keep only blocks whose lines this branch added. Hunk headers
-# from a zero-context diff give those line numbers directly.
+# With a base ref the line numbers come from `git diff base...HEAD`, which
+# describes the COMMITTED page. Reading the working tree here would put the two
+# in different coordinate spaces: any uncommitted edit above a section shifts
+# every offset below it, and the extractor then selects pre-existing blocks at
+# their shifted positions while still reporting the added-block count. Read the
+# committed page instead, so both sides agree even on a dirty tree — which is
+# the normal state when parallel sessions share a checkout.
+if base:
+    lines = subprocess.run(
+        ['git', 'show', f'HEAD:{page}'], capture_output=True, text=True, check=True
+    ).stdout.splitlines()
+else:
+    lines = page.read_text().splitlines()
+
+# Keep only blocks whose lines this branch added. Hunk headers from a
+# zero-context diff give those line numbers directly.
 added: set[int] = set()
 if base:
     diff = subprocess.run(
@@ -146,10 +160,25 @@ while IFS=$'\t' read -r idx lang start_line; do
       code=$?
       ;;
     bash | sh | shell)
+      # An environment-setup block is recorded, never executed. Two reasons, both
+      # about not corrupting the thing the receipt is measuring: `uv add` mutates
+      # pyproject.toml (and cannot even resolve inside this package's own
+      # checkout, where the requirement is the project), and a placeholder
+      # `export ORQ_API_KEY=...` would overwrite the real key with the literal
+      # "..." for every later block in the shared shell. These lines are
+      # instructions for the reader's own project, not claims about this repo.
+      if ! grep -qvE '^\s*(#|$|uv (add|pip install)\b|pip install\b|export [A-Z_0-9]+=(\.\.\.|<[^>]*>|"?your[^"]*"?)\s*(#.*)?$)' "$body"; then
+        printf 'command: —\nexit: — illustrative (environment setup for the reader'"'"'s project; not executed against this repo)\n\n' >>"$RECEIPT"
+        continue
+      fi
+      # `set -e` so a failing command fails the block: without it `rc=$?` sees
+      # only the last line, and a block whose real work errored still reports 0
+      # as long as it ends in something harmless like an export.
       # The epilogue goes in a file, on its own line: appending it to the body
       # inline lets a trailing comment in the block swallow it, which silently
       # drops every export the next block expects.
       {
+        printf 'set -e\n'
         cat "$body"
         printf '\nrc=$?\nexport -p >%q\nexit $rc\n' "$STATE"
       } >"$WORK/$idx.run.sh"
--- a/docs/guides/targets.md
+++ b/docs/guides/targets.md
@@ -199,7 +199,7 @@ The function may be sync or async. A sync one is run on a worker thread, so it n
 
 It receives the **full transcript** as `list[Message]` — one message on the opening turn, every prior turn afterwards — so a stateless function still sees context.
 
-Return a `str` and the wrapper boxes it into an `AgentResponse`; return an `AgentResponse` and it passes through untouched. Return anything else and it is `str()`-coerced without a warning, which is the same trap as the next paragraph's: the judge scores a Python repr as the agent's words. Convert to text yourself rather than letting that happen.
+Return a `str` and the wrapper boxes it into an `AgentResponse`; return an `AgentResponse` and it passes through untouched. Anything else is converted without a warning, and the two ways that goes wrong are worth telling apart. A `None` becomes the empty string, so the agent looks like it answered nothing and the judge scores a silence it never gave. Any other object is `str()`-coerced, which is the next paragraph's trap: the judge scores a Python repr as the agent's words. Return a string yourself rather than relying on either.
 
 `Message.content` is `str | list[ContentPart]`, not always a string. Call `content_to_text` on it as above — `str()` renders a Python repr that the judge then scores as the agent's words.
 
--- a/docs/guides/targets.md
+++ b/docs/guides/targets.md
@@ -153,7 +153,7 @@ If you write your own Responses-based target, call `evaluatorq.openresponses.inp
 
 `CallableTarget` wraps a plain Python function as an `AgentTarget`. You write a function that takes the conversation and returns the reply; the wrapper supplies the rest of the interface.
 
-Reach for it when the thing under test is already a function — an HTTP call, a local pipeline, a framework with no wrapper of its own — and you are testing what the agent *says*. Do not reach for it when the run needs to exercise tools: a callable declares none by default, and the strategy families gated on tools never fire. That case wants [your own target](#writing-your-own-target), or a `CallableTarget` with an explicit `agent_context=` (below).
+Reach for it when the thing under test is already a function — an HTTP call, a local pipeline, a framework with no wrapper of its own — and you are testing what the agent *says*. Do not reach for it when a **dynamic or hybrid** run needs to exercise tools: a callable declares none by default, and those two pipelines never fire the strategy families gated on tools. That case wants [your own target](#writing-your-own-target), or a `CallableTarget` with an explicit `agent_context=` (below).
 
 The wrapped function costs nothing to call, but the run around it does. The example below needs the red-team extra and a key — without the extra it raises `ImportError` on `huggingface-hub` when it fetches the default attack dataset, and without a key it raises `CredentialError` before the first attack:
 
@@ -199,7 +199,11 @@ The function may be sync or async. A sync one is run on a worker thread, so it n
 
 It receives the **full transcript** as `list[Message]` — one message on the opening turn, every prior turn afterwards — so a stateless function still sees context.
 
-Return a `str` and the wrapper boxes it into an `AgentResponse`; return an `AgentResponse` and it passes through untouched. Anything else is converted without a warning, and the two ways that goes wrong are worth telling apart. A `None` becomes the empty string, so the agent looks like it answered nothing and the judge scores a silence it never gave. Any other object is `str()`-coerced, which is the next paragraph's trap: the judge scores a Python repr as the agent's words. Return a string yourself rather than relying on either.
+Return a `str` and the wrapper boxes it into an `AgentResponse`; return an `AgentResponse` and it passes through untouched.
+
+**A `None` return becomes the empty string.** The agent then looks like it answered nothing, and the judge scores a silence your function never gave. Nothing warns.
+
+Any other object is `str()`-coerced instead, which is the next paragraph's trap in a different costume: the judge scores a Python repr as the agent's words. Return a string yourself rather than relying on either.
 
 `Message.content` is `str | list[ContentPart]`, not always a string. Call `content_to_text` on it as above — `str()` renders a Python repr that the judge then scores as the agent's words.
 
@@ -233,7 +237,11 @@ Wrap the function once and the same object goes to either entry point: `simulate
 
 ### What the attacker sees
 
-Given no `agent_context=`, `get_agent_context()` reports the function's `__name__` and the description `opaque callable target` — no tools, no memory stores, no instructions. That is the failure mode worth naming, because it is silent: strategies declared `requires_tools=True` are **skipped, not failed**, so a run against a callable can return a high resistance rate that means *those attacks were never attempted*. Read the attempt count, not only the rate.
+Given no `agent_context=`, `get_agent_context()` reports the function's `__name__` and the description `opaque callable target` — no tools, no memory stores, no instructions.
+
+**In dynamic and hybrid mode that is a silent failure mode.** Those pipelines pick strategies against the declared context, so a strategy marked `requires_tools=True` is **skipped, not failed** when the target reports no tools. A run can then return a high resistance rate that means *those attacks were never attempted*. Compare `report.summary.total_attacks` and `report.summary.evaluated_attacks` against what you expected to run, rather than reading the rate alone.
+
+Static mode does not filter this way at all. It replays dataset rows, which carry no strategy name, so declaring tools changes nothing about which attacks run — the same static run against the same dataset executes the same attacks whether the context declares tools or not. The example above is a static run for exactly that reason: it is the mode where a bare callable and a fully-declared one behave identically.
 
 Declare what the function can actually do and the planner writes against it:
 
@@ -270,9 +278,9 @@ print(asyncio.run(target.get_agent_context()).has_tools)
 # True — tool-gated strategies now apply
 ```
 
-### The two other hooks
+### The other two constructor options
 
-- **`reset_fn`** — a zero-argument callback invoked on `new()`. `red_team()` and `simulate()` call `new()` once per concurrent job, and a callable that closes over module-level state would otherwise carry one attack's leftovers into the next. The wrapper cannot see inside your function, so clearing that state is yours to do.
+- **`reset_fn`** — a zero-argument callback invoked on `new()`. `red_team()` and `simulate()` call `new()` once per concurrent job, and a callable that closes over module-level state would otherwise carry one attack's leftovers into the next. The wrapper cannot see inside your function, so clearing that state is yours to do. It is also the only cleanup hook you get: `CallableTarget` inherits the no-op `cleanup_memory`, so a run that mints memory entity ids logs a warning that adversarial data may persist and deletes nothing.
 - **`usage_fn`** — `(messages, response_text) -> TokenUsage | None`, for plumbing token counts out of a function that only returns a string. An exception raised inside it is logged and yields `usage=None` rather than failing the run. It must be **synchronous**, and that one is not forgiving: an `async def usage_fn` is never awaited, so the coroutine object reaches `AgentResponse` and the call dies with a pydantic `ValidationError` instead of degrading. On a page where every other function is `async def`, this is the easy mistake to make.
 
 ## Writing your own target
--- a/docs/guides/targets.md
+++ b/docs/guides/targets.md
@@ -155,6 +155,16 @@ If you write your own Responses-based target, call `evaluatorq.openresponses.inp
 
 Reach for it when the thing under test is already a function — an HTTP call, a local pipeline, a framework with no wrapper of its own — and you are testing what the agent *says*. Do not reach for it when the run needs to exercise tools: a callable declares none by default, and the strategy families gated on tools never fire. That case wants [your own target](#writing-your-own-target), or a `CallableTarget` with an explicit `agent_context=` (below).
 
+The wrapped function costs nothing to call, but the run around it does. The example below needs the red-team extra and a key — without the extra it raises `ImportError` on `huggingface-hub` when it fetches the default attack dataset, and without a key it raises `CredentialError` before the first attack:
+
+```bash
+uv add "evaluatorq[redteam]"
+export ORQ_API_KEY=...          # or OPENAI_API_KEY
+```
+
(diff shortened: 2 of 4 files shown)


.claude/skills/docs-autofill/run_receipt.sh
```diff
@@ -51,18 +63,37 @@ die() {
...
-if ! uv run python - "$PAGE" "$WORK" <<'PY'
+if ! uv run python - "$PAGE" "$WORK" "$BASE" <<'PY'
 import pathlib
+import re
+import subprocess
 import sys
 
 page, work = pathlib.Path(sys.argv[1]), pathlib.Path(sys.argv[2])
+base = sys.argv[3] if len(sys.argv) > 3 else ''
 lines = page.read_text().splitlines()
```
currentlycodinng: This reads the page from the working tree, but the added-line ranges two lines below come from `git diff base...HEAD` against the committed version. If the tree is dirty when a base ref is passed, the two coordinate spaces diverge silently. Reproduced: inserting 115 uncommitted lines above a section and re-running with the same base pulled the page's pre-existing blocks at their shifted working-tree offsets instead of the branch's new ones, and the script still printed RESULT: OK. Read via `git show HEAD:"$PAGE"` when BASE is set, or `git diff --quiet HEAD -- "$PAGE" || die "working tree is dirty"`.

docs/guides/targets.md
```diff
@@ -148,10 +149,128 @@ Each `respond(messages)` sends the **full transcript**. Nothing is stored on the
 
 If you write your own Responses-based target, call `evaluatorq.openresponses.input_items.messages_to_responses_input` rather than reproducing this.
 
+## The escape hatch: `CallableTarget`
+
+`CallableTarget` wraps a plain Python function as an `AgentTarget`. You write a function that takes the conversation and returns the reply; the wrapper supplies the rest of the interface.
+
+Reach for it when the thing under test is already a function — an HTTP call, a local pipeline, a framework with no wrapper of its own — and you are testing what the agent *says*. Do not reach for it when the run needs to exercise tools: a callable declares none by default, and the strategy families gated on tools never fire. That case wants [your own target](#writing-your-own-target), or a `CallableTarget` with an explicit `agent_context=` (below).
+
+```python
```
currentlycodinng, line 168: Ran this cold: no ORQ_API_KEY/OPENAI_API_KEY gives CredentialError, and even with a key but without the redteam extra it's ImportError on huggingface-hub. red-teaming.md:49 and agent-simulation.md:11 both state their prerequisite (extra + key) before the first block; this section is the only one on the page that doesn't. Running this block also uploads a real Experiment to Orq with no warning here, since red_team() has no upload_results opt-out, worth one sentence and a link to simulation-in-evaluatorq.md's equivalent note.

docs/guides/targets.md
```diff
@@ -148,10 +149,128 @@ Each `respond(messages)` sends the **full transcript**. Nothing is stored on the
...
+
+
+asyncio.run(main())
+```
+
+The function may be sync or async. A sync one is run on a worker thread, so it never blocks the event loop.
+
+It receives the **full transcript** as `list[Message]` — one message on the opening turn, every prior turn afterwards — so a stateless function still sees context.
+
+Return a `str` and the wrapper boxes it into an `AgentResponse`; return an `AgentResponse` and it passes through untouched.
```
currentlycodinng, line 202: This covers str and AgentResponse but not the third case: target.py:156 silently str()-coerces anything else, which is the same failure mode the next line warns about for Message.content (a repr scored by a judge as the agent's words). Worth folding into this sentence rather than leaving it implicit.

currentlycodinng (changes requested):
Went through this treating the receipt table and the persona verdicts as claims, not evidence, same as any other self-reported "already reviewed" framing. The receipt itself is honest: I extracted run_receipt.sh's block-selection logic and ran it standalone against HEAD~1, it selected exactly targets.md:158/202/230, and running each by hand reproduces the claimed exit code and first line verbatim, including the HuggingFace download line and the literal TypeError text. mkdocs build --strict and validate_mermaid.py both pass. Every one of the six "found, not fixed" src/ defects checks out against current HEAD (target.py's docstring really is backwards on strategy filtering, the async usage_fn bug reproduces exactly as described, ConfirmPayload.filtering_metadata really is hardcoded None at both construction sites, ok9_custom_hooks.py really does TypeError after the attacks bill). One citation is off: the TypeError is raised at redteam/runner.py:762, not :492, which is the function signature line.

Two things before this merges.

run_receipt.sh reads the page from the working tree (`page.read_text()`, line 74) but computes added-line ranges from `git diff base...HEAD` (line 81), against the committed page. Those two only agree because this branch's tree was clean when the receipt ran. I reproduced the skew directly: with 115 uncommitted lines inserted above the CallableTarget section and the same base ref, the extractor pulled the *pre-existing* blocks at their new working-tree offsets instead of the branch's three new ones, and still printed `RESULT: OK — 3 added blocks accounted for`. The tool that exists specifically to make "only re-run what changed" trustworthy has a path where it silently re-runs the wrong blocks and reports success. Read the page via `git show HEAD:$PAGE` when BASE is set, or refuse a dirty tree.

The flagship example at targets.md:158 fails cold for the reader it targets. Clean install, no key: `CredentialError: Missing LLM credentials`. With a key, no extra: `ImportError: ... requires the huggingface-hub package. Install the redteam extra`. Every sibling guide carries this upfront, red-teaming.md:49 and agent-simulation.md:11 both state the extra and the key before the first block. This is the only section on the page that doesn't, which also means the "First-timer: COMPLETED, 1 round" verdict isn't reproducible outside an environment that already has both in place.

Related and smaller: running that same block creates a real Orq experiment record and prints its dashboard URL, since red_team() has no upload_results opt-out (the PR correctly files this as a src/ defect). The section never warns the reader, where simulation-in-evaluatorq.md does for simulate(). One sentence, cross-linked, covers it. I ran this block during review and it did create a live record in the orq-research workspace under the key it found, worth checking whether that experiment should be deleted.

Smaller still: line 192 says a non-str return "passes through untouched" if it's an AgentResponse but doesn't say what happens otherwise, and target.py:156 silently str()-coerces anything that's neither, which is the exact repr-fed-to-a-judge trap the very next line warns about for Message.content. Worth folding into the same sentence.

On the open decisions the PR already surfaced: the two-copies-of-CallableTarget question and the ~3,554-word length (over the stated 2,500 guideline before this PR too) both read as genuine judgment calls outside what a docs-only routine can resolve, no objection to the leans stated.

.claude/skills/docs-autofill/run_receipt.sh
Baukebrenninkmeijer: Fixed in 0fa88a02, taking your first option: when `BASE` is set the extractor now reads the page via `git show HEAD:<page>`, so both the content and the diff's line numbers describe the committed page. I preferred that over the dirty-tree guard because parallel worktrees make a dirty tree the normal state here, and refusing would just make the tool unusable in the case it is most needed.

I reproduced your scenario against the fix — 115 uncommitted lines inserted above the section, same base ref — and it now selects the branch's own blocks at their committed offsets (158/202/230 at the time) rather than the shifted pre-existing ones. The receipt header states which side it read from, so a future reader does not have to infer it.

Your comment also turned up a second bug in the same file, which I have fixed alongside it. The bash epilogue captures `rc=$?` after the block body, so only the *last* command's status reached the receipt. My own newly added prerequisite block ended in a successful `export` after a failing `uv add`, and the receipt duly recorded `exit: 0` for a block that had errored — the same "silently reports success" shape you found. Bash blocks now run under `set -e`.

That fix then exposed the third thing: `uv add "evaluatorq[redteam]"` cannot resolve inside this package's own checkout (`self-dependencies are not permitted`), and a placeholder `export ORQ_API_KEY=...` would overwrite the real key with the literal `...` for every later block in the shared shell. So the runner now records environment-setup blocks without executing them, and the receipt row says so explicitly rather than claiming a pass. Flagging that as a new exemption for you to check rather than letting it pass unremarked, since it is exactly the kind of thing the exemption is easy to abuse for.

---

docs/guides/targets.md
Baukebrenninkmeijer: Fixed in 0fa88a02. The section now states the prerequisite before the first block, in the same shape the siblings use, and names both failure modes you hit so a reader recognises them:

> The wrapped function costs nothing to call, but the run around it does. The example below needs the red-team extra and a key — without the extra it raises `ImportError` on `huggingface-hub` when it fetches the default attack dataset, and without a key it raises `CredentialError` before the first attack.

The upload warning is in as a `!!! warning` immediately above the block, saying that `red_team()` has no `upload_results=False`, that pointing `ORQ_API_KEY` elsewhere is therefore the only way to keep a run off a shared workspace, and linking [What gets uploaded](https://github.com/orq-ai/evaluatorq/blob/docs/autofill-callable-target/docs/guides/simulation-in-evaluatorq.md#what-gets-uploaded) as you suggested.

You are right that this also undercuts the "First-timer: COMPLETED, 1 round" verdict I reported. That persona ran in an environment that already had the extra and a key, so it never met either error — the verdict was reproducible only there, and I should not have presented it without that qualification.

On the experiment your review run created in `orq-research`: I have deliberately not deleted it. Deleting workspace records is destructive and outside what this routine should do on its own, and I would rather you decide. Mine are in there too — this run created several, which the PR body discloses. If you want them cleaned up, say so and I will list what this session created so you can confirm before anything is removed.

---

docs/guides/targets.md
Baukebrenninkmeijer: Fixed in 0fa88a02 — folded into the same sentence rather than left implicit, and tied to the paragraph that follows it so the two read as one trap:

> Return a `str` and the wrapper boxes it into an `AgentResponse`; return an `AgentResponse` and it passes through untouched. Return anything else and it is `str()`-coerced without a warning, which is the same trap as the next paragraph's: the judge scores a Python repr as the agent's words. Convert to text yourself rather than letting that happen.

The underlying `str()` coercion at `target.py:156` stays filed as a `src/` defect in the PR body, since this routine may not touch code.

---

currentlycodinng (approved):
Re-verified at 0fa88a02, all three genuinely fixed, not just reworded.

run_receipt.sh:83-86 now reads the page via `git show HEAD:$page` when a base ref is set, instead of the working tree. Reran my original repro against the fixed script with 200 uncommitted lines inserted above the section: extraction is byte-identical to the clean-tree run, both correctly pull the branch's new blocks regardless of what's sitting uncommitted in the tree. Also checked it doesn't break on a page created (not modified) in the branch, or a base predating the page's existence, both fine.

targets.md:158 now states the prerequisite before the first block, in the same shape as red-teaming.md:49 and agent-simulation.md:11, and the warning at :165-166 accurately says red_team() has no upload_results opt-out, cross-linked to simulation-in-evaluatorq.md. Ran both failure modes cold to confirm the prose matches reality: no key raises CredentialError before any attack, a key without the redteam extra raises ImportError fetching the default dataset, matching what's now written.

:202 now names the str()-coercion case explicitly and it matches target.py:156.

mkdocs build --strict and validate_mermaid.py both still pass.

Two small things, not blocking, take or leave: the fix also added a pattern-based exemption so environment-setup blocks like `uv add ...` and `export ORQ_API_KEY=...` skip execution, which is reasonable (running uv add would mutate pyproject.toml, and the literal `...` in the export would poison the shell) but sits outside SKILL.md:108's current definition of `illustrative`, which is scoped to blocks that can't be executed by nature and explicitly warns against using it to retire a block that merely fails. Worth a sentence there so the tracked contract and the runner agree, since this is exactly the kind of exemption class that's easy to widen later without noticing. And :202's "the judge scores a Python repr" doesn't quite hold for a `None` return, which target.py:156 maps to an empty string rather than a repr, different silent trap (the agent appears to answer nothing) but same paragraph.

Baukebrenninkmeijer:
Thanks for the re-verification, and for testing the two cases I had not — a page created rather than modified in the branch, and a base predating the page's existence. Both of your "take or leave" notes were right, so I took them. Pushed as `e305a143`.

**The exemption now matches its contract.** You are right that the setup-block class sat outside `SKILL.md:108`, which scopes `illustrative` to blocks non-executable by nature and warns against using it to retire a block that merely fails. Leaving the tracked contract and the runner disagreeing is the worse half of the problem, since the contract is what a future run reads. `SKILL.md` now names the second class explicitly, and states the distinction that keeps it narrow: these blocks are exempt because executing them is *wrong*, not because it is hard — `uv add` mutates `pyproject.toml`, and the placeholder export would poison the shared shell. It repeats your concern in the file itself, that a block doing real work and happening to fail is not setup. The runner decides by pattern rather than on an author's say-so, which is the part that stops it widening quietly.

**The `None` case was a real inaccuracy, not a wording quibble.** I checked it before rewriting: `None` returns `''` and any other object returns its repr, so the sentence I wrote described only the second and quietly mis-described the first. It now separates them, because they fail differently for the reader — an empty string makes the agent look like it answered nothing and the judge scores a silence it never gave, which is arguably the more confusing of the two.

Receipt regenerated (4 blocks, 3 executed and green, 1 recorded as setup), `mkdocs build --strict` and `validate_mermaid.py` pass, `src/` still untouched.

Flagging one thing rather than assuming: this push lands after your approval, so it may have reset it. Both changes are the ones you named, but `SKILL.md` is a contract file rather than prose, so it is worth your eyes rather than mine.

The six `src/` defects remain filed and unfixed by design — this routine may not touch code. The `orq-research` Experiment records from your review run and from mine are also still there; I have not deleted any, and will list what this session created if you want them cleaned up.

---

docs/guides/targets.md
```diff
@@ -148,10 +149,138 @@ Each `respond(messages)` sends the **full transcript**. Nothing is stored on the
 
 If you write your own Responses-based target, call `evaluatorq.openresponses.input_items.messages_to_responses_input` rather than reproducing this.
 
+## The escape hatch: `CallableTarget`
+
+`CallableTarget` wraps a plain Python function as an `AgentTarget`. You write a function that takes the conversation and returns the reply; the wrapper supplies the rest of the interface.
+
+Reach for it when the thing under test is already a function — an HTTP call, a local pipeline, a framework with no wrapper of its own — and you are testing what the agent *says*. Do not reach for it when the run needs to exercise tools: a callable declares none by default, and the strategy families gated on tools never fire. That case wants [your own target](#writing-your-own-target), or a `CallableTarget` with an explicit `agent_context=` (below).
```
Baukebrenninkmeijer: The tool-gating claim is true only for the dynamic and hybrid pipelines. The filter lives in `redteam/adaptive/strategy_registry.py:108`, which is the dynamic leg; `redteam/runner.py:3166` explicitly ignores the strategy filter in static mode. Suggest scoping it: "the strategy families gated on tools never fire in the dynamic and hybrid pipelines".

docs/guides/targets.md
```diff
@@ -148,10 +149,138 @@ Each `respond(messages)` sends the **full transcript**. Nothing is stored on the
...
+
+
+asyncio.run(main())
+```
+
+The function may be sync or async. A sync one is run on a worker thread, so it never blocks the event loop.
+
+It receives the **full transcript** as `list[Message]` — one message on the opening turn, every prior turn afterwards — so a stateless function still sees context.
+
+Return a `str` and the wrapper boxes it into an `AgentResponse`; return an `AgentResponse` and it passes through untouched. Anything else is converted without a warning, and the two ways that goes wrong are worth telling apart. A `None` becomes the empty string, so the agent looks like it answered nothing and the judge scores a silence it never gave. Any other object is `str()`-coerced, which is the next paragraph's trap: the judge scores a Python repr as the agent's words. Return a string yourself rather than relying on either.
```
Baukebrenninkmeijer: Four ideas in one paragraph: `str` boxing, `AgentResponse` passthrough, `None` to empty string, and `str()` coercion. Worth splitting the `None` case into its own sentence or line — it is the one that silently corrupts a judge score, and it is currently the third clause of a long paragraph.

docs/guides/targets.md
```diff
@@ -148,10 +149,138 @@ Each `respond(messages)` sends the **full transcript**. Nothing is stored on the
...
+    # Invalid target type: function. Expected str or AgentTarget.
+```
+
+The error is immediate and loud, which is the good case — it raises on the way in rather than mid-run.
+
+Wrap the function once and the same object goes to either entry point: `simulate(target=CallableTarget(support_agent), ...)` takes it unchanged. Build the personas, scenarios and criteria around it exactly as in [Agent Simulation](agent-simulation.md) — swapping `target=` is the only difference from the examples there, so wrapping the function yourself is what buys you one target object for both surfaces.
+
+### What the attacker sees
+
+Given no `agent_context=`, `get_agent_context()` reports the function's `__name__` and the description `opaque callable target` — no tools, no memory stores, no instructions. That is the failure mode worth naming, because it is silent: strategies declared `requires_tools=True` are **skipped, not failed**, so a run against a callable can return a high resistance rate that means *those attacks were never attempted*. Read the attempt count, not only the rate.
```
Baukebrenninkmeijer: Two fixes here.

First, scope the claim. Strategy filtering by `requires_tools` happens in the dynamic and hybrid pipelines only — `redteam/runner.py:3166` ignores the strategy filter entirely in static mode. The example directly above this section runs `mode="static"`, so a reader runs it, reads this, and builds the wrong model of what just happened.

Second, "read the attempt count" does not name a field. The reader has a `RedTeamReport` in hand: say `report.summary.total_attacks` (and `evaluated_attacks`). One word of prose becomes actionable.

docs/guides/targets.md
```diff
@@ -148,10 +149,138 @@ Each `respond(messages)` sends the **full transcript**. Nothing is stored on the
...
+            )
+        ],
+    ),
+)
+
+print(asyncio.run(target.get_agent_context()).has_tools)
+# True — tool-gated strategies now apply
+```
+
+### The two other hooks
```
Baukebrenninkmeijer: "Hooks" collides with `hooks=PipelineHooks`, a real `red_team()` parameter, so a reader arriving here from the signature expects something else. "The other two constructor options" says what these are.

Baukebrenninkmeijer (commented):
Reviewed the diff and verified each claim against source, plus ran the examples. The addition is sensible and mostly accurate — the `async usage_fn` trap and the `red_team()`-refuses-a-bare-function contrast are both real and worth having on the page. Confirmed independently: the import path and constructor signature, `new()`-per-job via `BareTargetBackend.create_target`, the exact `TypeError` text, the default context (`key=<fn name>`, `description='opaque callable target'`, `has_tools=False`), the absence of `upload_results` on `red_team()`, the `evaluatorq.simulation` / `evaluatorq.redteam` export asymmetry, and a clean `mkdocs build --strict` with both anchors resolving.

Three things to fix — two inline, one below.

**Docstring drift the PR should take with it.** `src/evaluatorq/integrations/callable_integration/target.py:117` still says that without `agent_context` "all strategies (including nonsensical ones) will be applied". That is the opposite of what this page now says, and the page is right: `adaptive/strategy_registry.py:108` skips `requires_tools` strategies when the context reports no tools. Same-PR fix per the docs rule in CLAUDE.md. (Not inline-commentable — the file is not in this diff.)

**Smaller, take or leave:**

- Nothing tells the reader that `CallableTarget` cannot clean up memory entities. `BareTargetBackend.cleanup_memory` warns loudly at runtime when a target minted memory ids but inherits the no-op default. One clause on the `reset_fn` bullet would cover it.
- The `evaluatorq.redteam` export asymmetry is documented honestly, but the actual fix is a one-line export. Worth a follow-up ticket rather than prose.

**On the skill and runner changes:** sensible and narrow. Deciding the setup exemption by pattern in the runner rather than by an author's claim is the right call, and the `set -e` addition closes a real hole — a block ending in an `export` used to report 0 after its real work had failed. Reading the page from `HEAD` when a base ref is given is correct, and the comment explains why better than most. I tested the exemption regex under both `ugrep` and `/usr/bin/grep` on macOS; it matches as intended in both. One note, not a request: a setup block written `export ORQ_API_KEY=sk-abc123` — a realistic-looking placeholder rather than `...`, `<...>` or `your...` — falls outside the pattern and gets executed. That errs in the conservative direction, so leave it; just don't be surprised by it later.

docs/guides/targets.md
Baukebrenninkmeijer: You are right, and this was the worst thing on the page — a wrong model handed to the reader at exactly the moment they had just run the counterexample. Fixed in `e2ca6b1c`.

I traced it rather than taking the scoping on trust: `select_applicable_strategies_for_vulnerability` (the `requires_tools` gate at `strategy_registry.py:108`) is reached from `runner.py:2275`, which sits inside `_run_dynamic_or_hybrid` at `runner.py:1992`. Nothing on the static leg calls it, and `runner.py:3166` says why — static rows carry no strategy name at all.

Then I confirmed it end to end, because a trace can miss a second call path. Same static run, same vulnerability, once with a declared `ToolInfo` and once without:

```
STATIC no tools: total_attacks=2
STATIC tools   : total_attacks=2
```

Identical, so declaring tools changes nothing in static mode. The section now splits the two: the silent-skip warning is scoped to dynamic and hybrid, and static gets its own paragraph saying it replays dataset rows and behaves the same either way — including a note that the example above is static *for that reason*, so the contrast is the point rather than an accident.

The attempt-count sentence now names `report.summary.total_attacks` and `report.summary.evaluated_attacks`, and says to compare them against what you expected to run.

Worth recording how this got through: the practitioner persona that verified the tool-gating claim ran it in **dynamic** mode and saw exactly the skip behaviour I documented. I then generalised a dynamic-mode observation into an unqualified claim, and wrote a static example underneath it. The empirical check was real but did not cover the mode the page actually demonstrates.

---

docs/guides/targets.md
Baukebrenninkmeijer: Fixed in `e2ca6b1c`, scoped as you suggested. The scope line now reads "Do not reach for it when a **dynamic or hybrid** run needs to exercise tools: a callable declares none by default, and those two pipelines never fire the strategy families gated on tools."

Verification and the full trace are in [my reply on the other thread](https://github.com/orq-ai/evaluatorq/pull/193#discussion_r3948294816) — short version: the gate is reached only from `_run_dynamic_or_hybrid`, and a static run executes the same two attacks whether tools are declared or not.

---

docs/guides/targets.md
Baukebrenninkmeijer: Fixed in `e2ca6b1c`. The `None` case is now its own paragraph with its own emphasis, rather than the third clause of four:

> **A `None` return becomes the empty string.** The agent then looks like it answered nothing, and the judge scores a silence your function never gave. Nothing warns.

The `str`/`AgentResponse` pair stays one sentence above it, and the repr coercion gets a third. You are right that `None` is the one that deserves the separation — an empty string reads as a real answer to a judge, where a repr at least looks obviously wrong to anyone reading the transcript.

---

docs/guides/targets.md
Baukebrenninkmeijer: Fixed in `e2ca6b1c` — renamed to "The other two constructor options", your wording. You are right that "hooks" was actively misleading next to a real `hooks=PipelineHooks` parameter on the same function.

I also took your memory-cleanup note in the same edit, on the `reset_fn` bullet since that is where a reader looks for cleanup: `CallableTarget` inherits the no-op `cleanup_memory`, so a run that mints memory entity ids logs a warning that adversarial data may persist and deletes nothing (`BareTargetBackend.cleanup_memory`, `backends/base.py:194`).

---

Baukebrenninkmeijer:
All four inline items fixed in `e2ca6b1c`, replies on each thread. The static-mode scoping was the serious one and I have written up the trace and the A/B run [there](https://github.com/orq-ai/evaluatorq/pull/193#discussion_r3948294816); thank you for catching it, because the page was teaching the wrong model directly underneath a static example.

**On the docstring at `target.py:117` — I am not going to do it, and I want to be plain about why rather than quietly skipping it.**

You are right on the substance. That docstring says the opposite of what the code does, the page is correct, and CLAUDE.md's docs rule would normally put the fix in this PR. But this PR is written by the weekly docs-autofill routine, and the routine's governing skill states one boundary without exceptions:

> **Never edit `src/`. No exception, not even a one-character docstring fence.** A routine that can reach the code it documents will eventually document its way into changing behaviour, and the reviewer of a docs PR is not reading it for that. This is the hard boundary of the whole routine: it writes prose about code, and it never writes code.

A one-line docstring correction is exactly the edit that boundary is drawn to refuse — it is obviously safe in isolation, which is what makes it the precedent that erodes the rule. A reviewer skimming a docs PR is not reading `src/` diffs, so the cost of being wrong there is paid by someone who was not looking.

So the defect stays filed rather than fixed. It is the first entry under "Code and docs defects found, not fixed" in the PR body, with the file, the line, and `strategy_registry.py:108` as the contradicting source. It also went to `#research-assistant` when the run finished, so it has already reached someone who *can* touch `src/`.

If you would like it fixed now, the cleanest paths are: you push the one-line change onto this branch yourself, or I open a separate PR for it in a session that is not running under this routine. Say which and I will do it — I just cannot reach across that line from inside this run.

**Two smaller things you raised:**

- The `evaluatorq.redteam` export asymmetry — agreed it wants a one-line export rather than prose, and it is already an open decision in the PR body with that as the lean. I have not opened a ticket, since the routine has no Linear access; happy to if you want it tracked rather than living in the PR.
- The exemption regex and `export ORQ_API_KEY=sk-abc123` — agreed, and I have left it. A realistic-looking placeholder getting executed fails toward running a block rather than skipping one, which is the direction that cannot hide a broken example. Noting it here so the next person to touch that pattern finds the reasoning rather than rediscovering it.

CI is green on `e2ca6b1c`, `src/` untouched, receipt regenerated (4 blocks: 3 executed, 1 recorded as setup).

---

PR #194, opened 2026-09-07
fix(simulation): deliver evaluator failures through lifecycle hooks

Baukebrenninkmeijer:
## Problem

The first implementation of simulation evaluator results made important failures hard to observe and lifecycle behavior inconsistent. With fail-fast enabled, an evaluator could fail without its completion hook receiving the failure before `SimulationDroppedError` escaped. A result could also contain a numeric score without the evaluator's explanation, malformed criteria metadata could be reported as verified, and JSONL export could discard structured evaluator detail.

The practical failure looked like this:

```python
def on_evaluator_complete(event):
    record(event.result.error, event.result.raw_output)

try:
    await simulate(dataset, evaluators=[evaluator], exit_on_failure=True, on_evaluator_complete=on_evaluator_complete)
except SimulationDroppedError:
    pass
```

Before this follow-up, the callback could be skipped or receive incomplete detail, so callers could not reliably explain why a datapoint failed or reconstruct the evaluator output from a saved JSONL run.

## Solution

This follow-up makes evaluator failures part of the normal simulation result and lifecycle contract.

- Evaluator-completion callbacks run before fail-fast `SimulationDroppedError` propagation, so observers see the evaluator outcome.
- The primary simulation or drop error is captured explicitly, remains primary when lifecycle cleanup also raises, and cannot be replaced by an unrelated exception already being handled by the caller.
- Evaluator result contracts are enforced: duplicate evaluator names are rejected, malformed or otherwise unverified criteria evidence clears top-level `criteria_verified`, and non-numeric scores produce a derived error instead of silently looking valid.
- `SimulationResult.evaluator_details` preserves each evaluator's structured `raw_output`, including criteria evidence and conversation-quality component breakdowns.
- Existing numeric scores remain in `metadata["evaluator_scores"]` for report and dashboard readers.
- JSONL export retains the evaluator detail so a persisted run contains the same explanation as the in-memory result.
- JSONL export stringifies non-JSON-native custom evaluator detail instead of dropping the entire export.

## Verification

```text
uv run ruff check src
uv run basedpyright
```

All checks pass: 5,326 tests passed, 2 skipped, and 9 deselected; formatting, type checking, and strict documentation build also pass.

## Scope

Reports and dashboard projections for the new structured evaluator detail are intentionally deferred; existing consumers continue reading numeric scores from `metadata["evaluator_scores"]`.


src/evaluatorq/simulation/api.py
```diff
@@ -2351,9 +2384,11 @@ def _sim_evaluation_raw_output(name: str, result: SimulationResult, value: objec
             else:
                 return None
 
-        reason = unverified_reason(result)
+        reason = f'criteria_meta_invalid={len(invalid)}' if invalid else unverified_reason(result)
         if 'criteria_results' in raw and reason is None:
             reason = 'criteria_meta missing (lossy dict)'
+        if reason is not None:
+            result.criteria_verified = False
```
currentlycodinng, line 2397: This also fires on the lossy-dict branch (raw['criteria_results'] present, no criteria_meta, reason set two lines above), which previously left the top-level flag alone. Since unverified_reason(result) at :2387 reads this same flag, repeated calls on the same result now flip the reason string, verified: first call gives 'criteria_meta missing (lossy dict)', second gives 'criteria_verified=False'. Worse than a string flip: a criteria_met score of 1.0/PASS on the first pass reads 0.0/unverified on a second pass over the same object, since the flag this line sets is what the next pass's reason computation reads. Needs the same fix already applied to the invalid-metadata branch, snapshot reason before the write, or scope this write to only the invalid case.

src/evaluatorq/simulation/api.py
```diff
@@ -2576,25 +2611,48 @@ def _stamp_evaluator_scores(
...
+                seen_names.add(score.evaluator_name)
                 value = score.score.value
                 numeric = isinstance(value, (int, float)) and not isinstance(value, bool)
                 reason = (
                     score.error if score.error is not None else (None if numeric else f'non-numeric value {value!r}')
                 )
+                if reason is not None and score.error is None:
+                    score.error = reason
+                if score.score.raw_output is not None:
+                    sim_result.evaluator_details[score.evaluator_name] = deepcopy(score.score.raw_output)
```
currentlycodinng: deepcopy can raise on raw_output shapes the previous by-reference store tolerated (locks, anything unpicklable), which would abort here before results is assembled two lines below. Narrow today (only a custom evaluator in the public SIMULATION_EVALUATORS export could hit this), but it's the same population the export's new default=str fix was written to handle safely, worth picking one failure mode for that input rather than two commits disagreeing on it.

src/evaluatorq/simulation/api.py
```diff
@@ -2576,25 +2611,48 @@ def _stamp_evaluator_scores(
...
         sim_result = result_cache.get(id(dp_result.data_point))
         if sim_result is None or not dp_result.job_results:
             continue
+        seen_names = seen_names_by_result.setdefault(id(sim_result), set())
         for job_result in dp_result.job_results:
             for score in job_result.evaluator_scores or []:
-                if events_out is not None:
-                    events_out.append((sim_result, score))
+                if score.evaluator_name in seen_names:
+                    raise ValueError(f'duplicate evaluator name for simulation result: {score.evaluator_name!r}')
```
currentlycodinng: This raises before results is built at :2537, so a duplicate name discards an otherwise-completed run rather than logging and keeping the first score. Not reachable today since only one job is passed in, but it now contradicts _notify_evaluator_complete's docstring a few lines below, which still describes two same-named JobResults as an expected case that produces two events. Worth reconciling the doc and the guard either way.

currentlycodinng (changes requested):
Re-reviewed at 6aeec212. ruff/format/basedpyright clean, pytest -m 'not integration' gives 5310 passed / 4 skipped, matching this run's own claims within normal drift.

Of the 6 open items: the primary_error regression is properly fixed, walked every set/read of it through the finally block and it no longer picks up a caller's unrelated exception. The dead condition is genuinely simplified. The aliasing and duplicate-name-collision pair from evaluator_details is fixed on both halves, deep copy is real (verified by mutating a nested dict and confirming the source EvaluatorScore is untouched), and same-name collision now raises instead of silently overwriting. The jsonl export handles datetime/Decimal/set/custom-repr without dropping the batch. The identity assertion and the combined-case test are both real and both pass.

One of the six moved rather than closed, and it's the one that matters most since it changes recorded scores.

api.py:2390-2391 now sets `result.criteria_verified = False` whenever `reason is not None`, which also fires on the lossy-dict branch two lines above it (`raw['criteria_results']` present, no criteria_meta). That branch used to leave the top-level flag alone. Now the first call sets it False, and since unverified_reason(result) two lines up reads that same flag, a second call on the same result produces a different reason string, exactly the non-idempotence this thread already flagged once for the invalid-metadata branch. I reproduced it end to end: repeated calls on a lossy-dict result now read `'criteria_meta missing (lossy dict)'` then `'criteria_verified=False'`, where 5919b22d was stable across repeats. And it isn't just a string, a re-scored criteria_met result that returned 1.0 / PASS on the first pass returns 0.0 / "unverified, not met" on a second, since the score computation before this function runs and the recorded flag after it now disagree with each other across passes on the same object. Reports, the dashboard, and sim_views.py all read this flag downstream. This needs the same treatment the invalid branch got, snapshot reason before writing the flag, or don't let the lossy branch feed back into unverified_reason after the write.

Two smaller ones from testing the same seam. The new deepcopy at api.py:2653 can now raise on a raw_output shape the old by-reference store tolerated fine (a lock, anything unpicklable), which would kill an otherwise-successful run at the point of stamping, before results is even assembled two lines below at :2537, the same one-error-hides-the-work shape as the primary_error bug. Only reachable through a custom evaluator registered into the public export today, but the export widening a few lines away was written for exactly that population, so the two fixes are pulling in different directions on the same input. And the new duplicate-name ValueError at :2643 fires before results is built at :2537, so a completed, paid-for run gets discarded on an anomaly in score shape rather than logged and the first score kept, not unreachable today since only one job is ever passed in, but the guard now directly contradicts _notify_evaluator_complete's own docstring three lines below still describing two same-named JobResults as an expected, handled case.

Not flagging as a bug, but worth a decision: the None -> False widening for criteria_verified on lossy-dict and terminated runs is a semantic change to what that field means for existing saved runs, not purely a correctness fix, want to confirm that's intended before it ships.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Confirmed and fixed in 8f64238. The write-back is now scoped to the reasons that are re-derivable from the result (`terminated_by`, an already-false flag, invalid metadata); a missing audit alone no longer sets `criteria_verified=False`, so a repeat call reports `criteria_meta missing (lossy dict)` rather than reading back its own flag. Regression test: `test_criteria_met_lossy_raw_output_is_stable_across_repeat_scoring`.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Fixed in 8f64238. The `deepcopy` is wrapped: an uncopyable `raw_output` is stored by reference with a warning naming the cause, so the run survives with the same degrade-not-abort posture as the export's `default=str`. Covered by `test_uncopyable_raw_output_is_stored_by_reference`.

src/evaluatorq/simulation/api.py
Baukebrenninkmeijer: Fixed in 8f64238. A duplicate name now logs a warning and keeps the first score/detail instead of raising, and the duplicate is still appended to `events_out` — so `_notify_evaluator_complete`'s "one event per score" docstring holds as written. Covered by `test_duplicate_evaluator_scores_keep_the_first_and_still_emit_events`.

PR #195, opened 2026-09-07
fix: price orq/* router calls at the served model (RES-1573)

currentlycodinng:
`orq/*` routers resolve to a real model per request. Your workspace already has fourteen of them live (`orq/autorouter-anthropic-balanced`, `orq/autorouter-openai-cost`, …), and RES-1520 (orq-ai/orquesta-web#11438) adds the zero-config `orq/auto` and `orq/frontier` family on the same pattern. Nothing was needed to *reach* them from here — a run holding `ORQ_API_KEY` already talks to `/v3/router` — but three things behaved wrongly once one was named.

## Chat Completions were billed 4.55x over, today, on a live router

`price_usage` looked up the id the caller asked for. A router **is** in the catalogue, but at a single headline rate standing in for every model it can pick, so the call was charged at a rate nothing served. The router's own accounting is the ground truth here, and the Responses endpoint reports it, so the same usage can be priced three ways and compared:

```
orq/autorouter-anthropic-balanced → aws/eu.anthropic.claude-haiku-4-5   (13 in / 38 out)
  the router billed        $0.0002233
  priced at served model   $0.0002233   1.00x   ← this PR
  priced at requested id   $0.001015    4.55x   ← main
```

Pricing now goes to the model that answered (`served_model=`), falling back to the requested id so a provider answering under a dated snapshot id keeps the price it used to get. An `orq/*` request never takes that fallback, since its requested id is the headline rate: an unlisted or unreported served model, or a served id that is itself another router, leaves the call unpriced with one warning saying which. A rate the caller set for the router with `register_model` still applies, and a listed served model outranks it. The comparison strips the provider prefix, because the router answers `openai/gpt-5.6-luna` with the bare `gpt-5.6-luna` and that is the same model, not a reroute. A guardrail in `test_reuse_guardrails.py` fails any `price_usage` call in the package that names no `served_model`.

The Responses path is unaffected in practice — the router prices those itself, so `price_usage` is already a no-op there — but it passes `served_model=` too, so a router that ever answers unpriced lands on the right rate rather than the headline one.

## Prompt caching

`caching_applies` now includes `orq/*`, for the reason it already includes `agent/<key>`: the resolved model is not knowable when the breakpoint is written, and the router ignores rather than rejects the marker. Not hypothetical — the router measured above resolved to Anthropic.

## The panel boundary

A single judge may be a router; a jury may not. A router optimises each request alone and can seat three cards from one vendor, which is the correlated error a panel exists to cancel. Seats stay pinned even though the frontier arithmetic behind them now also runs server-side. Written into the `jury_presets` docstring and the jury guide rather than left to the next reader.

Docs: a Routers (`orq/*`) section in `configuration.md` covering how a run through one is priced, and why a comparison you intend to reproduce should name a model instead. It uses a live workspace router as the example, since the zero-config family is not live yet.

## Verification

- `ruff check src`, `ruff format --check src`, `basedpyright` (whole repo, 0 errors), `pytest -m 'not integration'` (all passing), `mkdocs build --strict`.
- Live against prod on four routers across both endpoints, including the cross-check against the router's own billing above (`scripts/manual_tests/check_served_model_pricing.py`).
- A router **does** serve the Responses endpoint (`orq/autorouter-openai-cost` → `openai/gpt-5.4`), so `qualified_model` passing a slash-bearing id straight through is correct for this class.

## Scope

Closes RES-1573. This fixes how evaluatorq prices and caches calls through the `orq/*` routers that are live today. Adopting `orq/auto` and `orq/frontier` themselves stays in RES-1528, which waits on RES-1520 shipping on the platform.

## Not in this PR

- Whether any default (`DEFAULT_PIPELINE_MODEL`) should move onto a router, which is a product call, and who owns the intelligence index and the reasoning-effort ladder, which needs the pool-preview endpoint from orq-ai/orquesta-web#11440 and is its own ticket.


Baukebrenninkmeijer (commented):

**3.5/5** — the served-model pricing is sound; one P1 in the fallback, and one question on merge timing.

I traced each `price_usage` call site and the prefix comparison against the code at the head sha. CI is green on all jobs.

### P1

- **An unpriced served model falls back to the router rate** (`src/evaluatorq/common/model_catalogue.py:473`) — when a router serves a model that the catalogue does not list, `price_usage` prices the call at the router headline rate. That is the wrong rate this PR removes, and the output shows it as a measured cost.

### Open Questions

- `docs/configuration.md` and `docs/llm-as-a-jury.md` show `model="orq/auto"` as a working id, but `orq/auto` returns 404 until RES-1520 ships. `docs.yml` publishes the docs on each push to `main`. Do we hold this merge until RES-1520 ships, or keep the `orq/auto` family out of the docs until then?

src/evaluatorq/common/model_catalogue.py
```diff
@@ -428,18 +428,49 @@ async def validate_reasoning_effort(effort: str, model: str, client: AsyncOpenAI
...
+        if info is None:
+            logger.debug(
+                'Model {} served the call but is not in the Orq catalogue; pricing at the requested {} instead',
+                served_model,
+                model,
+            )
+        else:
+            logger.debug('Pricing the call at {}, which served the request for {}', served_model, model)
+    if info is None:
+        info = await _lookup(model, client)
```
Baukebrenninkmeijer: **P1** — The requested id can be an `orq/*` router. If the catalogue does not list the served model, this line prices the call at the router headline rate. That is the wrong rate this PR removes, and the report shows it as a measured cost. For an `orq/*` request, do not fall back to the requested id. Keep the call unpriced.

docs/configuration.md
```diff
@@ -144,6 +144,21 @@ Costs are USD **per 1000 tokens**, matching what `/v2/models` publishes.
 
 The id is stored unprefixed, so `'openai/gpt-x'` and `'gpt-x'` register and resolve the same entry — register either spelling and both lookups find it. Registering both replaces rather than duplicates: there is one model. `reasoning_efforts=None` means "unknown, cannot pre-validate"; an empty set means the same thing and is normalized to `None`, because a literally-empty accepted-values list would reject every effort including the defaults.
 
+## System routers (`orq/auto`, `orq/frontier`)
+
+Orq's model garden carries a few router ids that resolve to a real model per request rather than naming one: `orq/auto` picks by prompt complexity across the models your workspace can call, `orq/auto-open` and `orq/auto-eu` do the same under an open-weight or EU-hosted constraint, and `orq/frontier` and `orq/frontier-cheapest` skip the complexity step and take the strongest or the cheapest top-third model. Pass one anywhere a model id goes:
+
+```python
+evaluator = llm_judge(name="quality", criteria="Is the answer correct?", model="orq/auto")
```
Baukebrenninkmeijer: **Open Question** — `orq/auto` returns 404 until RES-1520 ships, and `docs.yml` publishes this page on merge. Do we hold the merge until RES-1520 ships, or remove the `orq/auto` family from the docs until then?

src/evaluatorq/common/model_catalogue.py
```diff
@@ -428,18 +428,59 @@ async def validate_reasoning_effort(effort: str, model: str, client: AsyncOpenAI
...
+        info = await _lookup(served_model, client)
+        if info is None:
+            logger.debug(
+                'Model {} served the call but is not in the Orq catalogue; pricing at the requested {} instead',
+                served_model,
+                model,
+            )
+        else:
+            logger.debug('Pricing the call at {}, which served the request for {}', served_model, model)
+    if info is None and model.startswith('orq/'):
```
Baukebrenninkmeijer: **P1** — The guard returns before `_lookup(model, client)` runs, so the code does not read `_overrides`. A caller who calls `register_model('orq/autorouter-x', info)` then gets an unpriced call for that router. `register_model` exists to correct an entry the catalogue lists wrongly, and a caller's own rate is not the router headline rate. Read the override for the bare router id before you return.

---

src/evaluatorq/common/model_catalogue.py
```diff
@@ -428,18 +428,59 @@ async def validate_reasoning_effort(effort: str, model: str, client: AsyncOpenAI
...
+        logger.warning(
+            'Router {} was served by {}, which the Orq catalogue does not price; call stays unpriced '
```
Baukebrenninkmeijer: **P2** — This warning says the Orq catalogue does not price the served model. When `served_model` equals the router id, the catalogue does hold that entry. The debug message at line 468 also says the call is priced at the requested id, which this guard prevents. Give one reason for the unpriced call.

---

Baukebrenninkmeijer (commented):

**3.5/5** — the fallback fix is correct; one P1 on `register_model`, and one question on the ticket scope.

I traced the new guard at `model_catalogue.py:474` against `_lookup` and `register_model` at the head sha. CI is green on all jobs.

### Fixed since last review

| # | Item | Resolution |
|---|------|-----------|
| 1 | P1 An unpriced served model falls back to the router rate | ✅ |
| 2 | Open Question The docs name `orq/auto`, which is not live | ✅ |

### P1

- **A `register_model` override for a router id no longer applies** (`src/evaluatorq/common/model_catalogue.py:474`) — the guard returns before `_lookup(model, client)` runs, so the code does not read `_overrides`. A caller who registers a price for `orq/autorouter-x` gets an unpriced call.

### P2

- **The two log messages give different reasons for one call** (`src/evaluatorq/common/model_catalogue.py:476`) — the debug message at line 468 says the call is priced at the requested id. The warning then says the call stays unpriced.

### Open Questions

- RES-1528 names `orq/auto` and `orq/frontier`. These ids are not live yet, so this PR documents the workspace routers instead. Does RES-1528 close when this merges, or does it stay open for the named ids?



---

tests/common/test_model_catalogue.py
```diff
@@ -592,3 +592,78 @@ def test_parse_catalogue_survives_a_non_mapping_metadata():
...
+
+    monkeypatch.setattr(pricing, '_load_catalogue', fake_load)
+    priced = await pricing.price_usage(_usage(), 'orq/autorouter-openai-cost', served_model=served_model)
+    assert priced is not None
+    assert priced.total_cost is None
+    assert priced.priced_calls == 0
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize('served_model', ['anthropic/claude-unlisted', None])
```
Baukebrenninkmeijer: **P1** — Both parameters name a model the fake catalogue does not list, so the test never reaches the case where the served model is listed. In that case `_lookup(served_model)` returns a price and the caller's registered router rate does not apply. Add a third parameter that the fake catalogue lists. Then assert which of the two rates the call takes.

---

src/evaluatorq/common/model_catalogue.py
```diff
@@ -428,18 +428,64 @@ async def validate_reasoning_effort(effort: str, model: str, client: AsyncOpenAI
...
+    the call unpriced and says so, unless the caller has set the router's rate
+    with `register_model`. Compared unprefixed, because
+    the router answers a request for ``openai/gpt-5.6-luna`` with the bare
+    ``gpt-5.6-luna`` and that is the same model, not a reroute.
     """
     if usage is None or usage.total_cost is not None or usage.calls > 1:
         return usage
-    info = await _lookup(model, client)
+    info = None
+    rerouted = served_model is not None and served_model.split('/', 1)[-1] != model.split('/', 1)[-1]
```
Baukebrenninkmeijer: **P2** — `rerouted` is false for two different inputs, and the warning at line 478 reads one reason from it. A `served_model` equal to `model` makes `rerouted` false, and the warning then says the router `did not report which model served it`. The router did report it. An empty `served_model` makes `rerouted` true, and the warning names an empty model, where the old truthy test treated it as absent. Branch the message text on `served_model` itself.

---

src/evaluatorq/common/model_catalogue.py
```diff
@@ -428,18 +428,64 @@ async def validate_reasoning_effort(effort: str, model: str, client: AsyncOpenAI
...
+    if served_model is not None and rerouted:
```
Baukebrenninkmeijer: **P2** — `rerouted` already requires `served_model is not None`, so this condition tests it twice. Write `if rerouted:`.

---

Baukebrenninkmeijer (commented):

**3.5/5** — the override prices the call now; one test gap and two defects in the new warning.

I traced the new branch at `model_catalogue.py:465-480` against `_lookup`, `register_model` and the router tests at the head sha. All checks are green.

### Fixed since last review

| # | Item | Resolution |
|---|------|-----------|
| 1 | P1 A `register_model` override for a router id no longer applies | ✅ |
| 2 | P2 The two log messages give different reasons for one call | ✅ |

### P1

- **No test covers an override against a served model the catalogue lists** (`tests/common/test_model_catalogue.py:644`) — the parameters are `anthropic/claude-unlisted` and `None`, and the catalogue lists neither. The served model wins over the caller's rate in that case. A later change to the branch order moves the price and no test fails.

### P2

- **`rerouted` is false for two different inputs** (`src/evaluatorq/common/model_catalogue.py:465`) — a `served_model` equal to `model` gives false, and the warning then says the router `did not report which model served it`. An empty `served_model` gives true, and the warning names an empty model.
- **The condition tests `served_model` twice** (`src/evaluatorq/common/model_catalogue.py:466`) — `rerouted` already requires `served_model is not None`.

### Open Questions

- RES-1528 names `orq/auto` and `orq/frontier`, and both ids still answer 404. This PR proves the behaviour on the workspace routers instead. Does RES-1528 close when this merges, or does it stay open until RES-1520 lands?



---

src/evaluatorq/common/model_catalogue.py
```diff
@@ -428,18 +428,65 @@ async def validate_reasoning_effort(effort: str, model: str, client: AsyncOpenAI
...
+    is the same model.
     """
     if usage is None or usage.total_cost is not None or usage.calls > 1:
         return usage
-    info = await _lookup(model, client)
+    info = None
+    rerouted_to = (
+        served_model
+        if served_model
+        and not served_model.startswith('orq/')
```
Baukebrenninkmeijer: **P1** — the PR body reports that the router answers `openai/gpt-5.6-luna` with the bare `gpt-5.6-luna`. A served router id can arrive bare for the same reason. `autorouter-anthropic-balanced` passes this guard, `_lookup` finds the router entry at line 466, and the call takes that router's headline rate. That is the defect this PR removes for the prefixed spelling. Decide that the served id names a router from the entry `_lookup` returns, not from the `orq/` prefix.

---

src/evaluatorq/common/model_catalogue.py
```diff
@@ -428,18 +428,65 @@ async def validate_reasoning_effort(effort: str, model: str, client: AsyncOpenAI
...
     Returns ``usage`` unchanged when it is ``None``, already priced, spans more
-    than one call, or the model is absent from the catalogue. The multi-call
-    guard matters: pricing an aggregate at one model's rate would also set
+    than one call, the model is absent from the catalogue, or it is an
+    ``orq/*`` router whose served model cannot be priced. The multi-call guard
+    matters: pricing an aggregate at one model's rate would also set
     ``priced_calls == calls``, asserting full pricing for calls this function
     never saw and defeating `Usage.cost_is_partial`.
+
+    The call is priced at ``served_model``, the id the response came back
```
Baukebrenninkmeijer, line 457: **P2** — this paragraph says the call falls back to the requested id when the served id is not listed. Line 461 also skips an `orq/*` served id, listed or not. Add one sentence that names this case.

---

src/evaluatorq/common/model_catalogue.py
```diff
@@ -428,18 +428,65 @@ async def validate_reasoning_effort(effort: str, model: str, client: AsyncOpenAI
...
+    elif info is None:
+        info = await _lookup(model, client)
+        if info is not None and rerouted_to is not None:
+            logger.debug(
+                'Model {} served the call but is not in the Orq catalogue; pricing at the requested {} instead',
+                rerouted_to,
+                model,
+            )
     if info is None:
         logger.debug('Model {} is not in the Orq catalogue; call stays unpriced', model)
```
Baukebrenninkmeijer: **P2** — when the catalogue lists neither id, this log names `model` only. The router exit at line 478 names the served id. Add `rerouted_to` to this message when it is set.

---

Baukebrenninkmeijer (commented):

**3.5/5** — the three findings from the last review are closed; the new router guard tests only one spelling of a served id.

I traced `price_usage` at `model_catalogue.py:455-492` through each combination of a router and a non-router `served_model`, and I checked the three new tests against the branches they cover. All checks are green at this head sha.

### Fixed since last review

| # | Item | Resolution |
|---|------|-----------|
| 1 | P1 No test covers an override against a served model the catalogue lists | ✅ |
| 2 | P2 `rerouted` is false for two different inputs | ✅ |
| 3 | P2 The condition tests `served_model` twice | ✅ |

### P1

1. **The new guard tests only the prefixed spelling of a served router id** (`src/evaluatorq/common/model_catalogue.py:461`) — the PR body reports that the router answers `openai/gpt-5.6-luna` with the bare `gpt-5.6-luna`. A served router id can arrive bare for the same reason. `autorouter-anthropic-balanced` then passes this guard, `_lookup` finds the router entry, and the call takes a router headline rate.

### P2

- **The docstring does not name the new guard** (`src/evaluatorq/common/model_catalogue.py:447`) — it says the call falls back to the requested id when the served id is not listed. An `orq/*` served id is skipped even when the catalogue lists it.
- **The last unpriced exit drops the served id** (`src/evaluatorq/common/model_catalogue.py:491`) — when the catalogue lists neither id, the log names `model` only. The router exit above it names the served id.



---

Baukebrenninkmeijer (approved):
All three review rounds are closed and verified at `ce18193f`.

Round 3 items: the register_model override now has a test against a *listed* served model (`test_listed_served_model_outranks_a_registered_router_rate`), `rerouted_to` replaces the ambiguous `rerouted` flag so `''` and `None` both report "did not report which model served it", and the duplicated `served_model` check is gone. The extra defect found while fixing those — a served id that is itself `orq/*` being priced at that router's headline rate — is guarded and tested.

Ran `tests/common/test_model_catalogue.py tests/test_reuse_guardrails.py tests/unit/test_prompt_cache.py` locally at this sha: 109 passed. All 15 CI checks green, no conflicts.

The ticket split is the right call: RES-1573 for the pricing and caching fixes, RES-1528 left open for adopting `orq/auto` / `orq/frontier` once they resolve. Both still 404 as of today, checked live.

Two non-blocking notes for later, not for this PR:
- `scripts/refresh_jury_judge_rates.py` still says RES-1528 will replace the local capture with the system routers, which reads against `jury_presets.py` saying seats stay pinned locally.
- The `jury_presets.py` rationale for excluding routers cites vendor correlation, but the routers that exist today are single-vendor (`autorouter-anthropic-*`, `autorouter-google-*`). The durable reason is unpriceability and reproducibility — the catalogue lists the same headline rate for a vendor's cost and quality tiers, so it carries no information.

PR #196, opened 2026-09-08
feat(simulation): custom generation_instructions to steer persona/scenario generation (RES-1087)

KarinaKKarinaK:
## What

Adds a `generation_instructions` parameter to agent-simulation generation: a free-text steer applied to every generated persona AND scenario. Before this, only `seed` (a per-archetype string) and `edge_case_percentage` shaped generation, so there was no way to say "make all of these enterprise B2B buyers replying in German" across the whole batch.

Closes RES-1087.

## Surface

- SDK: `generate_and_simulate()`, `generate()`, `generate_personas()`, `generate_scenarios()`, `generate_persona()`, `generate_scenario()` all take `generation_instructions: str = ''`.
- Generators: `PersonaGenerator.generate()` / `ScenarioGenerator.generate()` inject it into the user prompt.
- CLI: `--generation-instructions` on both `sim run` and `sim generate`.
- Empty string (the default) leaves the built-in prompts byte-for-byte unchanged.

## Design choices

- Mirrors the existing `persona_seeds` / `scenario_seeds` threading exactly, so it is NOT a `SimulationConfig` field. It rides the same generate-only chain those seeds already use.
- Treated as a trusted caller instruction, like `seed`: it is meant to STEER generation, so it is passed through rather than wrapped in `delimit()` (which frames text as inert data and would defeat an instruction we want followed). The line is instruction-vs-data, not the channel: `agent_description` and `context` arrive through the same caller but are data the personas/scenarios describe, so they stay delimited. The output contract does not rely on prompt ordering: `generate_structured` enforces the `PersonaListResponse` / `ScenarioListResponse` schema via `response_format` on the Responses path, so the JSON shape holds regardless of where the steer sits. Only the `json_object` fallback would depend on prompt wording, so this pass-through must not be copied to a prompt-only, unvalidated call site.
- One knob for both dimensions. The CLI exposes a single `--generation-instructions` shared by personas and scenarios; to steer them differently, use the API and call `generate_personas()` / `generate_scenarios()` separately (this split is not reachable from the CLI flag).
- It composes with seeds and `edge_case_percentage` rather than replacing them.

## Tests

`tests/simulation/test_seeded_generation.py`, `test_generation_injection.py`, `test_cli.py`: the steer reaches both prompts on the auto path, composes with a seed, reaches the `_seeded` fan-out branch (per-seed prompts carry the steer), threads through `generate()` and `generate_and_simulate()`, is forwarded by both CLI commands (`generate` and `run`), and the empty default injects nothing.

## Verified

`ruff check src`, `ruff format --check src`, `basedpyright` (whole repo), and `pytest -m 'not integration'` (5438 passed) all green. Note: this checkout has untracked `* 2.py` duplicate files from local sync that trip the import-isolation and cache-control guardrail tests; those pass in a clean worktree, confirmed against this commit.

## Docs

Added `--generation-instructions` to the CLI reference tables and a note in the agent-simulation guide.

### DIFF
--- a/src/evaluatorq/simulation/api.py
+++ b/src/evaluatorq/simulation/api.py
@@ -624,6 +624,7 @@ async def generate_and_simulate(
     edge_case_percentage: float | None = None,
     persona_seeds: list[str] | None = None,
     scenario_seeds: list[str] | None = None,
+    generation_instructions: str = '',
     max_turns: int | None = None,
     llm_config: LLMCallConfig | None = None,
     evaluator_names: list[str] | None = None,
@@ -699,6 +700,14 @@ async def generate_and_simulate(
     for it. The other dimension still auto-generates; seeded x auto still
     crosses into the full grid. See `generate` for the full semantics.
 
+    ``generation_instructions``: Free-text steer applied to every generated
+    persona AND scenario (e.g. ``"enterprise B2B buyers, frustrated, replying
+    in German"``). Unlike a seed — which names one archetype and fans out to one
+    object — this stacks on top of the whole batch and composes with seeds and
+    ``edge_case_percentage``. Empty string (the default) leaves the built-in
+    prompts unchanged. To steer personas and scenarios differently, call
+    `generate_personas` / `generate_scenarios` separately.
+
     ``evaluation_description``: Optional human-readable note passed straight
     through to the uploaded experiment (shown as its description in the Orq UI).
     Pure metadata — nothing branches on it, and it only matters when
@@ -771,6 +780,7 @@ async def generate_and_simulate(
         edge_case_percentage=edge_case_percentage,
         persona_seeds=persona_seeds,
         scenario_seeds=scenario_seeds,
+        generation_instructions=generation_instructions,
         max_turns=max_turns,
         llm_config=llm_config,
         evaluator_names=evaluator_names,
@@ -811,6 +821,7 @@ async def _generate_datapoints_inner(
     persona_seeds: list[str] | None = None,
     scenario_seeds: list[str] | None = None,
     edge_case_percentage: float | None = None,
+    generation_instructions: str = '',
 ) -> tuple[list[SimulationDatapoint], AsyncOpenAI, bool, TokenUsage | None]:
     """Shared generation trio: personas/scenarios + first-message datapoints.
 
@@ -849,6 +860,7 @@ async def _generate_datapoints_inner(
             persona_seeds=persona_seeds,
             scenario_seeds=scenario_seeds,
             edge_case_percentage=edge_case_percentage,
+            generation_instructions=generation_instructions,
         )
         await await_maybe(gen_hooks.on_generate_inputs_ready(len(gen_personas), len(gen_scenarios)))
 
@@ -882,6 +894,7 @@ async def _generate_and_simulate_run(
     edge_case_percentage: float | None = None,
     persona_seeds: list[str] | None = None,
     scenario_seeds: list[str] | None = None,
+    generation_instructions: str = '',
     max_turns: int | None = None,
     llm_config: LLMCallConfig | None = None,
     evaluator_names: list[str] | None = None,
@@ -989,6 +1002,7 @@ async def _generate_and_simulate_run(
                                 persona_seeds=persona_seeds,
                                 scenario_seeds=scenario_seeds,
                                 edge_case_percentage=edge_case_percentage,
+                                generation_instructions=generation_instructions,
                             )
                             if emit_datapoints is not None:
                                 emit_datapoints(datapoints)
@@ -1065,6 +1079,7 @@ async def generate(
     generation_client: AsyncOpenAI | None = None,
     persona_seeds: list[str] | None = None,
     scenario_seeds: list[str] | None = None,
+    generation_instructions: str = '',
     edge_case_percentage: float | None = None,
     llm_parallelism: int | None = None,
 ) -> list[SimulationDatapoint]:
@@ -1080,6 +1095,11 @@ async def generate(
     object, overriding ``num_personas`` / ``num_scenarios`` for that dimension.
     The other dimension still auto-generates. Seeded x auto still crosses into
     the full persona x scenario grid.
+
+    ``generation_instructions`` is a free-text steer applied to every generated
+    persona and scenario (e.g. ``"enterprise B2B buyers, replying in German"``),
+    stacked on top of the auto/seed prompt. It composes with seeds and
+    ``edge_case_percentage``; empty string (the default) changes nothing.
 
     ``edge_case_percentage`` overrides the fraction of the auto-generated
     portion of each dimension that are edge cases (default 0.2 for personas,
@@ -1150,6 +1170,7 @@ async def generate(
                         persona_seeds=persona_seeds,
                         scenario_seeds=scenario_seeds,
                         edge_case_percentage=edge_case_percentage,
+                        generation_instructions=generation_instructions,
                     )
                 finally:
                     # Thread the in-flight exception into the stage-end meta, as
@@ -1177,6 +1198,7 @@ async def generate_personas(
     *,
     agent_description: str = '',
     context: str = '',
+    generation_instructions: str = '',
     llm_config: LLMCallConfig | None = None,
     generation_client: AsyncOpenAI | None = None,
 ) -> list[Persona]:
@@ -1187,6 +1209,9 @@ async def generate_personas(
     each archetype, the LLM fills every trait. Provider resolves via the shared
     factory: an injected ``generation_client`` → ``llm_config.client`` →
     ``ORQ_API_KEY`` → ``OPENAI_API_KEY``.
+
+    ``generation_instructions`` is an optional free-text steer applied to every
+    persona on top of its seed (e.g. ``"all replying in German"``).
 
     ``llm_config`` carries the model and everything around it: only the fields you set take effect,
     so an unset ``temperature`` still omits the parameter from the request.
@@ -1213,6 +1238,7 @@ async def generate_personas(
                 num_personas=1,
                 edge_case_percentage=0.0,
                 seed=seed,
+                generation_instructions=generation_instructions,
             )
             for seed in seeds
         ])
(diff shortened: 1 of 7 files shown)


KarinaKKarinaK:
Dogfooded live against the real generators (ORQ router), not just mocks:

`generate(agent_description="business banking support agent", num_personas=3, num_scenarios=1, generation_instructions="Every persona is a German small-business owner and writes ONLY in German")`

All 3 personas honored the steer end to end:
- Pragmatische Handwerksmeisterin: "Katrin führt einen kleinen Sanitärbetrieb in Leipzig..."
- Digitaler Softwareunternehmer: "Jonas betreibt als Einzelunternehmer ein kleines Softwarestudio in Hamburg..."
- Freundliche Caféinhaberin: "Monika führt in Nürnberg ein kleines Café..."

3/3 backgrounds in German, all German small-business owners. Confirms the instruction reaches the prompt and the model acts on it for the whole batch.

src/evaluatorq/simulation/api.py
```diff
@@ -1433,6 +1478,9 @@ async def _seeded(gen: Any, seeds: list[str], kind: str, kwarg: str) -> list[Any
         if edge_case_percentage is not None:
             persona_kwargs['edge_case_percentage'] = edge_case_percentage
             scenario_kwargs['edge_case_percentage'] = edge_case_percentage
+        if generation_instructions:
```
Baukebrenninkmeijer: **P2** — `_seeded` passes `generation_instructions` always. This branch passes it only when the string is not empty. An empty string adds nothing to the prompt in either case, so the two branches agree. Remove the `if generation_instructions:` guard.

tests/simulation/test_seeded_generation.py
```diff
@@ -160,3 +160,47 @@ async def test_generate_personas_scenarios_default_edge_case_percentage_unchange
...
+    assert "angry customer" in prompt
+    assert "reply only in French" in prompt
+
+
+@pytest.mark.asyncio
+async def test_no_generation_instructions_injects_nothing(captured):
+    """The default empty string must not add the caller-instructions block, so an
+    ordinary run's prompt is byte-for-byte unchanged from before the feature."""
+    await generate_persona("angry customer", agent_description="support agent")
+    assert "instructions from the caller" not in captured["prompts"][0]
```
Baukebrenninkmeijer, line 206: **P2** — These three tests cover `_generate_personas_scenarios` and `generate_persona`. Two paths stay untested.

The `_seeded` branch at `src/evaluatorq/simulation/api.py:1464` calls `gen.generate(...)` where `gen` is typed `Any`, so `basedpyright` does not check the keyword name there. `test_generation_instructions_compose_with_seed` uses `generate_persona`, which is a different function.

The `--generation-instructions` flag has no test. `--persona-seed` has one in `tests/simulation/test_cli.py:1589`, and `generate_and_simulate` has one in `tests/simulation/test_generation_injection.py:473`.

A broken keyword on either path produces personas and scenarios that no instruction steered. The run reports success.

Baukebrenninkmeijer (commented):

**4/5** — clean threading, accurate docs, two nits.

I traced `generation_instructions` from each public entry point down to `PersonaGenerator.generate` and `ScenarioGenerator.generate`. It arrives on the auto path, the seeded path and the CLI path. Every check passes.

### P2

- `persona_seeds` has a CLI test and an entry-point test. `generation_instructions` has neither. A dropped keyword gives unsteered output and no error.
- `_generate_personas_scenarios` sets the keyword always in one branch and conditionally in the other (`src/evaluatorq/simulation/api.py:1481`). Both branches give the same prompt.

currentlycodinng (commented):
## Verdict: COMMENT — ship-worthy, one wording fix requested

**Rating: 4/5** — solid implementation, but the security rationale in the docstrings should be corrected before it gets copied elsewhere

### Verified against base `e94b995` / head `6fdf892`

**Suite:** ruff check/format, basedpyright, `pytest -m 'not integration'` — 5422 passed, 4 skipped, 10 deselected, all clean.

**Empty-string equivalence:** confirmed byte-identical for all 4 shapes (persona/scenario × auto/seeded) with a mocked `generate_structured` comparing base vs head. Caught myself on a false pass first (PYTHONPATH resolving to head's module against a stale base worktree), re-ran with the module path explicitly checked.

### 🟡 Fix requested: the security framing, not the security
The "trusted channel, no privilege boundary" claim is fine — `seed` parity confirmed at `api.py:1481`, `persona_generator.py:194`, `scenario_generator.py:268`, all interpolated raw, no `delimit()`.

But the stated justification is the weak part: the docstrings lean on "the JSON-only instruction still comes after it in the prompt," which is recency priming, not a guarantee. What actually holds the line is that `generate_structured` runs with `response_format=PersonaListResponse` over the Responses API — rungs 1–2 of the structured-output ladder enforce the schema at the API level regardless of where the steer sits. Only rung 4 (`json_object`) degrades to prompt-only.

**Ask:** reword the docstrings and PR description to cite schema enforcement, not prompt ordering — the current wording would license a genuinely unsafe pattern if copied to a prompt-only call site. Also worth reconciling that `agent_description`/`context` arrive through the same caller channel yet get `delimit()`-ed anyway, so "same channel, no boundary" isn't the real distinction being drawn.

### P2s
- **Test coverage gap — confirmed, slightly wider than reported.** `generate()`, `generate_and_simulate()`, the `_seeded` fan-out branch, and both CLI commands have no coverage for this param, unlike the existing `persona_seeds` analogue at `test_generation_injection.py:472`. Wrote throwaway tests for all four paths, all passed clean — coverage gap only, not a hidden bug.
- **`api.py:1481` branch inconsistency — cosmetic, confirmed.** Both branches behave identically for every input including whitespace-only and None, since the shared default is `''` and the guard is plain truthiness. Unlike `edge_case_percentage`, whose divergent defaults actually require `is not None`.

No new bugs. Threading verified live through `generate`, `generate_and_simulate`, `cli._generate_impl`, `cli._run_impl`.

### 🔴 Needs live LLM testing / product call
- Whether a real model follows a steer across a whole batch without flattening persona/scenario diversity — unit tests can't answer this.
- Rung-4 behavior under a hostile steer when a provider falls through to bare `json_object`.
- **Product call:** the CLI exposes one shared steer for both personas and scenarios with no way to split them — the docstring's suggested workaround isn't reachable from the CLI flag.

KarinaKKarinaK:
Pushed bb0194ae. All points addressed.

**Security framing (cc's fix requested).** You're right, prompt ordering was the wrong justification. The docstrings and PR description now say the output contract is held by `generate_structured` enforcing the `PersonaListResponse` / `ScenarioListResponse` schema via `response_format` on the Responses path, so the JSON shape holds regardless of where the steer sits; only the `json_object` fallback depends on prompt wording, and I added an explicit note not to copy the pass-through to a prompt-only call site. I also fixed the delimit distinction: it is instruction-vs-data, not same-channel. `agent_description` and `context` come through the same caller but are data the personas/scenarios describe, so they stay `delimit()`-ed, while the steer (like `seed`) is an instruction we want followed.

**Test coverage (both reviewers).** Added the four paths that had none: `generate()` and `generate_and_simulate()` (assert the kwarg reaches the shared `_generate_datapoints_inner` seam), the `_seeded` fan-out branch (per-seed persona prompts carry both seed and steer), and both CLI commands (`--generation-instructions` reaches `_generate_impl` and `_run_impl`).

**api.py:1481 branch (both reviewers).** `generation_instructions` is now set unconditionally in the persona/scenario kwargs, matching the seeded fan-out and `agent_description`. Its `''` default is a no-op in the generators, so the truthiness guard was redundant and asymmetric with the seeded branch. `edge_case_percentage` stays conditional because its default genuinely diverges.

**Product call - one shared CLI flag (cc).** Kept as one flag but made the docs honest: the PR description now states the CLI exposes a single `--generation-instructions` for both dimensions, and splitting persona vs scenario steer requires the API (`generate_personas()` / `generate_scenarios()`), not reachable from the CLI. If you want two CLI flags, say so and I'll add them.

**Live LLM items (cc).** Whether a real model follows a steer across a batch without flattening diversity, and rung-4 behavior under a hostile steer, are genuinely live questions unit tests can't answer; happy to run a live batch through `generate` if you want that before merge, but they aren't code fixes.

Local: ruff check src clean, ruff format --check src clean, basedpyright 0 errors, the three affected test files pass (144).

Baukebrenninkmeijer (commented):

**4 → 5/5** — both nits from the last review are fixed. Nothing left to change.

**This is an approval.** The session that runs this routine cannot submit a formal GitHub approval today, so the review is filed as a comment. Read the rating, not the review state.

I re-read `bb0194ae` and checked the two claims the new docstrings make. `generate_structured` takes `response_format` and enforces the schema on the Responses leg; the bare `json_object` rung is the last one and is the only rung that depends on prompt wording. `agent_description` and `context` go through `delimit()` in both generators, and `generation_instructions` does not. Both statements are correct. `ScenarioGenerator.generate` and `PersonaGenerator.generate` both declare `generation_instructions: str = ''` and both guard with `if generation_instructions:`, so the unconditional pass in `_generate_personas_scenarios` is a true no-op at the default. I traced every entry point: `generate`, `generate_and_simulate`, `sim generate` and `sim run` all declare the parameter as `str`, so `None` cannot arrive. CI is green: 14 checks pass and `Deploy GitHub Pages` is skipped.

### Fixed since last review

| # | Item | Resolution |
|---|------|-----------|
| 1 | P2 `generation_instructions` has no CLI test and no entry-point test | ✅ |
| 2 | P2 `_generate_personas_scenarios` sets the keyword always in one branch and conditionally in the other | ✅ |



---

KarinaKKarinaK:
Both P2 comments are addressed in bb0194ae:
- the `if generation_instructions:` guard in `_generate_personas_scenarios` is gone; it now passes unconditionally, matching `_seeded` (empty string is a no-op either way).
- `test_generation_instructions_reach_the_seeded_fanout` now covers the seeded fan-out path with `generation_instructions`, so a kwarg typo there would be caught.

Threads resolved, CI green. @Baukebrenninkmeijer could you re-review and approve when you have a moment so this can merge.

PR #197, opened 2026-09-08
feat(simulation): extend_from_dataset - generate more datapoints seeded from an Orq dataset (RES-1084)

KarinaKKarinaK:
## What

Adds `extend_from_dataset()` — seed generation from an existing Orq dataset to produce *more* datapoints matching its distribution. Before this, dataset input was replay-only (`simulate(dataset_id=...)`); the only way to generate fresh cases was scratch generation from a description. This fills the Dataset/extension cell of the coverage matrix in `simulation/README.md`.

Closes RES-1084.

## Surface

- `extend_from_dataset(dataset_id, *, num_personas=3, num_scenarios=5, llm_config=None, agent_description=None, api_key=None)` — feeds the dataset's personas/scenarios to the generators as seeds, returns new similar-but-not-duplicate datapoints. Mirrors `extend_from_experiment()`.
- `datapoints_from_dataset(dataset_id, *, api_key=None)` — the public direct loader, added for parity with `datapoints_from_experiment` / `datapoints_from_traces`.
- Both exported from `evaluatorq.simulation`. SDK-only, matching `extend_from_experiment` (no CLI; only trace extension has one).

## Design

- **Consolidated, not copied.** The extension body (fetch seeds, feed them to `DatapointGenerator`) is identical between experiment and dataset extension. The shared half now lives in `simulation/_seed_extension.py` (`extend_from_seeds`, `seed_context`, `describe_agent`), and `extend_from_experiment` was refactored to call it. Its existing tests guard the refactor.
- **One dataset loader.** `datapoints_from_dataset` is now the canonical dataset-to-datapoints loader; `api.py`'s private `_fetch_simulation_datapoints_from_orq` (the `simulate(dataset_id=...)` path) delegates to it, so there's no duplicate fetch and the direct path is unchanged in behavior.

## Tests

New `tests/simulation/test_dataset_input.py` mirrors the experiment tests: direct extraction, missing-key error, bad-row index message, zero-rows error, the `simulate(dataset_id=...)` resolver still working through the delegation, and extension seeding the generators with a deduped seed context.

## Verified

In a clean worktree (CI-equivalent): `ruff check src`, `ruff format --check src`, `basedpyright` (whole repo), and `pytest -m 'not integration'` (5441 passed) all green. Note: this checkout has untracked `* 2.py` duplicate files from local sync that trip the import-isolation guardrail tests; those pass clean, confirmed against this commit.

## Docs

Datasets section in `simulation/README.md` now documents Direct + Extension (mirroring Experiments), the coverage-matrix cell flips to `extend_from_dataset()`, and the agent-simulation guide gets an extension example.


KarinaKKarinaK:
Dogfooded the extension live against the real `DatapointGenerator` (ORQ router), not just mocks. Seeded `extend_from_seeds` with two banking-support datapoints (a locked-out retiree, a debit-card dispute) and asked for 2x2:

Result: 4 datapoints, all fresh (no seed copies), all in the same banking/security domain:
- Freelancer Locked Out After Changing Phones :: Urgent Unauthorized ACH Withdrawals
- Freelancer Locked Out After Changing Phones :: Traveling Customer Needs Secure Phone Update
- Traveling Professional Investigating a Security Hold :: Urgent Unauthorized ACH Withdrawals
- Traveling Professional Investigating a Security Hold :: Traveling Customer Needs Secure Phone Update

Extends coverage in-distribution rather than duplicating the seeds, which is the whole point of the feature.

src/evaluatorq/simulation/datasets.py
```diff
@@ -0,0 +1,105 @@
...
+
+import os
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from evaluatorq.contracts import LLMCallConfig
+    from evaluatorq.simulation.types import SimulationDatapoint
+
+
+def _require_api_key(api_key: str | None) -> str:
```
Baukebrenninkmeijer: **P2** — `experiments.py:26` holds the same helper. Only the message differs: `Set it to load Orq datasets.` against `Set it to load Orq experiments.` The PR body calls this design "Consolidated, not copied", and this is the one place it does not hold. Move `_require_api_key` to `_seed_extension.py` and pass the noun as an argument.

Baukebrenninkmeijer (changes requested):

**2.5/5** — the code is sound. A required check is red.

I checked both refactors against the previous behaviour. `extend_from_experiment` keeps its error paths after the move to `_seed_extension.extend_from_seeds`. `_fetch_simulation_datapoints_from_orq` keeps its zero-row and bad-row errors after the delegation to `datapoints_from_dataset`. Every other check passes.

### P0

1. **The required check `Conventional title` fails.** The PR title and the single commit subject do not match. The title uses `-`. The commit subject uses `—`. The job reports: `The pull request title doesn't match the commit though ("feat(simulation): extend_from_dataset - generate more datapoints seeded from an Orq dataset (RES-1084)" vs. "feat(simulation): extend_from_dataset — generate more datapoints seeded from an Orq dataset (RES-1084)")`. The `Protect-main` ruleset lists `Conventional title` as a required status check, so the merge stays blocked. Edit the PR title to use `—`. No code change is needed.

### P2

- `datasets.py` copies `_require_api_key` from `experiments.py`. Only the last word of the message differs.

currentlycodinng (commented):
## Verdict: COMMENT — clean refactor, one mechanical CI blocker

**Rating: 4.5/5** (would be 5/5, the automated review's 2.5/5 was driven entirely by the CI check below, not code quality)

### What actually blocks merge
Required **Conventional title** check is red: PR title uses a hyphen, the single commit's subject uses an em-dash, and this ruleset requires them to match. No code change needed, just edit the title to use the em-dash.

### Verification
Didn't take the automated review's "code is sound" at face value. Built a differential harness: faked `fetch_dataset_batches`/`fetch_experiment_datapoints`, ran zero-row, empty-batch-only, bad-row at every index (including across a batch boundary), good rows, and missing-key cases against base and this branch side by side, comparing exception type and exact message.

- **18/20 identical byte for byte**, including row indices surviving across batch boundaries.
- 2 diffs are in `_fetch_simulation_datapoints_from_orq`'s new env-var fallback for an empty/None api_key — unreachable in practice, the sole call site resolves the key via `_require_orq_api_key` first. Not a real behavior change.
- Fed identical seeds through `extend_from_experiment` and `extend_from_dataset`, compared captured generator kwargs: byte-identical. No dataset-specific field gets dropped by the generalized shared code.

### Pre-existing, not new
`seed_context` dedupes scenarios by name with last-wins — two same-named scenarios with different goals can collapse and a blank goal overwrites a real one. Worth a follow-up ticket, not this PR's bug.

### Downgrading a nit
The duplicate-`_require_api_key`-helper P2 reads more like P3 to me: the identical `api_key or os.environ.get('ORQ_API_KEY')` pattern already appears at 20 sites including `traces.py:354`. Consistent with the codebase, not new duplication worth blocking on.

### Suite
ruff check / format, basedpyright all clean. `pytest -m 'not integration'`: 5425 passed, 4 skipped, 10 deselected. The `* 2.py` guardrail note in the PR body is a local-sync artifact, confirmed absent in a clean checkout.

### 🔴 Needs live testing (not resolvable here)
No test fetches a real Orq dataset — pagination and row shapes are only exercised by mocks. `extend_from_dataset`'s actual generation quality from dataset-seeded personas/scenarios has never run against a live LLM.

KarinaKKarinaK:
Pushed 1b342d30. Both points addressed.

**P0 - Conventional title check red.** The single commit's subject used an em dash while the PR title used a hyphen, and the check requires them to match for a single-commit PR. Rather than change the PR title to an em dash (we keep those out of titles), I amended the commit subject to the hyphen so it matches the PR title exactly. The check should go green.

**P2 - duplicated _require_api_key.** Consolidated. experiments.py and datasets.py each had a private _require_api_key differing only by one word of the message; both now call _seed_extension.require_orq_api_key(api_key, purpose=...), the module both already depend on for extend_from_seeds, so no new coupling. Messages preserved via the purpose arg ('load Orq datasets' / 'load Orq experiments').

Local: ruff check + format clean, basedpyright 0 errors, the two affected test files pass (20). No behavior change; the differential-harness paths cc verified are untouched.

On the 🔴 live-testing items (no test fetches a real Orq dataset; generation quality from dataset-seeded personas/scenarios hasn't run against a live LLM): those are genuine live checks, not code fixes. Happy to run a live dataset extension if you want it before merge.

Baukebrenninkmeijer (commented):

**2.5 → 4.5/5** — the red check is green and the copied helper is gone. One nit inline.

**This review is meant to clear my own change-request, and it cannot.** The session that runs this routine is not permitted to submit a formal GitHub approval today, so this is filed as a comment. A comment does not lift a `CHANGES_REQUESTED`, so **the merge stays blocked until Bauke dismisses the old review by hand**. Nothing in the code is holding it.

The branch was force-pushed and squashed, so `f59dbd61` is no longer an ancestor of `1b342d30`. I re-read the whole PR at head instead of an increment. Both blockers are resolved:

| # | Item | Resolution |
|---|------|-----------|
| 1 | P0 The required check `Conventional title` fails | ✅ |
| 2 | P2 `datasets.py` copies `_require_api_key` from `experiments.py` | ✅ |

The title and the commit subject now agree, and `Conventional title` passes. `_require_api_key` is gone from both modules; `require_orq_api_key` in `_seed_extension.py` is the single definition. I compared the old `extend_from_experiment` body against `extend_from_seeds`: `sim_llm_config`, the `DatapointGenerator` construction, the `generate_from_description` arguments and the `finally: close()` are unchanged. `simulate(dataset_id=...)` resolves the key through `_require_orq_api_key` before it calls `_fetch_simulation_datapoints_from_orq`, so the new guard inside `datapoints_from_dataset` is redundant and changes nothing. CI is green: 14 checks pass and `Deploy GitHub Pages` is skipped.

### P2

- **`describe_agent` takes a fallback without a log line** (`src/evaluatorq/simulation/_seed_extension.py:77`) — see the inline comment.



---

src/evaluatorq/simulation/_seed_extension.py
```diff
@@ -0,0 +1,103 @@
...
+def describe_agent(seeds: list[SimulationDatapoint]) -> str:
+    """Fallback agent description built from the seed scenarios' goals.
+
+    Seeds are guaranteed non-empty but individual goals are not, so when every goal is blank fall
+    back to a generic description instead of handing the generator a truncated 'goals such as: '
+    prompt.
+    """
+    goals = list(dict.fromkeys(dp.scenario.goal for dp in seeds if dp.scenario.goal))
+    if not goals:
+        return 'A general-purpose assistant agent; extend the seed personas and scenarios below.'
```
Baukebrenninkmeijer, line 84: **P2** — every seed goal is blank, so the generator gets a generic description instead of one built from the caller's data. Nothing records that. `CLAUDE.md` says a degraded path announces itself. Add a `logger.warning` here.

---

KarinaKKarinaK:
Addressed Bauke's review (commit 23d1d112).

The blank-goal fallback in describe_agent (_seed_extension.py) now announces itself: when every seed scenario goal is blank it emits a logger.warning naming the cause (count of seeds, and that generation will be less targeted) before returning the generic description. Matches the house rule that a degraded path logs.

Added the test: test_describe_agent_falls_back_when_all_goals_blank now asserts the warning fires (captured via a loguru sink) in addition to the non-degenerate description.

Local checks pass: ruff check src, ruff format --check src, basedpyright (whole repo) 0 errors, and the targeted test passes.

Baukebrenninkmeijer (approved):

**4.5 → 5/5** — the blank-goal fallback announces itself, and the test fails if the warning goes away.

Re-reviewed `1b342d30...23d1d112`. The warning text contains the literal `blank goal`, and the test asserts on that substring, so the assertion follows real behaviour. The sink goes away in the `finally` block, so it does not leak into the other tests in the process. `logger.warning('... {} ...', len(seeds))` is loguru brace style and renders correctly. CI is green: 14 checks pass and `Deploy GitHub Pages` is skipped.

### Fixed since last review

| # | Item | Resolution |
|---|------|-----------|
| 1 | P2 `describe_agent` takes a fallback without a log line | ✅ |

I approve to clear my own change-request from 8 September. No code holds this PR.




---

PR #200, opened 2026-09-14
docs: explain what a misconfigured env var does [axes]

Baukebrenninkmeijer:
## The gap

`gap: tier 1 env var contract (no second axis)`

Every environment variable is listed on `configuration.md` with its default and its meaning. What happens when you get one **wrong** is documented nowhere on the site. The shared reader added by #188 (`common/env_config.py`, RES-1297) parses and validates ten knobs and, on a value it cannot use, logs a `WARNING` and falls back to the default instead of raising — a user-visible change, since two of those knobs previously raised `ValueError`. That contract reached `CHANGELOG.md` and stopped there.

It ranked first because it is a first-hour usage path reachable from every surface — "I set the variable and nothing changed" has no answer on the site — and because it is demonstrable without a single live API call. The runner-up was `evaluatorq.__version__`, which is one sentence and whose value is already documented as `eq --version`.

## What the page covers

A new `## When a value is wrong` section on `configuration.md`:

- Which ten knobs are validated, what the rest do instead, and the two that behave differently (`EVALUATORQ_SPAN_MAX_TEXT_CHARS` parses on its own terms; `EVALUATORQ_LOG_LEVEL` raises and kills the dashboard — see Code defect).
- Each minimum, stated. The table's "below the knob's minimum" row was unactionable without them.
- A table of what each class of bad value does, including the `ORQ_OTEL_MAX_BATCH_SIZE` clamp, which is the one outcome that lands on a value you never set.
- Why the absence of a warning proves nothing: a variable that did nothing was rejected, outranked, or never read, and only the first says so.
- The CI failure mode, with a preflight assertion that catches it.
- Two runnable demonstrations, plus their real captured output.

Two consistency fixes on the same page follow from it: the `ORQ_DISABLE_TRACING` rows and the two sibling boolean rows named a narrower vocabulary than the code accepts, and now link to the new section.

**Deliberately not covered.** The precedence rules for `EVALUATORQ_LLM_MAX_TOKENS` and `EVALUATORQ_LLM_TIMEOUT_S` are a second topic on a different section of the page, and `tracing.md` and `faq.md` still describe `ORQ_DISABLE_TRACING`'s vocabulary narrowly. Both are written up under Open decisions rather than fixed here, because the routine's scope is one section.

## `[axes]` — what changed in the coverage skill, and why you have to look

The title carries the `[axes]` tag because this branch edits `.claude/skills/docs-coverage/axes.md`, the file that decides what counts as a docs gap at all. An agent quietly shrinking its own future workload is the one edit here a person is required to see, so here is the whole of it.

**Four `N/A` rows added.** The coverage sweep resolved four cells as structurally impossible for reasons the file did not list, so they were re-argued from scratch this run and would be again next week: `evaluatorq()` crossed with the HuggingFace, generated and replay data sources, and `simulate()` / `generate_and_simulate()` crossed with HuggingFace. In each case the parameter simply does not exist — `data=` accepts rows, a `DatasetIdInput` or an `ExperimentInput` and nothing else, and replay lives on `red_team()` and `simulate()` only.

**One stale gap claim retired.** The `API endpoint` note said red team's attacker pins `api='chat_completions'`, so a caller's `api` is inert, and told future runs to "treat that as a gap to document, not an `N/A`, until the code either honours it or warns." The code now warns: `LLMCallConfig.request_params` logs when an explicit `api=` contradicts one the caller set, and both attacker call sites render through it. The note's own condition has been met, so it no longer describes a gap. This is the edit worth a second look, because it is the one that removes something from the matrix rather than adding to it; the replacement text says what would re-open it.

**Not changed:** no axis was added, removed or renamed, and no tier rule was touched.

## Receipt

Final round, generated by the tracked runner `.claude/skills/docs-autofill/run_receipt.sh` against a fresh `EVALUATORQ_DIR`. Scope is the blocks this branch added; the page's pre-existing blocks were not re-run.

| Block | Command | Exit | Output (first lines) |
|---|---|---|---|
| 001 (`bash`, :150) | the CI preflight loop | `0` | *(no output — nothing set-but-empty in the runner's shell)* |
| 002 (`bash`, :165) | `EVALUATORQ_CATALOGUE_TIMEOUT_S=abc eq redteam runs` | `0` | `WARNING ... EVALUATORQ_CATALOGUE_TIMEOUT_S is not a number ('abc'); using default 30.0.` / `No runs found (directory /tmp/tmp.HeVvr21JJ1/runs does not exist).` |
| 003 (`console`, :169) | — | — illustrative (expected-output transcript) | — |
| 004 (`bash`, :175) | `ORQ_DISABLE_TRACING=maybe python -c "..."` | `0` | `WARNING ... ORQ_DISABLE_TRACING is not a boolean ('maybe'); using default False.` |
| 005 (`console`, :179) | — | — illustrative (expected-output transcript) | — |

`RESULT: OK — 5 added blocks accounted for, every executable one exited 0`

Block 001 exits `0` in the runner because nothing there is set-but-empty, which proves only half of an assertion. It was checked separately in both directions: a clean shell exits `0` with no output, and with two variables set to the empty string it prints two `::error::` lines and exits `1`.

`uv run --group docs mkdocs build --strict` and `uv run python scripts/validate_mermaid.py` both pass. Nothing under `src/` was touched.

## Persona verdicts

|---|---|---|---|
| First-timer | Decide, before adopting evaluatorq, whether a typo in a config variable crashes a run, silently misbehaves, or reports itself | COMPLETED | 1 |
| Platform user | Find out which values actually switch `ORQ_DISABLE_TRACING` off, after a colleague wrote `Yes` in a shared env file | COMPLETED | 1 |
| Practitioner | Work out why raising `EVALUATORQ_LLM_MAX_TOKENS=30_000` changed nothing | COMPLETED | 1 |
| CI engineer | Determine what an unresolved `${{ vars.X }}` does to a numeric and a boolean knob, and what to assert in CI | COMPLETED | 1 |

Every persona completed its task from the page alone. All four verified the page's claims by running them rather than reading them, and between them checked every row of the table and every one of the ten knobs. Three rounds were run, each followed by a regenerated receipt and a strict build: the reviewers, the fixes from the first six reports, then the fixes from the first-timer's report, which arrived last. No persona ended other than COMPLETED, so none was re-dispatched.

## Review

|---|---|---|
| The page said a variable that did nothing "has usually been rejected" — for the commonest case it was applied and then silently outranked | practitioner | Fixed — the opening now separates rejected, outranked and never-read, and says only the first leaves a log line |
| The table promised a `WARNING` unconditionally, but a knob only warns when something reads it, so a CI log grep gives a false all-clear when tracing is off | CI engineer | Fixed — the same paragraph now says these are read where they are used, with `ORQ_OTEL_MAX_QUEUE_SIZE` as the worked case |
| "Used to raise `ValueError` at import" — the two reviewers who checked it disagreed on whether the old readers resolved at import or at call time | hate (pedant) | Fixed by deletion — see Open decisions; an unverifiable version claim is worse than none |
| `EVALUATORQ_LOG_LEVEL` was covered by "everything else is a string handed on as it finds it", but a bad level raises and the dashboard never starts | hate (pedant) | Fixed — named as the one that does not land softly. Verified by running it |
| The two expected-output blocks were hand-typed, so they lost loguru's level padding — one space after `WARNING` where a real run prints two | conformance, hate | Fixed — both pasted verbatim from the receipt |
| The table had no row for `ORQ_OTEL_MAX_BATCH_SIZE` exceeding the queue size, which lands on a clamped value rather than the default | hate (pedant) | Fixed — row added |
| "Below the knob's minimum" was unactionable because no minimum was stated anywhere | hate (pedant) | Fixed — all six stated, plus that none has a maximum |
| Whitespace-only was covered for numbers but not for booleans; the reader had to test it | platform user, CI engineer | Fixed — one word in the table row |
| The sibling boolean rows claimed a narrower vocabulary than the code accepts, and the quick table said `1` only | hate (historian), conformance | Fixed — three rows widened, two now link to the new section |
| "The command still lists runs" contradicted the `No runs found` output directly above it | hate (pedant) | Fixed |
| The page names the CI hazard and leaves the reader to invent the mitigation | CI engineer | Fixed — added the preflight assertion the persona wrote and tested |
| Neither page says the warnings go to stderr, so a step redirecting stdout loses them | CI engineer | Fixed — one clause |
| A valid value with whitespace around it, such as `" 45 "`, was not covered — the table had whitespace-only but not whitespace-around | first-timer | Fixed — row added. Verified: `" 45 "` resolves to `45.0` with no warning |
| "Logs a `WARNING`" reads as one line per process; a knob read on every LLM call warns on every LLM call | first-timer | Fixed — one clause. Verified: three reads produced three identical warnings |
| The pasted output's second line showed a relative run-store path where the CLI prints an absolute one | first-timer, conformance | Fixed — the transcript now shows only the warning line, which is verbatim |
| "That is the failure mode worth knowing before you meet it" is a reassurance move restating the sentence above | cutter | Fixed — cut |
| "Watch it happen on a command that reads one of these knobs" announces the demonstration instead of being it | cutter | Fixed — cut |
| The `0` / `-1` clause duplicates the `EVALUATORQ_SPAN_MAX_TEXT_CHARS` table row further up the page | cutter | Fixed — cut |
| The CI contrast sentence restates the table | cutter | Left out — that sentence is the deliberate flag on the empty-value asymmetry; the table shows the rows, the sentence says what they cost |
| "Which variables this covers" bundles three categories in prose where a three-row table would partition them | conformance | Left out — it is a scope statement, not a choice the reader makes, and hate had already split it into three paragraphs |
| No probe one-liner for the two `EVALUATORQ_LLM_*` knobs, the two with a silent-override mode | practitioner | Left out — they are read inside a simulation, so a probe needs either a live run or a private function |
| `tracing.md` and `faq.md` still say `ORQ_DISABLE_TRACING=1` or `true` only | platform user | Left out — other pages, and incomplete rather than false. See Open decisions |
| Disabling tracing does not stop the results upload, which the persona expected one lever to cover | platform user | Left out — a different subject, already covered in the FAQ |
| `ORQ_DEBUG=1` prints the cleanest shell-level confirmation and appears on no page | platform user | Left out — belongs to `tracing.md` |
| The contract is a third copy of prose already in `env_config.py`'s docstring and `tracing/setup.py` | hate (historian) | Left out — rated green; the docstrings are not on the site, which is the gap being filled |
| `red_team(executive_summary=False)` raises `TypeError`; the CLI flag names do not map to the Python kwargs, and nothing says so | first-timer | Left out — a different page's subject, and a guess the persona made rather than something this page claimed |
| A first run uploads results to a live Orq workspace whenever `ORQ_API_KEY` is ambient, which the FAQ documents but "Get started in two steps" does not | first-timer, platform user | Left out — a different section. Two personas tripped on it, so it is worth someone's attention |

## Open decisions

These need a person. None is a defect in this PR.

- **Should an empty boolean warn, like an empty number already does?** Right now a number set to nothing warns and a boolean set to nothing says nothing. In CI that matters more than it sounds: if a repository variable meant to turn message capture *off* fails to resolve, the empty string reads as unset, the default is "capture on", and every prompt and response is exported with nothing in the log. The CI-engineer persona proved this end to end with a canary string, not by inference. Fixing it is a one-line change in `src/evaluatorq/common/env_config.py` plus its docstring and tests, which is outside this PR's documentation-only boundary. **Lean: fix the code and delete the two paragraphs this page spends routing around it.** The module docstring calls the asymmetry deliberate, so this is a decision, not a bug report — but a contract that needs this much prose is usually the thing to change.
- **How do readers of the other rows reach this section?** Three boolean rows now link to it; the six numeric rows do not, so someone debugging `EVALUATORQ_LLM_MAX_TOKENS=0` scrolls past the answer. **Lean: add the same one-clause link to the remaining six**, because the link lands where the reader already is.
- **Does a reference page carry version history?** The draft closed by saying the two `EVALUATORQ_LLM_*` knobs used to raise. Two reviewers checked it and reached different answers about whether the old raise happened at import or at first read, so the sentence went out rather than shipping a claim this run could not settle. The behaviour change is already in `CHANGELOG.md`, where `CLAUDE.md` says it belongs. **Lean: leave it deleted.**
- **Should the precedence rules move onto this page?** The practitioner persona's real problem was never validation: `EVALUATORQ_LLM_MAX_TOKENS` was accepted and then outranked by an `LLMCallConfig.max_tokens` that had been pinned to the default value. Precedence is three tiers (`LLMCallConfig.max_tokens`, then a per-call `max_tokens=`, then the env var) and both this page and `tuning.md` document two. `_max_tokens_advice` in `simulation/agents/base.py` already names this exact user in its docstring, and that sentence has never reached the site. **Lean: a follow-up PR on the Simulation LLM defaults section and `tuning.md`**, not a widening of this one.
- **`tracing.md` and `faq.md` describe `ORQ_DISABLE_TRACING` as `1` or `true` only.** Incomplete rather than wrong, and `tracing.md` is where a reader checking a tracing variable lands first. **Lean: a one-line follow-up on both**, kept out of here because editing unrelated pages is what the routine is told not to do.

## Code defect found, not fixed

**`EVALUATORQ_LOG_LEVEL` can stop the dashboard from starting, and it is the only variable on the page that bypasses the validation layer built for exactly this.** `src/evaluatorq/dashboard/launch.py` passes the raw environment string to `logging`, so `EVALUATORQ_LOG_LEVEL=INF0` raises `ValueError: Unknown level: 'INF0'` and the dashboard dies. Verified by running it. Every other tuning knob was consolidated onto `common/env_config.py` in `e94b9957` precisely so that a typo costs a warning and a default; this read was missed, and it is the only place the repo's own "a knob never raises" rule is broken. Someone should decide between routing it through `env_config` (which needs a small string/choice helper that does not exist yet) and validating the level name in `launch.py`. The page documents the current behaviour as an observed fact rather than as intended design, so that sentence is simply deleted when the code is fixed.

Lower priority sibling: `EVALUATORQ_DASHBOARD_ROOTS` is parsed with `json.loads` and raises on a malformed value. It is marked internal and set by `eq dashboard` itself, so a user is unlikely to reach it.

**A `CHANGELOG.md` sentence is also wrong**, found by the CI-engineer persona: the `env_bool` entry says a "set-but-empty/whitespace" value logs a `WARNING` and says the two `EVALUATORQ_*` booleans route through it. Read literally that promises an empty boolean warns. It does not. A reader who lands on the changelog first builds the wrong CI assertion.

## Reviewers, spend, and routine config

- **Missing lens: `content-evaluator`.** `docs-writing` calls for it as the broad editor alongside the three narrow lenses, and no such agent is defined in this environment. The review ran without it.
- **Partial lens: `hate`.** It ran, applied eight recommendations and returned three decisions, but reported it had no Agent tool in its session, so it could not dispatch its four critic subagents and ran the lenses sequentially itself. Its findings are verified rather than asserted, and several were confirmed independently by the conformance lens — but it was one reviewer, not four.
- **All four reader personas reported and completed.** The first-timer arrived after the other six reviewers, and its findings became the third round.
- **Live spend.** The page's own blocks make no API call, which was the disclosed estimate, and the receipt bears that out. The review did not hold to it. Three personas ran live work to check the page's claims rather than trust them: the platform user ran the quickstart evaluation and uploaded two experiment rows named `tracing-check` to the `orq-research` workspace, the CI engineer ran an evaluation with an LLM judge against a local OTLP receiver, and the first-timer drove a red team and a simulation through a fake provider — one of those red-team runs also uploaded results, because `ORQ_API_KEY` was ambient in its shell. That is the right instinct and the wrong blast radius. Disclosed here rather than left to be discovered, and the estimate at step 0 should have covered what the reviewers do rather than only what the page does. The routine still has no key of its own, so all of it landed in the normal workspace.
- **`routine.json` was not changed.**

Opened by the weekly docs-autofill routine.

https://claude.ai/code/session_01EZKnSyMpuQMLgzJj9B8cUW


Baukebrenninkmeijer:
One follow-up push, and a fourth file the body above does not describe.

**What failed.** The `Conventional title` check went red on the first two heads: `The pull request title doesn't match the commit ("docs: explain what a misconfigured env var does [axes]" vs. "docs: explain what a misconfigured env var does")`. `.github/workflows/pr-title.yml` runs `validateSingleCommitMatchesPrTitle`, and this is a one-commit PR, so the commit subject has to be the title exactly. The `[axes]` tag was in the title only.

**Fix.** Amended the commit subject to match the title. No content changed by that amendment. All 15 checks are green on `b72b43a` (`Deploy GitHub Pages` skipped, as expected on a PR).

**The fourth file.** The same push adds one paragraph to `.claude/skills/docs-autofill/SKILL.md`, recording the collision so the next weekly run does not rediscover it: the routine's own step 8 mandates a bare `[axes]` or `[BLOCKED]` tag on the PR title, and its step 6 mandates a single commit, so the two rules together fail this repo's title gate every time a tag is needed. The skill now says to amend the commit subject before pushing. It is a change to the routine rather than to the docs, which is why it is called out here rather than buried in the diff.

`routine.json` is still unchanged — that file is applied by hand and nothing in this PR touches it.

---

currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 5/5** — thorough for a docs PR, and honest about the boundary of its own scope

### Verification
Checked the two load-bearing claims against the actual code at `b72b43a9`, not just the receipt in the PR body:

- **`EVALUATORQ_LOG_LEVEL` bypasses the validation contract, confirmed.** `dashboard/launch.py:67` reads the raw env string straight into `logging.basicConfig(level=level, ...)` with no `env_config` involvement — an unrecognized level genuinely raises and kills the dashboard, the only knob on the page that breaks the repo's own "a bad value never stops a run" rule. Correctly reported as a defect and correctly left unfixed, this is a docs PR.
- **Bool/number empty-value asymmetry, confirmed.** `env_config.py`'s own module docstring states it plainly: "For a bool, empty is treated as unset" — no warning — while a number gets one. Matches what the page now documents and what's flagged in Open decisions.

Docs diff is exactly what the body claims: a new "When a value is wrong" section (`docs/configuration.md:120`) plus three consistency fixes on existing boolean rows. `.claude/skills/docs-coverage/axes.md` changes are the 4 `N/A` rows and 1 retired gap claim described, nothing sneaked in beyond that.

### 🟡 Judgment calls, already surfaced correctly by the PR itself, not mine to resolve
The PR's own "Open decisions" section already separates these out rather than burying them, which is exactly right:
- Whether an empty boolean should warn like an empty number does (a real CI/PII-adjacent risk, reproduced end to end).
- Whether the numeric-knob rows should link to the new section the way the boolean rows now do.
- Whether the LLM-knob precedence rules belong on this page or a follow-up.

None of those block this PR, and none should be quietly resolved by merging it.

### Nothing to add
The self-review already caught and fixed real inconsistencies (unconditional-WARNING claim, missing table rows, a stale version claim it correctly deleted rather than guessed at). I don't have anything on top of it.

arianpasquali (approved):

PR #201, opened 2026-09-15
ci: enable ruff C901 (max-complexity 15) with a per-file debt register

Baukebrenninkmeijer:
## Summary

Turns on mccabe cyclomatic-complexity checking (ruff `C901`, `max-complexity = 15`) via the existing `ruff check src` CI step. No new dependency, no new job.

53 functions in 33 files already exceed 15 (worst: `_run_dynamic_or_hybrid` at 108). Those files are listed in `per-file-ignores` next to the existing BLE001 debt register, so the gate binds new code now and the backlog is worked down in follow-up PRs, each of which removes its files from the register.

Why 15: violation counts at each ceiling were 8→162, 12→77, 15→53, 20→22, 25→12. Above 20 the gate stops meaning anything; 15 is the knee.

Known gap, stated in the config comment: per-file ignores exempt a *new* over-limit function in a registered file too. The follow-up PRs shrink the list to zero, which closes it.

## Test plan

- [x] `uv run ruff check src` passes on this branch
- [x] A synthetic complexity-16 function trips `C901 ... (16 > 15)`
- [x] `uv run ruff format --check src` unchanged


currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 5/5** — config-only change, verified for real, no gaps

### ✅ C901 gate is genuinely binding
Confirmed `"C90"` in select (pyproject.toml:84), `C901` dropped from the ignore list, `max-complexity = 15` set (pyproject.toml:138-143). Ran ruff myself at this SHA: `All checks passed!` with the gate on. Dropping the ceiling to 14 for a sanity check turns up 59 errors, so the gate is doing real work, not decoration.

### ✅ Debt register is exact, not padded
Computed the true pre-exception violation count by clearing per-file-ignores: 53 functions across 33 files. The register added in this PR lists exactly those 33 files. 0 padding entries, 0 gaps. Worst offender is `_run_dynamic_or_hybrid` in runner.py at complexity 108, matches the comment in the diff.

### ✅ CI green, nothing escapes the gate
Everything over 15 is covered by an exception, so merging doesn't turn CI red on main. This PR is independent of the follow-up burn-down PRs (202-205) and can merge on its own.

### 🟡 Worth knowing, not blocking
The register is per-file, so C901 is fully off for those 33 files until the burn-down PRs land — a new 40-complexity function added to runner.py today would merge silently. Also: the lint only covers `src`, not `tests/` or `examples/`. Both already called out in the PR's own comment.

Baukebrenninkmeijer:
Decision for tracking: `max-complexity` stays at 15 for now. A projection on the merged stack shows about 38 functions in 26 files over 12, 15 of them scoring exactly 15 and none above, so lowering to 12 would need a second burn-down round. Revisit once the stack has merged and the per-file register is empty.

PR #203, opened 2026-09-15
refactor: split request assembly from response handling in nine call paths (C901 burn-down 4/4)

Baukebrenninkmeijer:
PR D of the four-PR C901 burn-down. Every change here is behaviour-preserving: nested coroutines hoisted to module level, response-shape ladders extracted into named helpers, an aggregation `if` ladder replaced by a dispatch table, and one pagination/parse loop split from the request that produced it. No public signature changed, no log message changed, no error text changed, no test assertion changed.

## Function → C901 score

| Function | File | Before | After |
|---|---|---|---|
| `_call_responses_api` | `openresponses/target.py` | 22 | 3 |
| `_do_call` | `openresponses/target.py` | 19 | 11 |
| `respond` | `integrations/langgraph_integration/target.py` | 19 | 10 |
| `execute_response` | `common/llm_call.py` | 16 | 7 |
| `_run_jury_core` | `common/jury.py` | 17 | 15 |
| `llm_jury` | `llm_jury.py` | 16 | 10 |
| `fetch_experiment_datapoints` | `fetch_data.py` | 19 | under 15 |
| `fetch_agent_info` | `simulation/utils/run_store.py` | 16 | under 15 |
| `_resolve_or_generate_datapoints` | `simulation/api.py` | 17 | under 15 |
| `run_streaming_evaluation` | `evaluatorq.py` | 18 | under 15 |

`_run_jury_core` lands at exactly 15, the ceiling. Anything added to it re-breaks the gate.

## Debt register

Removed from the `C901` entry in `[tool.ruff.lint.per-file-ignores]`: `openresponses/target.py`, `integrations/langgraph_integration/target.py` (keeps `BLE001`), `common/jury.py`, `common/llm_call.py`, `llm_jury.py`, `fetch_data.py` (keeps `BLE001`), `simulation/utils/run_store.py` (keeps `BLE001`).

Two files keep their entries on purpose. `simulation/api.py` still holds `_simulate_core` (27) and `evaluatorq.py` still holds `evaluatorq` (32, down from 37 only as a side effect of hoisting children of the nested `run_streaming_evaluation`). Both are bucket 2 in the plan, explicitly out of scope, so their entries stay until that work lands.

## Checks run

```
uv run ruff check src              → All checks passed!
uv run ruff format --check src     → 233 files already formatted
uv run basedpyright                → 0 errors, 0 warnings, 0 notes
uv run pytest -m 'not integration' → 5470 passed, 2 skipped, 10 deselected
```

Each task also ran its own before/after baseline on its own test files and matched the count exactly: `tests/openresponses` 93, langgraph unit + `tests/integrations` 58, `tests/common` + panel-of-judges + llm_jury + reuse guardrails 613, `tests/simulation` 1292, `tests/test_send_results.py` 28.

## Rulings made during the work

**Formatter reflow from a dedent is allowed.** Hoisting a nested body dedents it, and `ruff format` then collapses expressions that no longer need wrapping. A reviewer flagged two such lines in `openresponses/target.py` as forbidden reformatting. Preserving the original wrapping would fail `ruff format --check src`, which is a CI gate, so the reflow stands. This covers only reflow forced by the dedent — untouched code was not reformatted anywhere.

**`simulation/api.py` keeps its `C901` entry.** The plan's Task D4 text says to remove three entries, but its own Out-of-scope section keeps `simulation/api.py` on the register because `_simulate_core` is un-refactored. Removing the entry would have turned `uv run ruff check src` red. The Out-of-scope section wins.

## Notes for review

`langgraph_integration/target.py` reuses nothing from `common.output_adapters`: `output_to_text`, `output_to_messages`, `_adapt_tool_call`, `inputs_to_messages` and `coerce_content_text` were each checked, and none preserve LangGraph's state-message traversal and tool-result pairing semantics. Behaviour preservation outranked deduplication here.

`common/llm_call.py`'s new `_build_response_params` calls `LLMCallConfig.request_params` rather than rebuilding the parameter dict, so the reserved-key guard still runs.

One deferred minor, not fixed because it is outside the refactor: `integrations/langgraph_integration/target.py` initialises `output_items` and overwrites it unconditionally a few lines later. The final review agreed it is not a merge blocker.


currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 4.5/5** — heaviest verification of the stack, holds up

### ✅ Full suite + ruff confirmed for real
5454 passed, identical on base and head. 12 C901 offenders on base down to 2 (both explicitly out of scope, both keep their exceptions) — `_run_jury_core` lands exactly at 15, zero headroom, matches what the PR says.

### ✅ Error-handling scope checked per call path, not assumed
This is the risky part of this kind of split — an except block silently narrowing to cover only half of what it used to. Checked all 9 individually: openresponses/target.py's BadRequest retry still wraps both attempts, the TimeoutError still wraps the full retry, fetch_data.py's JSONDecodeError→ValueError and empty-rows raise both moved together into the same helper. No scope regression found anywhere.

### ✅ State threading verified correct
evaluatorq.py and llm_jury.py both moved from `nonlocal` to list-boxed mutable state (`[0]`, `[False]`, `[client]`) — checked every read/write site, no stale-copy bug.

### ✅ The real rewrite (not just extraction) got the most scrutiny
`_build_response_params` in llm_call.py isn't a pure move — it now builds a fresh pydantic model per call instead of dict updates. Ran a differential harness across 10,206 input combinations against it: 0 mismatches, including identical exception types/messages. Same treatment for `_do_call` against a fake AsyncOpenAI client (6 scenarios) and `_run_jury_core` (30 cases) — all byte-identical to base.

### 🔴 Needs live testing — flagging so these don't get counted as verified
Four paths only exercise on a real backend: the 400-retry-drop-reasoning path in openresponses/target.py:182-198, orq-router-only request fields (thread_body_param, memory.entity_id), the 404-vs-other status split in fetch_agent_info, and the langgraph integration target (needs the langgraph runtime). Mocks can't trip these, so they're unverified by anything I ran.

### 🟡 Minor
langgraph_integration/target.py:385 has a dead list initializer, already acknowledged in the PR body. `_build_response_params` being an actual rewrite rather than pure extraction is worth knowing even though it's behavior-identical.

Baukebrenninkmeijer:
# PR Review — `git diff Baukebrenninkmeijer/cicd-complexity-checks...Baukebrenninkmeijer/c901-d` (PR #203)

Lenses run: code (general quality), errors (silent failures), types (type design), comments, spec, minimal. Skipped: tests (no test file changed; the existing suite was used as the behaviour oracle instead, and it holds — 5470 passed before and after). Failed: none.

Repo checks on the PR head, before any change: `uv run ruff check src` clean, `uv run ruff format --check src` clean, `uv run basedpyright` 0 errors, `uv run pytest -m 'not integration'` 5470 passed / 2 skipped. The behaviour-preservation claim survived a line-by-line trace on every file: no exception scope moved, no log string changed, no merge precedence flipped, no default value shifted, and every previously-captured closure variable is threaded through with the same read/write timing. Nothing below is a correctness regression.

What follows is about the *shape* of the result. Three of the ten refactors move complexity across a function boundary without reducing it, and one of those introduces a sentinel value that is a genuine trap for the next editor.

## Critical (0)

None.

## Important (4)

- **[code, minimal, types, comments — convergent, 4 lenses] `src/evaluatorq/common/llm_call.py:353-395` — `_build_response_params` fabricates a throwaway `LLMCallConfig(max_tokens=1)` purely to borrow `request_params()`, then hand-patches away the two artefacts that borrowing introduces.** `execute_response` has no config; it takes loose scalars. `request_params` unconditionally writes `base['max_output_tokens'] = self.max_tokens`, so the sentinel `1` lands in the result and line 387 pops it back out whenever the caller asked for no token cap. Separately, `request_params` deliberately strips `temperature` when it resolves to `None` (contracts.py:334-338, with the comment "the provider rejects an explicit null"), and line 389 re-inserts the null to match this path's pre-existing contract. I traced every branch — unset cap, real cap, cap via `extra_kwargs`, `temperature=None` via `extra_kwargs` — and the net output matches the pre-PR code exactly. It is correct today only because of those two compensating patches. The next field `request_params` adds unconditionally to `base` leaks into every Responses call built here with nothing to pop it, and a maintainer who "cleans up" the sentinel pop caps every uncapped Responses call at one token. The CLAUDE.md row this was reaching for ("Building call params → `LLMCallConfig.request_params`; never a hand-built `extra_kwargs` dict — it skips the reserved-key guard") does not bite: the pre-PR code called `check_reserved_keys(extra_kwargs, _RESERVED_RESPONSES_KEYS)` directly, so the guard was never missing. Structural fix: keep the extraction (that is what buys the C901 win) and drop the config detour, restoring the literal `extra_body` / `extra_kwargs` handling inside the new helper, with a docstring saying why this path does not route through the config. Cost: none — the helper stays at complexity 10 and `execute_response` stays at 7.
- **[minimal, types, comments — convergent, 3 lenses] `src/evaluatorq/common/jury.py:178-233` — the `_AGGREGATION_HANDLERS` dispatch table is 55 lines expressing a two-way branch, and it adds a fourth hand-maintained mirror of the aggregator keyword set that no parity test covers.** `_aggregate_mode` is the only handler with distinct behaviour; `_aggregate_default` is registered five times to fill out the dict, and both handlers `del` half of the shared four-parameter signature because neither implementation fits it. `_aggregate_panel_verdict` then re-derives the handler with a second `isinstance(aggregator, str)` check, duplicating a classification `_run_jury_core` already made at line 727. The repo's own rule for a hand-maintained registry (assert `set(Enum) == set(registry)` at import, freeze with `MappingProxyType`) is not followed, and `tests/common/test_jury.py::test_aggregator_registry_parity` pins `_AGGREGATORS` and `_AGG_KIND` but never mentions the new table — so a sixth aggregator added later silently falls through to `.get(aggregator, _aggregate_default)` with no signal that a tie-handling decision was skipped. The original was an eight-line `if aggregator == 'mode': ... else: ...`. Structural fix: keep the extraction for the complexity budget, but put the original `if`/`else` inside `_aggregate_panel_verdict` and delete the table, the type alias and both handler stubs. `_run_jury_core` stays at exactly 15 either way.
- **[types, minimal — convergent] `src/evaluatorq/openresponses/target.py:71-186` — `_create`, `_normalise_response` and `_do_call` were hoisted to module level but still take `self: OrqResponsesTarget` as a parameter, so they are methods wearing a free function's clothes.** Between them they read `self._client`, `self.config`, `self.tools`, `self.instructions`, `self.memory_entity_id`, `self.retry_attempts` and `self.retry_statuses` — nearly the whole object, not a couple of captured locals. The call sites are also inconsistent: `_create(self=self, ...)` and `_do_call(self=self, ...)` pass it by keyword while `_normalise_response(self, ...)` passes it positionally. The closure that the hoist was meant to remove did not go away either; it reappeared one line down as `with_retry(lambda: _do_call(self=self, ...))`. Nothing marks `self` as special any more, so a second target instance passed in beside it would type-check. Structural fix: make all three private methods and call `with_retry(lambda: self._do_call(...))`. C901 is unaffected — a method's complexity is its body's, and `_do_call` measures 11 either way.
- **[minimal, types, comments — convergent, 3 lenses] `src/evaluatorq/evaluatorq.py:376-476` — the streaming path now carries two different "this is secretly a mutable box" conventions side by side.** `total_datapoints` became `list[int]` and `stop_polling` became `list[bool]` so the hoisted `_poll_progress` can observe them, while `progress_ref` a few lines away is already a `dict[str, int]` doing the identical job. At the call site `_poll_progress(progress, total_datapoints, progress_ref, stop_polling)` a reader cannot tell which boxed argument is which without opening the body, and a maintainer who "simplifies" `stop_polling: list[bool]` to a plain `bool` silently breaks the poller's stop signal. Structural fix: one small mutable dataclass holding `total`, `processed` and `stop`, passed by reference to both helpers. That removes all three cells, the stringly-keyed dict access, and two of the eight positional parameters.

## Suggestions (7)

- **[spec] PR body, complexity table — two "after" figures are understated.** `fetch_experiment_datapoints` (fetch_data.py) and `_resolve_or_generate_datapoints` (simulation/api.py) are both listed as "under 15" but measure at exactly **15**, the ceiling. That is the same at-the-limit fragility the body explicitly discloses for `_run_jury_core`, and it deserves the same warning: anything added to either function re-breaks the gate. Every other figure in the table checks out exactly (`_call_responses_api` 22→3, `_do_call` 19→11, `respond` 19→10, `execute_response` 16→7, `_run_jury_core` 17→15, `llm_jury` 16→10, `fetch_agent_info` →3, `run_streaming_evaluation` →13, `_simulate_core` 27, `evaluatorq` 32). The `pyproject.toml` debt-register diff also matches the body exactly, with no stale `C901` ignore left behind and none removed from a file that still needs one.
- **[comments] `src/evaluatorq/fetch_data.py:205` — `_rows_from_page(page, ...)` is named for a pagination loop that does not exist.** An experiment export is a single body; the only caller passes the whole `export_text` after the download completes. The PR body repeats the mistaken model ("one pagination/parse loop split from the request"). Rename it to say what it does — parse one JSONL export body — and note in a docstring that it raises rather than returning empty, because that raise used to live outside the `async with` block and its scope is easy to misread now that it moved.
- **[minimal, comments] `src/evaluatorq/llm_jury.py:404-547` — `_llm_jury_scorer` takes 26 keyword arguments bound by `functools.partial`, and `_llm_jury_judge_fn` duplicates `_run_single_judge`'s full 13-parameter signature to forward it unchanged.** Neither is wrong. The judge adapter is load-bearing: `run_jury` calls `judge_fn(judge_model)` positionally while `_run_single_judge` is keyword-only, so a bare `partial` cannot bridge them — but nothing in the code says so, and it reads as gratuitous. The 26-argument scorer is the price of hoisting a closure whose captured environment really was that large; whether that is better than the closure it replaced is a judgement call I have left to the author (see the follow-up comment). At minimum both need a docstring, and `client_state` needs a line saying it is a single-element mutable cell rather than a list of clients.
- **[comments] `src/evaluatorq/common/jury.py` — the rationale comment explaining why `mode` is special-cased was deleted during the extraction and not re-homed.** The one comment that did survive ("Pass ALL votes — custom callables may want abstained/failed votes") now sits above the generic dispatch call, where it describes only the default branch; the `mode` path structurally ignores `votes`. Both belong on the extracted function.
- **[comments] `src/evaluatorq/integrations/langgraph_integration/target.py:98` — `_extract_tool_calls` also extracts text items.** It builds the entire ordered output list, interleaving included; the name describes half of it. `_extract_final_message` next to it both mutates the list it is handed and returns it, which leaves the caller's `output_items = _extract_final_message(result_messages, output_items)` looking like a rebind when it is not.
- **[spec, code] `src/evaluatorq/integrations/langgraph_integration/target.py:385` — the deferred defect the PR body admits is real and harmless.** `output_items: list[OutputMessage] = []` is unconditionally overwritten thirteen lines later and nothing reads it in between. It is one line to delete rather than to carry.
- **[minimal] The claim that nothing in `common.output_adapters` could be reused by the LangGraph helpers holds up.** `output_to_text`, `output_to_messages`, `_adapt_tool_call`, `inputs_to_messages`, `coerce_content_text`, `content_to_text` and `tool_result_to_text` all operate on `Output`/`OutputMessage`/OpenResponses wire shapes, never on the LangChain `AIMessage`/`ToolMessage` objects (`get_message_type`, `.tool_call_id`, `.tool_calls`) that this file's local `_lc_*` helpers handle. Recording it here so the question is not re-litigated on the next pass.

## Conflicting verdicts

None. Where lenses overlapped they agreed on the defect and differed only in severity; the higher claim is used above.

## Next

1. Drop the fabricated `LLMCallConfig` from `_build_response_params` — it is the one finding whose failure mode bites a future editor rather than this diff.
2. Collapse the jury dispatch table back to its two branches, and re-home the `mode` rationale comment onto the extracted function.
3. Make the three `openresponses/target.py` helpers methods again.
4. Replace the streaming mutable cells with one dataclass.

Baukebrenninkmeijer:
## Questions for the author

@Baukebrenninkmeijer — three things I could not decide on your behalf. Everything else from the review is fixed and pushed.

1. **`_llm_jury_scorer`'s 26 keyword arguments.** The finding: hoisting the `scorer` closure out of `llm_jury()` turned its captured environment into a 26-parameter signature bound by `functools.partial`. It is correct and it does lower `llm_jury` from 16 to 10, but a 26-argument function is harder to read than the closure it replaced, and every future field has to be threaded through two places. The options are (a) leave it as is, (b) bundle the configuration into one frozen dataclass — something like `_JuryScorerConfig` — so the scorer takes `params` plus one config object, or (c) put the closure back and find the complexity elsewhere in `llm_jury`. My recommendation is (b): it is mechanical, changes no behaviour, keeps the complexity win, and makes the partial's payload a named thing rather than a list. I did not do it because it is a larger edit than the rest of this pass and it is your call whether it belongs in a burn-down PR or a follow-up. I added docstrings to both `_llm_jury_scorer` and `_llm_jury_judge_fn` in the meantime so the shape is at least explained.

2. **Two functions sit at exactly the C901 ceiling and the PR body does not say so.** `fetch_experiment_datapoints` and `_resolve_or_generate_datapoints` are listed in the table as "under 15" but both measure exactly 15, the same position you explicitly flag for `_run_jury_core` ("anything added to it re-breaks the gate"). The options are (a) correct the PR body so all three at-the-ceiling functions carry the same warning, or (b) spend a little more on one or both of them now so there is headroom. My recommendation is (a) — the numbers are accurate as a pass/fail statement, the disclosure is what is missing — but (b) is worth considering for `fetch_experiment_datapoints` specifically, since it is the function most likely to grow when the export API changes. I did not edit the PR body myself.

3. **`_llm_jury_judge_fn` exists purely to bridge a calling convention.** `run_jury` invokes `judge_fn(judge_model)` positionally while `_run_single_judge` is keyword-only, so a plain `functools.partial` cannot fill `model`. The adapter is therefore load-bearing, not redundant — but it duplicates a 13-parameter signature to forward it unchanged, which is a maintenance edge every time `_run_single_judge` gains a parameter. The alternative is to widen `_run_single_judge` to accept `model` positionally and delete the adapter entirely. My recommendation is to leave it and keep the docstring I added, because making one parameter positional on a 13-keyword function trades a small duplication for a call-site ambiguity, but you own that signature and may prefer the deletion.

Baukebrenninkmeijer:
Closed by GitHub when the merged base branch was deleted; it could not be retargeted because GitHub had grouped it into a native stack. Same branch, same commits, reopened as #223 against main.

PR #206, opened 2026-09-16
refactor: remove the deprecated Streamlit dashboards

Baukebrenninkmeijer:
`eq redteam ui` and `eq sim ui` launched Streamlit dashboards and have printed `Warning: eq … ui is deprecated; use eq dashboard instead.` since July 2026. This removes them.

## What is removed

- The `ui` typer command on both `eq redteam` and `eq sim`.
- `src/evaluatorq/redteam/ui/`, `src/evaluatorq/simulation/ui/` (including `token_display.py`) and `src/evaluatorq/common/ui/` (`launch_streamlit`) — roughly 3,400 lines of Streamlit view code with no remaining importer.
- `streamlit`, `plotly` and `watchdog` from the `redteam` and `simulation` extras. `redteam` is now `huggingface-hub` + `vl-convert-python`; `simulation` is `orq-ai-sdk` + `vl-convert-python`. `vl-convert-python` stays — `common/reports/vega.py` renders every chart through it. The `watchdog` pin in the `dev` dependency group is untouched; it is there for the docs tooling, not for the removed viewers.
- The two `redteam/ui` / `simulation/ui` entries from the basedpyright `exclude` list, and the `redteam/ui/dashboard.py` line from the ruff per-file-ignores register.

## The replacement

`eq dashboard`, the FastHTML dashboard, reads the same saved run JSON files from the same default stores. There is no migration beyond the command name: `eq redteam ui path/to/report.json` becomes `eq dashboard path/to/report.json`, and a bare `eq dashboard` browses both stores.

## This is user-facing, and deliberately not marked breaking

Two commands a user could type are gone, and two extras install fewer packages. It is landed as `refactor:` rather than `feat!:` because a breaking marker halts every release until a human forces a bump, and that trade is not worth making for a path that has been printing a deprecation warning at every invocation for months. The removal is described in `CHANGELOG.md` under `Breaking Changes` so a reader upgrading finds it there.

## Tests

Removed (their only subject was the Streamlit path): `test_ui_warns_that_dashboard_is_deprecated` in `tests/redteam/test_cli.py`; the five `test_ui_*` cases in `tests/simulation/test_cli.py`; `test_redteam_ui_help` / `test_sim_ui_help` in `tests/dashboard/test_launch.py`; the four `_FakeStreamlit`-driven `test_sim_ui_dashboard_*` cases in `tests/simulation/reports/test_criteria_audit_surfaces.py`; the two `token_display` cases in `tests/unit/test_simulation_token_reporting.py`.

Kept and adjusted: `test_eq_dashboard_help` loses its "deprecated aliases remain callable" comment; the criteria-audit file keeps its scorer, report-section, HTML-export, Markdown-export and FastHTML-dashboard coverage of the same behaviour, so every rule the deleted cases pinned is still pinned on a live surface. `tests/redteam/test_export_html_charts.py` — the guardrail that the red-team HTML export uses Vega-Lite rather than plotly — is unchanged and now has nothing left to catch it out.

## Prose and docs

`docs/installation.md` (extras table), `docs/cli-reference/dashboard.md`, `docs/dashboard.md`, `docs/cli-reference/overview.md` and `docs/cli-reference/simulation.md` are updated — the "Primary UI" admonitions were hedges against the Streamlit viewers and now say "The UI". `ROADMAP.md`, `src/evaluatorq/redteam/ARCHITECTURE.md`, `CLAUDE.md`'s extras note, and the `.claude/skills/docs-coverage/axes.md` impossible-combination row are corrected. A dozen "parity with the Streamlit dashboard (dashboard.py:NNN)" comments in `dashboard/` and `redteam/reports/export_html.py` pointed at line numbers in files this PR deletes; they are reworded to describe the behaviour instead.

## Checks

`ruff check src`, `ruff format --check src`, `basedpyright` (whole repo), `pytest -m 'not integration'` (5456 passed), `mkdocs build --strict` and `scripts/validate_mermaid.py` all pass. `uv.lock` is regenerated and committed.

## One thing left alone

`docs/cli-reference/redteam.md`, `docs/cli-reference/simulation.md` and `docs/cli-reference/overview.md` each claim the subcommand group is "registered only when the extra is installed". That is false — `src/evaluatorq/cli.py` registers both unconditionally and says so in its own docstring. It is a pre-existing error, not one this PR introduces, so fixing it here would be a drive-by; raising it instead.


currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 5/5** — clean deletion, verified nothing dangles

### ✅ Deprecation predates this PR, not just asserted
Confirmed at base: `redteam/cli.py:701` and `simulation/cli.py:2093` both already emitted a deprecation warning pointing at `eq dashboard`, introduced back in July 2026 (133c56d6). This is following through on an already-announced sunset, not a surprise removal.

### ✅ No dangling references anywhere
Grepped src, tests, docs and pyproject.toml for every deleted module path (redteam.ui, simulation.ui, common.ui, token_display, launch_streamlit, streamlit/plotly/watchdog imports) — one hit, the CHANGELOG entry itself. Confirmed myself: nothing in `src/` references streamlit or the old ui paths anymore.

### ✅ CLI verified live at head
`eq --help` lists dashboard/redteam/sim, no crash on startup. `eq redteam ui` and `eq sim ui` both correctly error `No such command 'ui'`. `eq dashboard` (the replacement) still works.

### ✅ Tests: 5448 passed, 1 pre-existing failure unrelated (same live-catalogue drift issue seen across this whole stack, reproduces on base too). Deleted tests all targeted deleted code — checked that the criteria-audit test file was only partially trimmed and confirmed the equivalent FastHTML-path tests for that same coverage still remain.

### ✅ The 44 added lines are honest
One CHANGELOG entry under Breaking Changes, plus doc/docstring cleanup dropping "Streamlit parity" references and removing the now-dead ruff/pyright excludes for the deleted dashboard file. `pyproject.toml` correctly drops streamlit/plotly/watchdog from the extras.

🟡 One cosmetic nit: dashboard/sim_views.py:540-541 has an over-long comment line left over after the parity note was stripped. Not blocking.

Baukebrenninkmeijer:
Decision for tracking: this PR removes the Streamlit dashboards and their CLI subcommands, which is breaking in substance, but it deliberately stays a `refactor:` commit so it does not block releases with a computed major bump. The next `feat:` or `fix:` on main ships it as a minor or patch release; the removal is called out in this PR body and in CHANGELOG under Unreleased.

PR #207, opened 2026-09-16
refactor(simulation): split _simulate_core into resolve, execute, persist and publish (C901 bucket 2)

Baukebrenninkmeijer:
Part of the C901 bucket-2 burn-down. `simulation/api.py::_simulate_core` was a 318-line orchestration that ran datapoint resolution, run-settings derivation, the whole confirm/run/complete lifecycle with its nine-clause exception stack, report publication and persistence in one body. This splits it along the phase boundaries the source already marked with comments, and takes the file off the `C901` debt register in `pyproject.toml`.

Nothing about behaviour changes. Every `await`, `try`/`except`/`finally`, lifecycle hook (`on_confirm`, `on_stage_start`, `on_run_start`, `on_run_complete`, `on_stage_end`), manifest-writer call (`complete`, `cancel`, `fail`), `logger.*` line and `set_span_attrs` call keeps its text, its arguments and its position relative to the others. `_simulate_core`'s own signature, docstring and return type are byte-identical. Every import that was function-local stays function-local — this module's lazy imports are load-bearing, and the first review round caught an attempt to hoist four of them to module scope.

## Complexity

| Function | Before | After |
|---|---|---|
| `_simulate_core` | 27 | 8 |
| `_resolve_max_turns` (new) | — | 1 |
| `_resolve_run_settings` (new) | — | 4 |
| `_execute_simulation` (new) | — | 10 |
| `_publish_run` (new) | — | 4 |
| `_persist_run` (new) | — | 5 |

`_RunSettings` and `_ExecutionOutcome` are frozen dataclasses carrying the phase outputs, so no phase takes a mutable bag to fill in.

## Checks run

```
uv run ruff check src/evaluatorq/simulation/api.py --select C901 --config "lint.per-file-ignores={}"   # All checks passed!
uv run ruff check src
uv run basedpyright                 # whole repo: 0 errors, 0 warnings, 0 notes
uv run pytest -m 'not integration'  # 5470 passed, 2 skipped, 10 deselected — identical to the pre-change baseline
```

## Rulings

`SimulationRunMeta` now lives in the module's `TYPE_CHECKING` block rather than as a runtime function-local import. It appears only in annotation position (three sites), the module carries `from __future__ import annotations`, and nothing introspects it at runtime — so the name is never needed at import time. If that is wrong the cost is an `AttributeError` on a `get_type_hints` call that does not exist today.

Based on `Baukebrenninkmeijer/c901-d` because that branch already touched `_resolve_or_generate_datapoints` in this same file. `Baukebrenninkmeijer/c901-s2` stacks on top of this one for `simulation/runner/simulation.py`.

## Known minor, deliberately left

`_RunSettings.target_label` is written but never read by the caller — `target_label` is a genuine local inside `_resolve_run_settings`, where it builds `run_meta['target']`, but exposing it as a field on the returned dataclass buys nothing. It is dead output state. Deleting it means a follow-up commit on this branch and a rebase of the stacked #210, which costs more than the field does; flagging it here so nobody spends a review round rediscovering it. Happy to drop it if a reviewer would rather have it gone.


currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 5/5** — clean split, verified with real offline scenarios not just diff-reading

### ✅ Byte-identical code motion
Diffed ignoring whitespace: the only non-scaffolding lines touched are two re-formed imports and one opener line. Every moved function body is untouched.

### ✅ No closure-variable leak, the classic bug in this refactor pattern
Checked `target_label` specifically since it's the field most likely to get dropped in a split like this — confirmed on base it's used only inside the `run_meta` dict, so its absence from the re-bound locals in `_simulate_core` is correct, not a miss. `results` and `experiment_url_out[0]` now thread cleanly through an `_ExecutionOutcome` object.

### ✅ Error-handling scope preserved exactly
The full `on_confirm` gate → try/except `SimulationDroppedError`/`BaseException`/finally block moved verbatim into `_execute_simulation`, including the traceback re-raise. Confirmed the outer try still covers it at the call site — `SimulationCancelledError`/`BaseException` handling is unchanged.

### ✅ Publish/persist ordering unchanged, verified directly
`_publish_run` (report enrichment) still runs before `_persist_run` (write + manifest complete) — same order as before the split, checked both the function bodies and the call site.

### ✅ Ran real offline scenarios, base vs head
Used the repo's own stub user-sim/judge harness (no LLM needed) for two scenarios — a 2-datapoint/3-turn run and a raising-target failure case. Full hook-event traces and persisted report JSON compared: identical on both sides.

### ✅ ruff C901 confirmed
`_simulate_core` was 27 on base, all split functions pass at max-complexity 15 on head — genuinely fixed, ignore correctly dropped.

🟡 Two cosmetic nits, non-blocking: `_RunSettings.target_label` (api.py:1500) is a dead field, never read anywhere. And the comment on `_persist_run` ("CLI-only report enrichments run before persistence") describes what `_publish_run` does now, not `_persist_run` — leftover from before the split.

Baukebrenninkmeijer:
# PR Review — `gh pr diff 207` (head `Baukebrenninkmeijer/c901-s1`, base `Baukebrenninkmeijer/c901-d`)

Lenses run: code, minimal, comments, types, errors, tests. Skipped: spec (no ticket id resolves from the branch name, the commits, or the PR title/body, so the PR body itself is the statement of intent and the `code` lens judged against it). Failed: none.

Repo checks, run verbatim from `CLAUDE.md` before dispatching any lens: `uv run ruff check src` passed, `uv run ruff format --check src` passed (233 files), `uv run basedpyright` passed on the whole repo (exit 0), `uv run pytest -m 'not integration'` passed. No Critical findings come from the tooling.

The central claim of this PR — that the extraction is behaviour-preserving — was verified rather than taken on trust. The `errors` lens diffed each extracted body against the corresponding inline block in `origin/Baukebrenninkmeijer/c901-d` and found them byte-identical modulo indentation, confirmed that `_execute_simulation` is awaited inside the same outer `try` so both `except` arms still see every exception path (including the re-raise from its own `finally`), confirmed that `return _ExecutionOutcome(...)` is unreachable on every raising path exactly as the post-try code was before, and confirmed `run` is still guaranteed bound at `return run`. That lens returned NONE. The `SimulationRunMeta` move into `TYPE_CHECKING` also checks out: the module carries `from __future__ import annotations`, no `get_type_hints` call and no pydantic model touches either new dataclass or the TypedDict, so the annotation is never evaluated at runtime. The findings below are therefore all about the shape of the result, not about a behaviour regression.

## Critical (0)

None.

## Important (5)

- [code, minimal, types] `src/evaluatorq/simulation/api.py:1500` — `_RunSettings.target_label` is dead output state. It is computed inside `_resolve_run_settings`, stored on the returned dataclass, and never read by any caller; `run_meta['target']` is built from the local variable before the dataclass exists, and `_simulate_core` only reads `evaluator_names`, `target_name`, `target_model` and `run_meta` off it. The PR body acknowledges this and asks whether to drop it. The repo's own house rule is "a helper with no caller is a bug" — a field with no reader is the same defect one level down. Structural fix: delete the field and keep `target_label` as a plain local inside `_resolve_run_settings`. Cost: two lines. Convergent across three lenses.
- [types] `src/evaluatorq/simulation/api.py:1502` — `_RunSettings.target_model` is typed `Any`, and the call site at line 1878 immediately writes `target_model=target_model if isinstance(target_model, str) else None`. An `isinstance` narrowing at the consumer is the signal that the producer's type is wrong: the field is being asked to carry "whatever `getattr` returned" when the only value `build_simulation_run` accepts is `str | None`. Structural fix: narrow once in `_resolve_run_settings`, directly after the `getattr` chain that produces it, type the field `str | None`, and delete the call-site `isinstance`. Cost: one line moved, one line deleted, and the type checker now enforces the invariant instead of a runtime check restating it.
- [code, minimal] `src/evaluatorq/simulation/api.py:1520-1530, 1687` — `_resolve_run_settings` and `_publish_run` each receive `config` and, alongside it, scalar parameters that are already fields of that same `config`. `_resolve_run_settings` takes `model`, `max_turns`, `datapoint_parallelism` and `evaluation_name`; `_publish_run` takes `model`. By the time either is called, `config` has already been rebound via `model_copy` with the resolved `max_turns`, so every one of those values is identical to the corresponding `config.*` field. The pre-refactor code had no such hazard because it read plain locals once with no second signature holding both the alias and its source. This is a newly introduced duplicate source of truth: a future change that rebinds `config` again between the alias assignment and the call silently diverges, and nothing in the signature says the two must agree. Structural fix: drop the redundant keywords and read the fields off `config` inside the helpers, then delete the now-unused `model`, `datapoint_parallelism`, `save` and `run_output` local aliases in `_simulate_core`. Cost: a smaller signature and one fewer way to be wrong. Convergent across two lenses. The same argument applies to `_persist_run`, which takes `save` and `run_output` as loose scalars while its sibling `_publish_run` takes `config` — the two helpers should agree on how they are fed.
- [comments] `src/evaluatorq/simulation/api.py:1590, 1711, 1535, 1860, 1841` — five comments moved verbatim with their code and their spatial references no longer resolve. `_execute_simulation` says a decline becomes "`SimulationCancelledError` → writer.cancel below", but `manifest_writer.cancel()` is now in `_simulate_core` and `manifest_writer` is not even a parameter of `_execute_simulation`. `_persist_run` opens with a comment about "CLI-only report enrichments run before persistence" that describes the recommendations and executive-summary calls, which now live in `_publish_run` — the comment is orphaned above a body it does not describe. `_resolve_run_settings` says it mirrors "the save block's precedence", and the save block no longer exists under that name. `_simulate_core` says an aborted run is "re-raised above" when the re-raise is now inside a function it calls, and says failures are guarded "including a hook raising before/inside the inner try or in its finally" when there is no longer an inner try at that level. Structural fix: replace the spatial references with the function names that now own the behaviour, and move the orphaned `_persist_run` comment to `_publish_run` where its subject went. Cost: wording only, no code change. This matters more than usual here because these comments are the only place the hook-ordering and exception-preservation contract is written down.
- [comments] `src/evaluatorq/simulation/api.py:1512, 1520, 1574, 1687, 1711` — none of the five new helpers has a docstring, against the convention of the file they were extracted from. `_resolve_target`, `_resolve_or_generate_datapoints`, `_simulate_via_evaluatorq`, `_attach_recommendations`, `_agent_key_of`, `_compose_sim_hooks` and `_log_saved_run` all carry one; only genuinely trivial one-liners go without. `_execute_simulation` is the acute case: it is roughly 110 lines encoding a confirm-gate, stage-bracket, drop-handling and terminal-hook contract that a caller must know, and right now the only way to learn it is to read every inline comment in the body. Structural fix: give each of the five a docstring stating its contract in the caller-facing terms the rest of the file uses, and for `_execute_simulation` and `_persist_run` specifically state what they raise and what they swallow. The stale spatial comments above then have somewhere accurate to point.

## Suggestions (4)

- [tests] `src/evaluatorq/simulation/api.py:1747` — the save-failure path in `_persist_run` must still call `manifest_writer.complete(report_path=None, ...)`, and nothing pins that. The existing `tests/simulation/test_run_store.py::test_simulate_save_failure_still_returns_results` checks that the results survive an `auto_save_run` explosion but never looks at the manifest. If the extraction had misplaced the `manifest_writer.complete(...)` call relative to the guarded block, a run whose report write failed would sit at `running` forever and no test would notice. The repo's house rule is explicit: "Test the failure branch you documented." A test asserting the manifest lands `completed` with no report path after a save failure is a handful of lines against the existing helpers in `tests/simulation/test_composite_and_manifest.py`.
- [tests] `src/evaluatorq/simulation/api.py:1702` — `_publish_run`'s executive-summary guard promises in its own comment that simulations "remain useful ... if summary generation itself fails", and no test makes `populate_run_executive_summary` raise anything other than the credentials error already swallowed a layer below. The four genuinely hard lifecycle branches are well covered — the `on_confirm` decline, the `SimulationDroppedError` partial-results path, an `on_run_complete` that raises while a primary error is in flight, and an `on_stage_end` that raises — which is what makes "5470 passed, unchanged" meaningful evidence at all. These two are the gaps in an otherwise well-pinned surface.
- [minimal] `src/evaluatorq/simulation/api.py:1512-1517` — `_resolve_max_turns` is a single expression with complexity 1 and exactly one call site, and the extraction is a net line increase once the signature and annotations are counted. It reads as a helper that exists to move branches off `_simulate_core`'s complexity score rather than to name a concept. Inlining it back is the minimal answer, but it costs roughly two points of complexity on `_simulate_core`, which is the number this PR exists to bring down — so this is a judgement call for the author rather than a clear defect.
- [types, minimal] `src/evaluatorq/simulation/api.py:1499, 1506` — `_RunSettings.evaluator_names` duplicates `run_meta['evaluator_names']` within the same object, with no invariant tying the two together; unlike `target_label` it does have a reader, so this is a readability cost rather than dead state. Separately, `_ExecutionOutcome` is single-writer, single-reader and immediately destructured, and `frozen=True` on a dataclass holding a mutable `list` buys no real immutability — though named fields still read better than a bare tuple, so this is not worth churning.

## Conflicting verdicts

None. The `comments` lens rated two of its findings Critical; they are recorded as Important above, because a comment whose spatial reference has gone stale misleads a reader without changing what the program does, and this skill reserves Critical for things that must not merge.

## Next

1. Delete `_RunSettings.target_label`, narrow `target_model` to `str | None` at its point of construction, and collapse the duplicated scalar parameters on `_resolve_run_settings`, `_publish_run` and `_persist_run` so `config` is the single source of truth. These three are the whole of the structural debt the extraction introduced and they are all cheap.
2. Repoint the five stale comments at the functions that now own their subject, and give the five new helpers docstrings — `_execute_simulation` first, since it holds the contract.
3. Add the manifest assertion to the save-failure test so `_persist_run`'s terminal-write guarantee is pinned by something other than inspection.

Baukebrenninkmeijer:
@Baukebrenninkmeijer

## Questions for the author

I fixed every Important finding I was confident about and pushed them to this branch. Four judgement calls I deliberately did not make on your behalf, because each one trades off against something only you can weigh.

1. **`_resolve_max_turns` is an extraction with no conceptual weight — inline it, or keep it?** The minimality lens flagged it at high confidence: a single expression, cyclomatic complexity 1, exactly one call site, and a net line increase once the signature and type annotations are counted. It reads as a helper that exists to move branches off `_simulate_core`'s complexity score rather than to name a concept. The catch is that inlining it costs roughly two points of complexity on `_simulate_core`, which is the number this PR exists to bring down, and `_simulate_core` currently sits at 8 against a threshold that leaves little headroom. Options: (a) leave it as is; (b) inline it and re-measure the C901 score. My recommendation is (a) — the helper is harmless, it is genuinely testable in isolation, and paying complexity headroom on the function this PR just rescued is a bad trade. I left it alone.

2. **`max_turns` is still passed to `_resolve_run_settings` separately from `config`, and I want you to know why.** I collapsed `model`, `datapoint_parallelism` and `evaluation_name` into reads off `config`, but not `max_turns`. The reason is purely type-level: `config.max_turns` is typed `int | None`, and `SimulationRunMeta['max_turns']` wants a plain `int`. By the time the call happens, `config` has already been rebound via `model_copy` so the two values are always equal at runtime — the separate parameter exists only to carry the narrowed type. Options: (a) leave the parameter, with the docstring explaining it, which is what I did; (b) introduce a resolved-config type where `max_turns` is non-optional, so the narrowing is expressed once in the type system rather than in a parameter. Option (b) is the genuinely structural answer but it is a change to `SimulationConfig`'s shape and reaches well outside this PR's diff. My recommendation is (a) now, (b) as a follow-up if the same narrowing shows up a third time.

3. **`_RunSettings.evaluator_names` duplicates `run_meta['evaluator_names']` inside the same object.** I deleted `target_label` because nothing read it, but `evaluator_names` does have a reader, so the same argument only half applies: it is a readability cost, not dead state. The type lens wanted it dropped in favour of reading `run_settings.run_meta['evaluator_names']` at the single call site. I disagree slightly — `run_meta` is a hook payload whose keys are display-oriented, and reaching into it for a value `build_simulation_run` needs couples the two for no gain. Options: (a) leave the field; (b) drop it and read the TypedDict key. My recommendation is (a), and I left it. Worth overruling me if you think of `run_meta` as the canonical record rather than as a payload.

4. **Two failure branches in the new helpers are documented but not tested, and I only closed one of them.** I added an assertion to `tests/simulation/test_run_store.py::test_simulate_save_failure_still_returns_results` pinning that `_persist_run` still marks the manifest `completed` with no report path when the report write blows up — that one was cheap and it guards a real "stuck at running forever" failure. The other gap I left: `_publish_run`'s executive-summary guard promises in its own comment that simulations "remain useful ... if summary generation itself fails", and no test ever makes `populate_run_executive_summary` raise anything other than the credentials error that is already swallowed a layer below. Both gaps predate this PR; the refactor just moved the code into a place where the promise is now written in a docstring. Options: (a) merge as is and file a follow-up; (b) add the executive-summary failure test here. My recommendation is (a), because the branch is unchanged by this PR and the house rule that motivates the test ("test the failure branch you documented") bites hardest on branches you wrote, not ones you relocated. Say the word and I will add it here instead.

Baukebrenninkmeijer:
Decision for tracking: `_TurnOutcome` keeps two optional fields rather than becoming a tagged union. Cosmetic, deferred until something needs to branch on the variant.

PR #208, opened 2026-09-16
refactor: split evaluatorq() into params, dataset and finish phases (C901 bucket 2)

Baukebrenninkmeijer:
`evaluatorq()` is the package's public evaluation entry point and was the worst single function in `src/evaluatorq/evaluatorq.py` at a ruff `C901` score of 32 against a limit of 15. This splits it along the phase boundaries the source already marked with comments, without changing behaviour, and takes the file off the `C901` debt register in `pyproject.toml`.

Two commits. The first hoists the two nested evaluation coroutines to module level — they were the bulk of the score, since a nested `def`'s branches count against its enclosing function. The second extracts the five remaining phases. Nothing moved into or out of the tracing session, the `AsyncExitStack`, `llm_concurrency_limit`, or `with_progress`; `_enter_single_trace` enters the run span on the caller's exit stack so the span's lifetime is unchanged. No log line, error message, span name or span attribute was touched, and no test assertion changed.

| Function | Before | After |
|---|---:|---:|
| `evaluatorq` | 32 | 3 |
| `_run_streaming_evaluation` (hoisted from `run_streaming_evaluation`) | — | 13 |
| `_normalise_params` | — | 5 |
| `_resolve_streaming_inputs` | — | 4 |
| `_finish_evaluation` | — | 4 |
| `_resolve_experiment_input` | — | 3 |
| `_run_in_memory_evaluation` (hoisted from `run_evaluation`) | — | 3 |
| `_enter_single_trace` | — | 2 |

The file's highest remaining score is `_run_streaming_evaluation` at 13 and `check_pass_failures` at 10, both pre-existing and both under the limit, so `"src/evaluatorq/evaluatorq.py" = ["C901"]` comes out of `[tool.ruff.lint.per-file-ignores]`. That is the only line this PR changes in `pyproject.toml`, which is shared with three sibling C901 PRs.

Checks run, all green:

```
uv run ruff check src/evaluatorq/evaluatorq.py --select C901 --config "lint.per-file-ignores={}"
uv run ruff check src
uv run basedpyright
```

The unit suite reports `5470 passed, 2 skipped, 10 deselected` both before and after the refactor, and `5471` with the regression test added below. `basedpyright` is clean across the whole repo.

## Early-return audit

Two sibling C901 splits hit the same defect: a value the original code had already recorded by the time control left a path stopped being recorded there, because a `break`/fall-through became an early `return`, or because a helper started returning a value the caller then appended only conditionally. I audited this split for it — 34 exit paths across all seven new functions (every `return`, every `raise`, every fall-off-the-end, and every exception that can propagate out of an `await`), each checked against `git show Baukebrenninkmeijer/c901-d:src/evaluatorq/evaluatorq.py` for what the original had written to a list, dict, counter, progress service, span or caller-visible local before that point. **Nothing is lost or moved.** The streaming coroutine's four exits still perform the same `total_datapoints[0] +=`, `processing_tasks.append`, `cancelled_for_fetch.add`, `progress_ref` increments, both `safe_update_progress` calls and `all_results.extend` on exactly the paths that performed them before, and its `finally` block is intact. `_normalise_params` still raises before `tracing_session(...)` opens, so a rejected call still produces no trace. `_finish_evaluation`'s sink append still mutates the caller's list — the frozen dataclass freezes the field binding, not the list object — under both unchanged guards. `validated`, `run_span` and `upload_response` became helper-locals and are provably never read in `evaluatorq()` afterwards.

The audit did turn up a **test** gap, now closed by the third commit. `dataset_id` is resolved inside the streaming branch and read after it, at upload time, making it the one value in `evaluatorq()` that crosses a branch boundary — and it had no coverage at all, because every streaming test runs with `_send_results=False` and every `send_results` test passes `dataset_id=None`. The `_experiment_url_out` sink had the same gap on the streaming path. A change that stopped propagating either would have shipped silently. The new test fails if either write is dropped; I verified that by injecting both mutations and confirming red, then restoring and confirming green.

Two rulings worth flagging for the reviewer:

The streaming phase takes a frozen `_StreamingInputs` dataclass rather than eight positional parameters, and the params phase takes `_EvaluationInputs` and returns `_ResolvedEvaluationInputs` for the same reason. That follows the plan's rule for phases needing more than about six inputs. The cost if it is the wrong call is verbosity, not behaviour — the alternative was an eight-argument call site.

The new `TYPE_CHECKING` import of `TracingContext` is relative (`from .tracing.context import ...`), matching the ten relative imports already at the top of this file and the relative import already inside `evaluatorq()`'s single-trace branch. `CONTRIBUTING.md` asks for absolute imports; converting only the new line would make the file internally inconsistent, and converting all eleven is drive-by reformatting in a PR whose whole point is that nothing else changed. Say the word and I will convert the file's imports in a separate commit.

Based on `Baukebrenninkmeijer/c901-d` rather than `Baukebrenninkmeijer/cicd-complexity-checks`, because that PR also edits this file and stacking avoids the conflict. Merge it first.


currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 5/5** — this is the public library entry point, treated it accordingly, holds up

### ✅ Public signature verified byte-identical
Since `evaluatorq()` is the library's main API, I diffed its full signature (AST-compared, not eyeballed) — every parameter, default, and the return type annotation are unchanged. No breaking change hiding inside this refactor.

### ✅ 13 differential calls, byte-identical output
Ran real calls covering kwargs/dict/EvaluatorParams input forms, with/without evaluators, single_trace, inference=False, job-raises, both ValueError paths, both no-API-key paths, and the deprecated `parallelism` alias. `diff base.out head.out` → identical.

### ✅ Phase ordering and error scope both preserved exactly
`resolve_datapoint_parallelism` still runs before `_normalise_params`, so the deprecated `parallelism` alias still feeds validated params correctly. The only try/except/finally in the original (inside the streaming branch) moved verbatim into `_run_streaming_evaluation`, including the cancellation bookkeeping. `_finish_evaluation` gets exactly the same 11 values it read from closure before, now via an explicit `_EvaluationFinishInputs`.

### ✅ ruff C901 confirmed
32 → 15 on `evaluatorq()`, and the file-level ignore is correctly removed from pyproject.toml.

🟡 Nit only: `_normalise_params` rebinds several fields to locals that get immediately overwritten by the validated extraction — dead assignments, harmless but slightly misleading to read.

Test suite identical on base and head (5454 passed); the one live-catalogue-drift failure is pre-existing across this whole stack.

Baukebrenninkmeijer:
# PR Review — #208 (`gh pr diff 208`, base `Baukebrenninkmeijer/c901-d`, head `222bd765`)

Lenses run: code quality, minimality, tests, error handling, type design, comments. Skipped: spec — no ticket id resolves from the branch name, the commit messages or the PR title, so the PR body served as the statement of intent and the code-quality lens judged against it directly. Failed: none.

Repo checks first, all green on the head commit: `uv run ruff check src`, `uv run ruff format --check src`, `uv run basedpyright` (whole repo, unscoped) and `uv run pytest tests/unit/test_evaluatorq.py` all pass.

The headline verdict is good news: two lenses independently diffed every extracted helper line-by-line against `Baukebrenninkmeijer/c901-d` and both concluded the split is faithful. Error precedence in the streaming coroutine, the `finally` cancellation sequence, the `cancelled_for_fetch` bookkeeping, the `ValueError` that fires before `tracing_session` opens, and the `_experiment_url_out` append all behave exactly as they did. The early-return audit in the PR body holds up. Everything below is about the shape of the new code, not about behaviour it changed.

## Critical (0)

None.

## Important (6)

- [minimality ⇄ type design] `src/evaluatorq/evaluatorq.py:211-221, 369-379, 404-411, 680-689` — Three of the new helpers open by copying every field of their input dataclass into a same-named local, and the call site in `evaluatorq()` then copies every field of the returned dataclass back into the function's own parameter names. That is roughly forty lines that do nothing but rename `inputs.x` to `x`. Worse, re-binding the parameter names throws away the narrowing the resolved dataclass just bought: `resolved.jobs` is `list[Job]`, but the parameter `jobs` is declared `list[Job] | None`, so assigning one to the other widens it straight back. Structural fix: delete all four prologues and read `inputs.x` / `resolved.x` at each use site. Cost: a handful of slightly longer expressions inside the helpers, and one new local name for the post-experiment data so the narrowed value is not stored back into the wide parameter. Convergent across two lenses.

- [type design] `src/evaluatorq/evaluatorq.py:252, 258` — `_normalise_params` needs `cast('list[Job]', ...)` twice, and the second one carries no comment saying which invariant it leans on. The casts exist only because the local `jobs` was first bound from the wider `list[Job] | None`, so basedpyright pins it to that type across the if/else. A cast is an unenforced claim; a future edit that leaves `jobs` as `None` on either branch would type-check silently. Structural fix: pull the branch into a small `_resolve_jobs(validated) -> list[Job]` helper whose every branch returns, so the narrow type is inferred rather than asserted, and both casts disappear. Cost: about ten lines for the helper.

- [type design] `src/evaluatorq/evaluatorq.py:308-312` — `_resolve_experiment_input` takes and returns `... | None`, but the `None` is unreachable: the only caller passes `resolved.data`, which `_normalise_params` already guarantees is non-`None`. A return type wider than the real postcondition pushes work downstream — the `else` branch at line 733 has to `cast('list[DataPoint]', data)` partly to erase a case that cannot happen. Structural fix: drop `| None` from both the parameter and the return, and restructure the body so every branch returns, which makes the narrowed return type inferable instead of asserted. (The type-design lens filed this as Critical; I am downgrading it to Important because there is no runtime consequence today — it is a type-level defect, not a live bug.)

- [comments] `src/evaluatorq/evaluatorq.py:313-315` — Two comments moved into `_resolve_experiment_input` and stopped being true there. "then fall through to the in-memory data path below" describes the old inline body of `evaluatorq()`; inside this helper there is no "below", the function just returns. "The validator already guarantees inference=False here" cites a value the helper cannot see, since its signature takes no `inference` parameter at all. Structural fix: reword both to describe the helper's actual contract — the caller treats the result as in-memory data, and `inference=False` is a precondition the caller has established, not something this function checks.

- [comments] `src/evaluatorq/evaluatorq.py` — All seven new helpers and all four new dataclasses ship with no docstring. Two of them carry real contracts that used to be legible inline and now have no doc surface anywhere: `_resolve_streaming_inputs` raises `ValueError` on a falsy `ORQ_API_KEY`, and `_run_streaming_evaluation` has a specific error-precedence rule (a caller `CancelledError` wins and discards collected task errors; two or more errors wrap in `_StreamingEvaluationError`; exactly one is re-raised bare). `_ResolvedEvaluationInputs.jobs` is a third: it is typed plain `list[Job]` but is the synthetic `[_replay_recorded_response]` placeholder whenever `inference=False`, which is invisible from the type. Structural fix: a one-to-three-line docstring per helper stating the preconditions the caller must already have established and what the helper raises or returns.

- [tests] `src/evaluatorq/evaluatorq.py:340` — `_resolve_streaming_inputs`'s `ValueError` for a missing `ORQ_API_KEY` on a `DatasetIdInput` run has no test anywhere in the repo. `grep -rn "must be set to fetch datapoints" tests/ src/` matches only the raise site. The sibling `ExperimentInput` branch is covered by `tests/unit/test_no_inference_experiment.py::test_evaluatorq_experiment_requires_api_key`; the streaming branch is not. The refactor pulled this check into its own function precisely so it could be tested in isolation, and nothing does. Structural fix: mirror the existing experiment test with `monkeypatch.delenv('ORQ_API_KEY')` and `pytest.raises(ValueError, match='ORQ_API_KEY')`.

## Suggestions (7)

- [minimality ⇄ type design] `_EvaluationInputs` and `_ResolvedEvaluationInputs` carry nine of the same field names as the existing `EvaluatorParams` pydantic model, so the same data now has three parallel shapes in one file. Collapsing them (return the `EvaluatorParams` instance plus the two genuinely re-derived values) would remove around fifty-five lines. This is a real design call rather than a defect, so it is in the questions comment rather than fixed here.

- [type design] `_resolve_streaming_inputs` returns a bare `tuple['Orq', str, bool]` while every sibling phase got a named dataclass. The blast radius of swapping two same-shaped elements is one line, so this is low-stakes, but it is the one holdout from the pattern the rest of the PR establishes.

- [type design] `_EvaluationFinishInputs` and `_StreamingInputs` are `frozen=True` over payloads their own helpers mutate — `experiment_url_out` is a sink that `_finish_evaluation` appends to, `results` is a list, and `tracing_context.parent_context` is reassigned by `_enter_single_trace`. `frozen` freezes the binding, not the object, so a reader who takes `frozen=True` as "safe to share across concurrent tasks" would be wrong for exactly the fields that matter most.

- [tests] The `include_messages` override when `inference=False` (and its `logger.debug`) is not exercised through `evaluatorq()` by any test. This is a pre-existing gap the refactor merely made easier to isolate, not a regression this diff introduces.

- [tests] `_normalise_params`'s catch-all `ValueError` ("Either 'params' or both 'data' and 'jobs' ...") has no direct test. `test_inference_true_without_jobs_raises` exercises a different `ValueError` raised inside `EvaluatorParams` validation. Lowest-value of the three coverage gaps, since it only guards obviously-wrong call sites.

- [minimality] The new test's eight-line docstring is a design essay relative to its two assertions. Its claims were checked and are accurate — it is verbose, not wrong.

- [type design] The file has no `from __future__ import annotations`, so the `TYPE_CHECKING`-gated `Orq` and `TracingContext` annotations must stay string-quoted, and they all correctly are. Plain `@dataclass` does not call `get_type_hints()`, so nothing breaks at import or construction time. Flagged only as a maintenance trap: dropping a quote later, without adding the future import first, would raise `NameError` at class-definition time.

## Conflicting verdicts

- `tests/unit/test_evaluatorq.py:813-867` — the minimality lens called the new test's docstring disproportionate; the comments lens verified its claims and found no overclaim. These are not actually in conflict: both agree the content is accurate, they disagree only on whether the length is worth it. No call needed.

## Next

1. Delete the four unpack-to-locals prologues and read the dataclass fields directly, so the resolved types stop being widened back at the call site.
2. Extract `_resolve_jobs` to remove both `cast('list[Job]', ...)` calls.
3. Narrow `_resolve_experiment_input` to drop the unreachable `| None`.
4. Fix the two orphaned comments and add docstrings to the helpers with hidden contracts.
5. Add the missing `ORQ_API_KEY` test for the streaming branch.

Baukebrenninkmeijer:
## Questions for the author

@Baukebrenninkmeijer — three things from the review I deliberately did not decide for you. Everything else from the report is fixed in `e00bfe3d`.

1. **Should `_EvaluationInputs` and `_ResolvedEvaluationInputs` exist at all?** Nine of their ten field names are already field names on the `EvaluatorParams` pydantic model, so the same data now has three parallel shapes in one file, and `_normalise_params` builds an `EvaluatorParams` in the middle just to take it apart again. The options are (a) keep both dataclasses as they are, which is what is on the branch and what your PR body argues for on the grounds that the alternative was an eight-argument call site; (b) delete both and have `_normalise_params` take the same keyword arguments `evaluatorq()` already has and return the `EvaluatorParams` instance plus the two values that genuinely get re-derived, `jobs` and `evaluators_list`, which removes roughly fifty-five lines; (c) keep `_EvaluationInputs` as the call-site bag and drop `_ResolvedEvaluationInputs` in favour of returning `EvaluatorParams` plus the two derived values. My recommendation is (c): the input bag really does spare you an eleven-argument call, but the resolved bag is close to a hand-written copy of a model you already own, and keeping it means every future field added to `EvaluatorParams` has to be added in two more places. I left it alone because this is a shape decision with no defect behind it, and because the PR's whole premise is that nothing else changed.

2. **Should `_resolve_streaming_inputs` return a named type instead of a bare `tuple['Orq', str, bool]`?** Every other phase in this PR got a named dataclass specifically so its fields read as `inputs.dataset_id` rather than as a position. This one is the holdout, and its three elements are destructured on a single line by its only caller, so the blast radius of swapping two of them is one statement. The options are to add a fourth small frozen dataclass for consistency, or to leave the tuple as a deliberate exception for something this local. My recommendation is to leave it: a three-element tuple destructured immediately at the one call site is legible, and adding a class purely for symmetry adds a type nobody reads. I am flagging it only because a reviewer who noticed the pattern will ask.

3. **Is the `frozen=True` on the two dataclasses that hold mutable payloads worth keeping?** `_EvaluationFinishInputs.experiment_url_out` is a sink that `_finish_evaluation` appends to, `results` is the caller's list, and `_StreamingInputs.tracing_context` has its `parent_context` reassigned by `_enter_single_trace`. `frozen` freezes the bindings, not the objects, so a reader who takes it as "safe to share across concurrent tasks" would be wrong for exactly the fields that matter most. The options are to drop `frozen=True` from those two, since it is not buying the immutability it appears to claim, or to keep it for the binding safety and rely on documentation. I took the documentation route for now — both classes gained a docstring saying in as many words that the freeze covers bindings only and naming the fields that are written through — because dropping the keyword is a real semantic change to types you introduced in this PR and I would rather you make that call. If you want them unfrozen, it is a two-line change.

Baukebrenninkmeijer:
GitHub grouped this PR into a native stack whose base branch was merged and deleted, which leaves its mergeability stuck at unknown and blocks retargeting. Same branch and commits, reopened as a fresh PR against c901-d; link in the next comment.

Baukebrenninkmeijer:
Reopened as #224.

PR #209, opened 2026-09-16
refactor(redteam): split run_attack's turn loop into attacker, marker and target phases (C901 bucket 2)

Baukebrenninkmeijer:
`run_attack` in `redteam/adaptive/orchestrator.py` scored 32 against ruff's `C901` limit of 15, and the file sat on the per-file debt register in `pyproject.toml`. It is a 555-line method whose turn loop interleaved adversarial generation, content-filter retries, objective-marker parsing and the target call. This splits it along those seams and clears the register entry. Nothing about what it does changes.

| Function | Before | After |
|---|---|---|
| `run_attack` | 32 | 13 |
| `_attacker_turn` (new) | — | 8 |
| `_objective_marker_phase` (new) | — | 6 |
| `_adversarial_generation` (new) | — | 3 |
| `_target_turn` (new) | — | 2 |
| `_record_target_failure` (new) | — | 2 |
| `_unresolved_timeout_error` (new) | — | 3 |
| `_start_attack_progress`, `_accumulate_target_usage`, `_target_attempt_span`, `_record_attempt_response` (new, module level) | — | ≤ 3 each |

Three commits. The first hoists the three nested function definitions out of the turn loop and pulls the progress-start block, the target usage accumulation and the unresolved-timeout tail into helpers (32 → 23). The second extracts the three large phases, each returning a frozen dataclass rather than filling shared state — `_AttackerTurn` carries the error fields that drive the loop's `break` and the flag that drives its `continue`, `_MarkerPhase` carries the stripped prompt and the objective state (23 → 13). The third is a fix described below.

Every `await`, `try`, `async with`, `logger.*`, `set_span_attrs` and progress-hook call appears exactly once, in the same relative order; both reviewers produced that as an explicit checklist. No call moved into or out of `regenerate_on_content_filter` or `call_target_with_retry`. The prompt-cache render moved into `_attacker_turn` but is still computed per turn against the live transcript with `volatile_tail=0`, and the marked copy is still never written back.

## Two behaviour regressions the review caught

Both were introduced by the split and both are fixed on the branch; they are worth naming because they are the shape of bug this kind of refactor produces.

The first: `truncation_warnings.append(turn + 1)` ended up after the early `break` instead of before it. A turn whose attacker response was cut off at `max_tokens` and came back empty would have silently dropped out of `OrchestratorResult.truncated_turns`. The append now runs before the break, and `_attacker_turn` reports `truncated=False` on exactly the paths that returned before the original's truncation check.

The second: one attempt's token usage was folded into the accumulator *after* `response.choices[0].message.content` was read rather than before it. An empty `choices` list raises `IndexError` there, and the original had already counted the billed attempt by that point while the split would have thrown it away. `_adversarial_generation` now takes a `record_usage` callback and calls it at the original position, inside both spans.

## Rulings

The plan's seam list named `_record_turn` and `_should_stop`. Neither exists as a separable block in the current source — the `Turn` append is a one-liner appearing twice, and the early-exit checks are spread across four `break` sites that each set different error locals. The seams here are drawn on the actual control flow instead.

The reviewer asked for three re-wrapped lines inside moved bodies to be restored. Two are what `ruff format` produces at the reduced indentation, so restoring them would fail `ruff format --check src`. The third needed a `# fmt: skip` pragma to hold, which is more noise than the rewrap it protects; the rendered log message is byte-identical either way. All three are left as the formatter wrote them.

## Checks

```
uv run ruff check src                  # All checks passed!
uv run ruff format --check src         # 233 files already formatted
uv run basedpyright                    # 0 errors, 0 warnings, 0 notes
uv run pytest -m 'not integration'     # 5470 passed, 2 skipped, 10 deselected
uv run ruff check src/evaluatorq/redteam/adaptive/orchestrator.py --select C901 --config "lint.per-file-ignores={}"
                                       # All checks passed!
```

The test count is identical to the branch point. No test file is touched.

`pyproject.toml` loses one line — the `redteam/adaptive/orchestrator.py` C901 entry. The other entries belong to the three sibling C901 branches and are untouched.


currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 5/5** — verified against 14 real multi-turn scenarios, not just read

### ✅ History and turn state thread correctly
`_target_turn` derives the transcript from the live turns list by reference before each call — confirmed turn 3 sees the full accumulated history, not a stale snapshot. `adversarial_messages` accumulation stays inline in `run_attack` across the split, so cache-breakpoint logic still recomputes against the live transcript per turn.

### ✅ Marker sees exactly what it saw before
The objective-marker phase parses only the attacker's output, same as base — it never saw the target's response in either version, so there's no turn-misalignment risk to introduce here in the first place.

### ✅ All termination and error-scope points unchanged
Every break/continue/raise sits at the same relative position as base — error-fields break, retry-continue, marker-stop break, target-failure break, and the objective-achieved check still evaluated outside the span after the turn body. No try/except was widened or narrowed; the attacker's own timeout/exception handling stays local to `_attacker_turn`, nothing gets swallowed into an end-of-loop catch-all.

### ✅ 14 differential scenarios, byte-identical
Clean multi-turn, marker-with-residual-prompt, rejected turn-0 marker, target exception/timeout, attacker truncation (empty and non-empty), content-filter, exception, timeout-with-recovery, trailing unresolved timeout, all-turns-timeout, empty-choices response. Full JSON diff of every attacker/target message, error field, and usage counter across all 14: identical.

### ✅ ruff C901 confirmed: 32 → 13

No 🔴 needs-live-testing items — every branch including timeouts and content filters is reachable through the mocked client.

Baukebrenninkmeijer:
# PR Review — #209 (`Baukebrenninkmeijer/c901-o` against `Baukebrenninkmeijer/cicd-complexity-checks`, head `fa02beec`)

Lenses run: code quality, silent failures, type design, comments, tests, minimality. Skipped: none. Failed: none.

Repo checks on the head commit before review: `ruff check src` clean, `ruff format --check src` clean, `basedpyright` 0 errors, `pytest -m 'not integration'` 5469 passed with one failure in `tests/redteam/test_retry_wiring.py::test_orchestrator_target_call_has_exactly_outer_retry_attempts`. That test passes in isolation, exercises `call_target_with_retry` directly, and touches no file in this diff — it is a local flake from several suites running concurrently on this machine, not a defect in the PR.

The spec claims check out empirically. `ruff check src/evaluatorq/redteam/adaptive/orchestrator.py --select C901 --config "lint.per-file-ignores={}"` passes, the per-function complexity table in the PR body matches what ruff reports (`run_attack` 13, `_attacker_turn` 8, everything else at or below 6), and `pyproject.toml` loses exactly the one orchestrator entry with no sibling branch's entry touched. Two lenses independently traced the old inline loop against the new helper chain statement by statement and found no divergence in order of operations, exception scope, variable lifetime, break/continue semantics, span attributes, progress-hook calls or logging. Both regressions named in the PR body are genuinely fixed in the final state. The refactor is sound; the findings below are about what the split left unguarded.

## Critical (2)

- **[tests] `tests/redteam/test_orchestrator_coverage.py:789-807` — the truncation-ordering regression would have shipped green.** `test_truncation_warning_tracked` asserts `result.truncated_turns == [1, 3]` but only drives turns with non-empty content, so it never enters the empty-content branch where the bug lived. `test_adversarial_empty_max_tokens` is the only test that drives that exact branch — empty content with `finish_reason='length'` — and it asserts `error_type` and `error_code` and nothing else. I verified this by reintroducing the regression (moving `truncation_warnings.append(turn + 1)` back below the early break) and running the suite: it passes. The fix the PR body describes is real but entirely unpinned. Structural fix: assert `truncated_turns` in the test that already reaches the branch. Cost: one line.

- **[tests] `src/evaluatorq/redteam/adaptive/orchestrator.py:684` — the token-accounting regression is equally unpinned.** Nothing anywhere drives `run_attack` with a completion whose `choices` list is empty. The `choices=[]` tests in `tests/redteam/test_llm_client_threading.py` exercise the OWASP evaluator and scorer, not the orchestrator; `test_content_filter_raised_as_error_is_classified` raises from the mock client rather than returning a 200-shaped response, so it never reaches the `response.choices[0]` read at all. Reverting the fold order would silently drop a billed attempt from `token_usage_adversarial` and report the attack as costing nothing, with no test failing. Same verification: I reintroduced it and the suite stayed green. Structural fix: a test that builds a response with `choices = []`, asserts the clean abort, and asserts the attempt is still counted.

Both of these matter more than usual here. This is a refactor whose own PR body documents that the split introduced two behaviour regressions during development. Both were caught by review rather than by the suite, which is the definition of an untested seam.

## Important (5)

- **[types, comments] `orchestrator.py:410-426` — `_AttackerTurn` states its central invariant in prose and enforces nothing.** ⇄ convergent. The docstring has to say `error_fields` and `retry` are mutually exclusive because the type accepts a value with both set, which would make the caller both break and continue. No branch violates it today and no test would notice if one started to. Structural fix: a tagged union of three outcome types, at the cost of three small classes and a `match` at the call site; the cheaper enforcement is a `__post_init__` that rejects the illegal combination.

- **[types] `orchestrator.py:421` — `error_fields` is a positional 6-tuple where four consecutive members are `str`.** `tuple[str, str, str, str, dict[str, Any], int]` is constructed at four sites and unpacked at one. Transposing `error_type` with `error_stage`, or `error_stage` with `error_code`, produces no type error, no test failure, and a report whose error classification is quietly wrong. This hazard did not exist before the split — those were six separately named locals assigned in place. Structural fix: a frozen `_AttackError` with six named fields, so a transposition becomes a keyword typo.

- **[comments] `orchestrator.py:415-416` — the `_AttackerTurn` docstring is wrong about `truncated`, in the exact direction that caused the regression.** It says `truncated` is "only meaningful when both are absent". The opposite is true: `_attacker_turn` deliberately sets a real `truncated` value on both of its error-returning branches, and `run_attack` reads it before it branches on `error_fields`, precisely so a content-filtered or `max_tokens`-truncated turn is still recorded. A reader who trusts this docstring would conclude the read is misplaced and move it below the break — which is the regression, restored.

- **[comments] `orchestrator.py:710, 1029` — neither `_attacker_turn` nor `_target_turn` has a docstring, and both sit on a retry path.** CLAUDE.md's rule is that a call path picks one retry layer and says which in the docstring. `_attacker_turn` wraps `regenerate_on_content_filter`, `_target_turn` wraps `call_target_with_retry`, and neither states that it adds no second layer. The inline comment inside `_attacker_turn` describes content-filter behaviour but never makes the "only layer" claim. Ten new helpers were added and only the dataclass carries a docstring.

- **[tests] `orchestrator.py:700-708` — neither branch of `_unresolved_timeout_error` is exercised.** Every test in `tests/redteam/test_adversarial_timeout_budget.py` ends by hitting `max_consecutive_adversarial_timeouts`, which is the give-up branch inside `_attacker_turn`, not this one. The scenario this helper exists for — the loop ending naturally while a timeout is still outstanding and under budget — is never reached. Its non-empty-`turns_record` branch logs "the last turn was dropped silently" and then returns no error at all, which is exactly the kind of documented degradation CLAUDE.md requires a test for.

## Suggestions (4)

- **[types, minimality] `orchestrator.py:627-685` — the `record_usage` callback duplicates a value that already crosses the boundary.** ⇄ convergent, but see the conflict below before acting on it. `_adversarial_generation` returns `usage` in its tuple and also hands it to a one-implementation callback whose only job is to mutate the caller's accumulator.

- **[types] `orchestrator.py:429-434` — `_MarkerPhase.stop` is a boolean where the two stop reasons already differ.** The phase computes "objective achieved" versus "rejected turn-0 marker", logs different messages for each, then flattens both to `stop=True` and lets the caller re-derive the distinction from `objective_achieved`. A `stop_reason` naming the outcome would remove the re-derivation.

- **[comments] `orchestrator.py:854` — "finish_reason / unusable_kind were computed in the generation retry loop above" is now stale.** In the old inline loop that comment bridged a real distance. After the move the computation sits four lines up in the same `try` block, and it runs after `regenerate_on_content_filter` returns rather than inside its retry loop, so the sentence is both redundant and inaccurate.

- **[comments] `orchestrator.py:637-640` — the "the shared `regenerate_on_content_filter` helper decides whether to retry" comment sits inside the callee, which has no retry logic.** It reads as though `_adversarial_generation` participates in the retry decision. It belongs where the `regenerate_on_content_filter` call actually is.

The minimality lens judged `_start_attack_progress`, `_target_attempt_span` and `_record_attempt_response` to be genuine single-purpose extractions rather than line-count shims, and found the author's refusal to extract the planned `_record_turn` and `_should_stop` seams justified: the turn append is one statement with different side effects on each side, and the four break sites each carry a different field set, so a boolean predicate would only relocate the same unpacking. Measured on normalised lines, roughly 79 percent of the new file is a verbatim reindent of the old one, which is consistent with a mechanical extraction.

## Conflicting verdicts

- **`orchestrator.py:627-685` — the minimality lens's proposed fix for `record_usage` would reintroduce the regression the third commit fixed.** It recommends deleting the callback and folding usage in `_generate_attack_turn` after the unpack. That cannot work as written: `_adversarial_generation` reads `response.choices[0].message.content` before it returns, so an empty `choices` list raises `IndexError` and the caller never reaches its fold — the billed attempt is dropped, which is precisely regression two. The only safe version of that simplification also moves the `choices[0]` read out to the caller, which in turn moves the billing fold outside the two spans the author deliberately kept it inside. I have not applied it; the trade-off is in the questions comment. The finding that the callback is redundant stands; the proposed remedy does not.

## Next

1. Pin both documented regressions with tests, since the suite currently proves neither.
2. Replace the positional error tuple with a named type and enforce the `_AttackerTurn` invariant instead of describing it.
3. Fix the `truncated` docstring before it talks someone into undoing the fix.

Baukebrenninkmeijer:
## Questions for the author

@Baukebrenninkmeijer — I applied the fixes I was confident about and pushed them. Four things I could not decide for you, because each is a design call or depends on intent you have and I do not.

**1. The `record_usage` callback: keep the indirection, or move the billing fold out of the spans?**

Two lenses independently flagged `_adversarial_generation`'s `record_usage: Callable[[TokenUsage], None]` as redundant — the same `usage` value already crosses the boundary in the return tuple, and the callback has exactly one implementation and one call site. The minimality lens's proposed fix was to delete the callback and fold usage in `_generate_attack_turn` after the unpack.

I did not apply it, because as written it reintroduces the exact regression your third commit fixed. `_adversarial_generation` reads `response.choices[0].message.content` before it returns, so an empty `choices` list raises `IndexError` and the caller never reaches its fold. The billed attempt is dropped.

The options are: (a) keep the callback as it is — I have documented it, and the new test now pins the behaviour it protects; (b) move the `choices[0]` read out to `_attacker_turn` as well, so `_adversarial_generation` returns `(response, usage)` and the caller folds then reads — this removes the callback safely, but it also moves the billing fold outside the two spans, which your PR body says you deliberately kept it inside; (c) leave the callback but accept it as permanent.

My recommendation is (a). The indirection is genuinely a little awkward, but it is the thing making the "every attempt counts, even a malformed one" guarantee hold, and paying one callback for that is a good trade. (b) is defensible if you never cared about the fold being inside the spans — you would know whether anything downstream reads those spans expecting the usage to be recorded within them. I have left it alone rather than guess.

**2. Should `_AttackerTurn` become a tagged union rather than a validated record?**

The type lens rated this Critical: `_AttackerTurn` has three mutually exclusive outcomes — an error that breaks the loop, a retry that continues it, and a success payload — and the type accepts a value with all three set at once. I took the cheaper enforcement: the error field is now a named `_AttackError` object rather than a positional 6-tuple, and `__post_init__` rejects a value that sets both `failure` and `retry`.

The structural alternative is three small frozen dataclasses behind a `Literal`-tagged union, with a `match` at the call site instead of two sequential `if`s. That makes the illegal state unrepresentable rather than merely rejected at runtime, and basedpyright would enforce exhaustiveness. It costs about thirty lines and rewrites the branch you just finished carefully splitting.

My recommendation is to leave it as I have it for this PR and open a follow-up if you want the union. The runtime guard closes the actual risk — a future branch constructing a contradictory value — and doing the union now means re-reviewing the control flow a second time on a branch whose whole value is that the control flow was verified unchanged. But it is your call whether a refactor PR is the right place to also fix the type.

**3. Is `_MarkerPhase.stop` hiding a distinction the caller should see?**

`_objective_marker_phase` computes two different stop reasons — a genuine accepted objective success, and a rejected turn-0 marker that left no usable prompt. It logs a different message and sets a different span `finish_reason` for each, then flattens both into `stop=True`. The caller breaks without knowing which, and anything downstream that wants the distinction re-derives it from `objective_achieved`.

Options: leave it, or replace `stop: bool` with `stop_reason: Literal['achieved', 'no_prompt'] | None`.

I did not change it, because the caller genuinely does not need the reason today and adding a field with no reader would be dead flexibility. Worth a decision only if you know of a consumer that wants it.

**4. `_unresolved_timeout_error` leaves `error_turn` unset — deliberate, or an oversight the refactor inherited?**

When the loop ends with an outstanding timeout and no turns recorded, the result gets `error`, `error_type`, `error_stage`, `error_code` and `error_details`, but `error_turn` stays `None`. Every other error path sets it. The pre-refactor code behaved the same way, so this is not a regression and I preserved it exactly — I also added a test that asserts `error_turn is None` there, which pins the current behaviour either way.

If it was always intended (there is no turn to point at, so `None` is honest), nothing to do and the test now says so. If it was an oversight, `error_turn = 1` or the count of attempted turns would make the reports consistent, and the assertion I added should flip in the same change. I could not tell which from the code, and the test I wrote would need editing under the second reading, so I would rather you say than have me guess and bake a guess into an assertion.

Baukebrenninkmeijer:
Two reviewer commits (`26f7fe1c`, `dd447749`) were dropped from this PR by a stale rebase before it merged; #225 lands them on top of main.

PR #211, opened 2026-09-16
refactor(redteam): split _run_dynamic_or_hybrid into its seven phases (C901 bucket 2)

Baukebrenninkmeijer:
`_run_dynamic_or_hybrid` was a 1079-line orchestration scoring 108 on ruff `C901` against a limit of 15. This splits it along the `# Step N` comments the source already carried, one private module-level phase per step, with frozen dataclasses where a phase needed more than a handful of inputs. Nothing about the run changes: same hook order and payloads, same spans and span attributes, same log lines, same error text, same public signature and docstring.

It landed as three commits, one per phase group, each reviewed against the constraint that every `await`, `try`, `with`, hook call and `logger.*` line in the original appears exactly once in the new code, in the same relative order.

## Scores

| Function | Before | After |
|---|---|---|
| `_run_dynamic_or_hybrid` | 108 | under 15 (no longer listed) |
| `red_team` | 57 | 57 — unchanged, PR 2 in this stack owns it |
| `_run_static` | 30 | 30 — unchanged, PR 3 in this stack owns it |

Every function extracted here is at or under 15; none of them appears in the register. `redteam/runner.py` stays in `pyproject.toml`'s `per-file-ignores` because two of its functions are still over the limit — PR 3 removes the entry.

## Phases extracted

Step 1 and 1b became `_retrieve_agent_contexts`, `_resolve_capability_llm_client` and `_classify_capabilities` (returning a frozen `_ClassifiedCapabilities`); the nested `_ctx_end_meta` was hoisted to module level while the two `on_stage_end` emit loops stayed inline, because they carry hook ordering. Step 2 became `_estimate_dynamic_datapoints`, `_cap_strategy_breakdown`, `_estimate_static_datapoints` and `_build_confirm_payload`, with the `on_confirm` call and its `CancelledError` left at the top level. Steps 3 through 4b became `_prepare_string_targets`, `_generate_agent_target_datapoints` and `_build_agent_target_jobs`. Steps 5 through 7 became `_build_redteam_evaluators`, `_execute_attacks`, `_group_results_by_target`, `_convert_target_report`, `_collect_filter_warnings`, `_finalize_merged_report` and `_cleanup_target_memory`.

## Rulings

Three judgement calls are worth naming, since each could have been made differently.

The `all_at_cleanup_info` registration, the `at_mem_ids` list creation and the loop-carried `shared_at_dps` assignment all stayed in the caller rather than moving into the extracted agent-target phase. The cancellation cleanup path dedupes on `id(pt.memory_entity_ids)`, so the list object's identity — not just its contents — has to survive the new function boundary. Keeping the creation and registration in the caller makes that unconditional instead of something a future edit could quietly break.

`_collect_filter_warnings` and `_cleanup_target_memory` return their warning strings for the caller to extend onto `merged.pipeline_warnings`, where the original appended to that list inside the loop, interleaved with `CLEANUP` hook calls. No hook and no other reader observes `merged` between those points, so the final list content and order are identical. If an exception were raised mid-cleanup the collected warnings would be dropped, but `merged` is only returned on the success path, so an exception discards it either way.

`from evaluatorq import evaluatorq` stays a function-local import inside `_execute_attacks`. Hoisting it to module level would be a circular import.

## Checks run

```
uv run ruff check src
uv run basedpyright
```

All green. The suite is 5470 passed, 2 skipped, 10 deselected — identical to the baseline measured on the base branch before any change, after every one of the three commits.

Beyond the per-commit reviews, the whole branch was audited once for a bug class a sibling refactor hit: a loop body becoming a function turns a `break` or `continue` into an early `return`, and a write the original had already made by that point — an accumulator append, a counter, a dict mutation — silently stops happening. The audit found no lost writes. No `break` or `continue` became a return here, and `classification_errors`, `strategy_breakdown`, `all_at_cleanup_info`, `at_mem_ids`, the `hybrid_source` datapoint tagging and `shared_at_dps` all keep object identity across the new boundaries.


currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 5/5** — the worst offender in the codebase (complexity 108), split and independently verified as behavior-identical

### The title undersells it
"Seven phases" undercounts — this is 19 new top-level helpers plus 10 frozen input dataclasses. Worth knowing when reading the diff, since it's a bigger restructure than the title implies, but a bigger restructure isn't a red flag on its own if it's verified — see below.

### ✅ No closure state lost
Checked every one of the 119 fields across the 10 `_*Inputs` dataclasses — all read by their consumer, none dead. The two genuinely closure-captured values (`classification_errors`, `cap_llm_client is not None`) are correctly threaded through an explicit return object rather than silently dropped.

### ✅ Branching preserved almost line-for-line
Statement-level diff of the whole replaced region: exactly one control-flow line changed, and it's a pure reflow (multi-line condition to one line, same operands, no logic change). All 6 hybrid-mode checks, both raise sites, all breaks/continues survive at the same decision points.

### ✅ All 7 exception seams verified 1:1
Every try-block in the original has a matching one at head with the same span. The two highest-risk ones (cancellation handling around attack execution, and the finally-block around report saving) are byte-identical when diffed directly against base.

### ✅ 13 differential scenarios, byte-identical
Dynamic/hybrid modes, capped datapoints, multi-category, multi-target, memory cleanup, confirm-rejected in both modes, mid-attack failures, capability-classifier raising — full hook-event sequence, confirm payload, and report canon compared. Identical across all 13, no live LLM needed.

### ✅ ruff C901: 108 → 14, every new helper under 15
And the "real diff" gut check holds up: `git diff -w` shows +984/-340, of which 421 lines are mechanical (dataclass blocks, unpack lines, signatures, docstrings) and the rest is keyword-expanded call sites. Zero added logic beyond the 19 returns.

### 🟡 One thing to close out in the next bucket
pyproject.toml:197 still blanket-exempts the whole file (`BLE001, C901`) because `red_team` (57) and `_run_static` (30) are still over the limit. This PR's win isn't enforced yet — a new 40-complexity function added to this file today would merge silently. Not this PR's job to fix, but flag it so it doesn't get forgotten once the remaining two functions are split.

Baukebrenninkmeijer:
# PR Review — `gh pr diff 211` (`Baukebrenninkmeijer/c901-r1` vs `Baukebrenninkmeijer/cicd-complexity-checks`)

Lenses run: equivalence (behaviour preservation), code quality, error handling and silent failures, type design, comments and docstrings, test coverage, minimality. Skipped: spec, because the PR body *is* the spec here and the equivalence lens covers it directly. Failed: none.

## Verification of the PR's own claims

Both headline claims hold, and I want to state that before the findings, because in a refactor this size it is the thing that matters most.

**The C901 goal is met.** Running ruff with `--isolated --select C901 --config 'lint.mccabe.max-complexity=15'` against the pre- and post-change files: `_run_dynamic_or_hybrid` is listed in the old file and absent from the new one, and none of the newly extracted functions appears either. The only remaining entries are `red_team` (57) and `_run_static` (30), which this PR does not touch, exactly as described.

**The extraction is behaviour-preserving.** I verified this mechanically rather than by reading, in three independent ways.

1. Token counts over the old function body versus the union of the new caller and its phases are identical: `try:` 7/7, `except` 7/7, `finally:` 2/2, `with ` 14/14, `raise ` 7/7, `logger.` 8/8 with an identical message multiset, `on_stage_start` 5/5, `on_stage_end` 9/9, `on_confirm` 1/1, `CancelledError` 3/3, `asyncio.gather` 2/2, `await_maybe(` 23/23, `with_redteam_span(` 11/11, `set_span_attrs(` 11/11, `PipelineStage.` 20/20, `_check_filter_results(` 5/5. The six extra `await`s are exactly the six new phase-function calls.
2. A multiset diff of every normalised source line (comments, blank lines and the `x = inputs.x` unpack boilerplate stripped, `inputs.` prefixes removed) leaves only fifteen lines present in the old version and absent from the new. Every one is accounted for by a signature change, a line-wrap collapse, or the `merged.pipeline_warnings.append(...)` calls becoming `filter_warnings.append(...)` / `cleanup_warnings.append(...)` inside the two collector helpers. No statement was lost, added, or altered.
3. An execution-order comparison: I re-inlined every phase call back into the caller programmatically and diffed the resulting ordered sequence of 76 side-effecting statements (awaits, raises, hook emissions, span opens, logger calls, `_save_stage` / `_save_report` / `_send_cleaned_results` / `cleanup_memory_entities`) against the old body. The sequences are identical apart from one multi-line `await await_maybe(...)` that is now a single line.

Object identity is preserved where it is load-bearing. `at_mem_ids` is still created in the caller, registered into `all_at_cleanup_info`, handed to `create_dynamic_redteam_job`, and stored on `PreparedTarget` as the same list object, so the cancellation path's `id(pt.memory_entity_ids)` dedupe still works. `strategy_breakdown` is mutated in place by `_cap_strategy_breakdown` rather than rebuilt. The `classification_errors` dict that `_classify_capabilities` returns is the same object the `_ctx_end_meta` calls read.

The error-handling lens found no regression: every `try` / `except` / `raise` / `logger.warning` in the extracted region matches the pre-refactor source one-to-one, and both flagged risk areas (`_cleanup_target_memory`'s accumulate-then-extend pattern and the two `CancelledError` paths) trace through to callers unchanged.

The four CI checks passed on the branch as submitted: `ruff check src`, `ruff format --check src`, `basedpyright` (0 errors), `pytest -m 'not integration'` (5470 passed, 2 skipped, 10 deselected).

## Critical (0)

None. Nothing in this diff is a correctness or safety regression.

## Important (9)

- [types, code] `runner.py:2547-2554` — `_AgentTargetJobInputs.at_backend` and `at_dyn_job` were both typed `Any`, which is a type regression this PR introduced: before the extraction the checker could see `at_backend = BareTargetBackend(at)` and the return of `create_dynamic_redteam_job(...)` directly. `PreparedTarget`, in the same file, already types the exact same two values as `Backend` and `Callable[..., Any]`. Structural fix: use those types, which are already imported. Cost: two annotations. Raised independently by two lenses, which is a confidence signal. **Fixed on this branch.**
- [types] `runner.py:2557, 2472, 2275, 2335` — `_build_agent_target_jobs` returned an anonymous `tuple[Any, list[Any], list[Any], list[Any]]` that the caller unpacks positionally into four values which then go straight into `PreparedTarget`, whose fields are concretely typed `Callable[..., Any]` and three `list[DataPoint]`. Three same-shaped `list[Any]` values unpacked by position means swapping two on a later edit type-checks clean. The same `Any`-where-`DataPoint`-is-known pattern appeared on `_generate_agent_target_datapoints`'s return, on `_estimate_static_datapoints`'s return (`load_owasp_agentic_dataset` is annotated `-> list[DataPoint]`), and on `_TargetPrepInputs.static_data`. **Fixed on this branch** by carrying the concrete types across the boundaries. The stronger version, a named result dataclass instead of a bare 4-tuple, is a question for you below.
- [types] `runner.py:2949, 3461` — `all_at_cleanup_info: list[tuple[AgentContext, list[str], Any]]` is a hand-rolled positional triple declared twice, once as a caller local and once as a dataclass field, with the third element weakened to `Any` where `Backend` is known. Nothing stops an append in the wrong order. **Partially fixed:** the third element is now `Backend`. The NamedTuple that would remove the positional unpacking is left for you, because it touches the cancellation path that #212 and #213 will rebase over.
- [minimality, code] `runner.py:3207-3242, 3151` — `_finalize_merged_report` mutated its `merged` argument in place *and* returned it, with the caller writing `merged = _finalize_merged_report(merged=merged, ...)`. The reassignment reads as if a new report were produced when none is, which is exactly the ambiguity that leads a later reader to assume the input is untouched. Structural fix: return `None` and call it as a bare statement. **Fixed on this branch**, with a one-line docstring saying it mutates in place.
- [comments] `runner.py:2910` — the `# Step 3: Prepare the first target fully — generates shared datapoints.` comment now sits above a call to `_prepare_string_targets`, which also contains the former `# Step 4: Prepare remaining string targets with shared datapoints` block (still present verbatim inside the function). The comment describes roughly half of what the call does, and the reader has no other signal that Step 4 happens in there. **Fixed on this branch** by rewording the comment to cover both steps; splitting the function so one comment maps to one call is left for you.
- [comments] `runner.py:2339` — `_prepare_string_targets` is a 108-line module-level function driven by a 30-field input bag, with no docstring. It used to sit under `_run_dynamic_or_hybrid`'s docstring and between the `# Step 3` and `# Step 4` comments, which gave a reader the phase context; extracted, it has neither, so a reader arriving from a stack trace gets thirty fields and no sentence. **Fixed on this branch** with a one-line docstring.
- [tests] `runner.py:3644-3671` — `_collect_filter_warnings` is entirely uncovered. A targeted coverage run (`pytest -m 'not integration' -q tests/redteam --cov=evaluatorq.redteam.runner --cov-report=term-missing`) reports lines 3657 and 3659-3665 as missing across 1816 passing tests, and grepping the test tree for `zero strategies selected`, `generation error:` and `no applicable strategies found` returns nothing. This function assembles user-facing warning text; if a future edit inverts the `gen_error` truthiness check or breaks the `cat_key.startswith('_')` guard, nothing fails. **Fixed on this branch** with a test per branch.
- [tests] `runner.py:3218-3242` — `_finalize_merged_report`'s two zero-result branches are an `if`/`elif` pair producing opposite warnings, and neither is asserted anywhere. This is the shape of logic a careless later edit inverts silently. **Fixed on this branch.**
- [tests] The "the suite is identical before and after" safety argument is load-bearing for the mechanical extraction and hollow as a coverage argument. `tests/redteam/test_runner.py` and its siblings drive `_run_dynamic_or_hybrid` by patching it, so they never reach the phase functions it now delegates to. Two paths remain uncovered after my additions and I did not write tests for them because they need heavier fakes than I wanted to add to someone else's PR: the `CancelledError` / `KeyboardInterrupt` cleanup block in `_execute_attacks` (lines 3491-3517, including the `id(pt.memory_entity_ids)` dedupe), and both degraded branches of `_classify_capabilities` (lines 2083-2091 and 2116-2126, the "no LLM client resolvable" and "classifier raised" paths). See the questions comment.

## Suggestions (5)

- [minimality, code, types] The nine `_*Inputs` frozen dataclasses (`_TargetPrepInputs` 30 fields, `_AgentTargetDatapointInputs` 19, `_ConfirmPayloadInputs` 19, `_AttackExecutionInputs` 13, `_AgentTargetJobInputs` 9, `_CapabilityClassificationInputs` 8, `_DynamicEstimationInputs` 8, `_TargetReportInputs` 7, `_RedTeamEvaluatorInputs` 6) each have exactly one construction site and one call site, and each is destructured field-by-field into identically-named locals on the first lines of the callee. Meanwhile five other new phase functions in the same PR (`_retrieve_agent_contexts`, `_resolve_first_agent_context`, `_estimate_static_datapoints`, `_group_results_by_target`, `_cleanup_target_memory`) take plain keyword arguments, so the file now carries two calling conventions for the same problem. Worth knowing before deciding: `PL` is not in this repo's ruff `select` list, so `PLR0913` (too-many-arguments) is not enforced — plain keyword signatures would lint clean, and no rule required the bags. This is the largest open question on the PR, so it is in the follow-up comment rather than fixed here; reversing it would rewrite most of the diff and collide with #212 and #213.
- [minimality] `_run_dynamic_or_hybrid` is still 516 lines after the split, and the phases that absorbed the most code take the most inputs (`_prepare_string_targets`: 30 in, one list out). The cyclomatic score fell because branching moved into callees, not because any branch was eliminated. That is a real gain against the linter and the PR is honest about being a mechanical split, but it is worth saying out loud that a reader reconstructing the control flow now opens nineteen functions and nine dataclasses instead of scrolling one.
- [minimality] Five new helpers have one call site and a short body with no branching worth isolating: `_ctx_end_meta`, `_resolve_first_agent_context`, `_resolve_first_caps_for_estimate`, and more arguably `_cap_strategy_breakdown` and `_group_results_by_target`. The last two carry real decision logic (the round-robin allocation loop and the job-name matching) and earn their place — I have now written tests against both, which settles it. The first three read as extractions done to shorten the caller. This is a matter of taste and I am not asking for them back.
- [minimality] `_cap_strategy_breakdown` returns its own `max_dynamic_datapoints` argument unchanged while mutating `strategy_breakdown` in place. The return value carries no information the caller did not already have. The same reasoning that applies to `_finalize_merged_report` applies here, though the case is weaker because `est_dynamic = _cap_strategy_breakdown(...)` does read naturally as a value update. I left it alone.
- [comments, pre-existing] `runner.py:2732` — the `_run_dynamic_or_hybrid` docstring still says "For each target, `_prepare_target` retrieves agent context, generates agent-specific dynamic datapoints, and produces a job closure." That was already imprecise before this PR (`_prepare_target` only ever handled string targets; `AgentTarget` objects go through `_generate_agent_target_datapoints` plus `create_dynamic_redteam_job`), so it is carried over rather than introduced. The refactor does put one more layer between the claim and the code. Worth a "For each string target" the next time the docstring is touched.

## Conflicting verdicts

- `runner.py:3654-3660` — the minimality lens called `filter_warnings.extend([<one f-string>])` inside a `for` loop a defect that should be a plain `.append(...)`. That is right about the smell and wrong about the cause: `.append(...)` in that position trips ruff's `PERF401`, which is why the `.extend([...])` is there. In the old code the same loop appended to `merged.pipeline_warnings`, an attribute, so `PERF401` never fired; the rule only started applying once the target became a local list in a fresh function. I resolved it a third way — dropping the loop in favour of `filter_warnings.extend(<f-string> for cat in unresolved)`, which satisfies `PERF401` and reads as one statement. The content and order of the produced warnings are unchanged, and the new tests pin them.

## What I changed on this branch

Two commits.

`refactor(redteam): tighten types and fix mutate-and-return in extracted phases` — `src/evaluatorq/redteam/runner.py` only, 26 insertions and 20 deletions, confined to hunks this PR introduced so the #212 / #213 rebase stays tractable:

- typed `_AgentTargetJobInputs.at_backend` as `Backend` and `at_dyn_job` as `Callable[..., Any]`;
- gave `_build_agent_target_jobs`, `_generate_agent_target_datapoints` and `_estimate_static_datapoints` concrete return types, and typed `_TargetPrepInputs.static_data` and the `at_static_dps` local as `list[DataPoint]`;
- typed the third element of `all_at_cleanup_info` as `Backend` at both declaration sites;
- made `_finalize_merged_report` return `None` and dropped the caller's no-op reassignment;
- reworded the `# Step 3` comment to cover Steps 3 and 4, and gave `_prepare_string_targets` a one-line docstring;
- replaced the `for` loop plus `extend([one])` in `_collect_filter_warnings` with a single `extend` over a generator.

`test(redteam): cover the phase helpers extracted from _run_dynamic_or_hybrid` — a new `tests/redteam/test_runner_phases.py`, 18 tests, no change to `src/`. It covers `_cap_strategy_breakdown`'s round-robin allocation and its termination conditions, `_collect_filter_warnings`' three message variants plus the two cases that must warn about nothing, `_finalize_merged_report`'s two mutually exclusive branches, `_group_results_by_target`'s exact job-name matching (including the unmatched and no-job-results cases), and `_ctx_end_meta`'s classification-error payload.

All four CI checks pass after both commits: `ruff check src`, `ruff format --check src`, `basedpyright` on the whole repo (0 errors), and `pytest -m 'not integration'` at 5488 passed, 2 skipped, 10 deselected.

## Next

1. Decide the calling-convention question in the follow-up comment. It is the one thing that changes the shape of this PR, and it is cheapest to decide before #212 and #213 rebase on top.
2. Decide whether the file-level `C901` ignore should become two function-level `# noqa: C901` suppressions, so this PR's own gain is enforced now rather than waiting on #213.
3. Decide whether the two remaining uncovered paths — the cancellation cleanup in `_execute_attacks` and the classification-failure branches in `_classify_capabilities` — get tests in this PR or a follow-up ticket.

Baukebrenninkmeijer:
@Baukebrenninkmeijer

## Questions for the author

Four things I could not decide on your behalf. Each states the finding, the options, and what I would do.

### 1. The nine `_*Inputs` dataclasses versus plain keyword arguments

**The finding.** You introduced two different ways of passing arguments to a phase function in the same PR. Nine phases take a single frozen dataclass — `_TargetPrepInputs` (30 fields), `_AgentTargetDatapointInputs` (19), `_ConfirmPayloadInputs` (19), `_AttackExecutionInputs` (13), `_AgentTargetJobInputs` (9), `_CapabilityClassificationInputs` (8), `_DynamicEstimationInputs` (8), `_TargetReportInputs` (7), `_RedTeamEvaluatorInputs` (6). Five others — `_retrieve_agent_contexts`, `_resolve_first_agent_context`, `_estimate_static_datapoints`, `_group_results_by_target`, `_cleanup_target_memory` — take plain keyword arguments. Every dataclass has exactly one construction site and one call site, is never reused, and is destructured field-by-field into identically-named locals on the first lines of the callee, so the body reads exactly as it would with a plain signature. Across the nine that is roughly 350 lines of declaration, construction and unpack that a plain `*, a, b, c` signature would not need.

**The thing that decides it.** `PL` is not in this repo's ruff `select` list in `pyproject.toml`, so `PLR0913` (too-many-arguments) is not enforced. I checked specifically because a too-many-arguments rule would have justified the bags entirely. Nothing required them.

**Options.**
- (a) Convert all nine to plain keyword-only signatures. One convention, roughly 350 fewer lines, and the diff shrinks toward being a pure move.
- (b) Keep the dataclasses but delete the unpack blocks and reference `inputs.field` at each use site. Removes about 120 lines of the boilerplate but leaves the two conventions and makes the bodies diverge textually from the original, which costs you some of the "same code, moved" review property.
- (c) Go the other way and convert the five plain-kwarg phases to dataclasses too, for one convention at the cost of more boilerplate.
- (d) Leave it, and say in the PR body that the bags are argument bundling for readability at the call site, not types expressing an invariant.

**My recommendation: (a), but not in this PR.** It is the right end state and it is the convention you already chose for five of the fourteen phases, so it is not a new idea. But it rewrites most of the diff, and #212 and #213 are stacked on this file — doing it here makes their rebase much worse than it needs to be. I would land this PR as-is, put (a) in a follow-up once the stack is merged, and in the meantime add (d)'s one sentence so a reviewer does not read `frozen=True` as buying correctness it does not buy. `frozen=True` does not freeze the `list` and `dict` members anyway.

### 2. Should `_build_agent_target_jobs` return a named result instead of a 4-tuple?

**The finding.** It returns `(at_target_job, at_dps, at_static_dps, at_all_dps)`, which the caller unpacks positionally. Three of the four are `list[DataPoint]`. I have tightened the annotations so they are no longer `Any`, but three same-typed lists unpacked by position still means that swapping two on a later edit type-checks clean and produces a run where the static and dynamic legs are silently crossed. All four values land in `PreparedTarget` a few lines later, which already has names for them.

**Options.**
- (a) Return a small frozen `_AgentTargetJobs` dataclass with the four named fields.
- (b) Return a `NamedTuple`, which keeps the existing unpacking call site working unchanged while adding names.
- (c) Leave the tuple.

**My recommendation: (b).** It is a three-line addition, the call site does not change at all, and it removes the positional hazard. I did not do it myself because it adds a tenth new type to a PR that is already being asked about having too many, and that felt like your call rather than mine.

### 3. The file-level `C901` ignore leaves this PR's own gain unenforced

**The finding.** `pyproject.toml:197` still reads `"src/evaluatorq/redteam/runner.py" = ["BLE001", "C901"]`. Your reasoning in the PR body is correct — `red_team` is at 57 and `_run_static` at 30, both untouched here, and #213 removes the entry. But the ignore is file-scoped, not function-scoped, so between now and #213 landing there is no CI signal if `_run_dynamic_or_hybrid` or any of the nineteen new functions grows back past 15. The gain this PR exists to produce is currently asserted in the PR description and nowhere else.

**Options.**
- (a) Drop `C901` from the `per-file-ignores` entry now and put `# noqa: C901` directly on `red_team` and `_run_static`. The gate re-engages immediately for every other function in the file, including the one you just fixed.
- (b) Leave it and let #213 handle it.

**My recommendation: (a), but check with the owner of #213 first.** It is the correct shape and it is two lines. The reason I did not do it is that #213 is described as owning that `pyproject.toml` entry, and if it also converts `_run_static` you would end up resolving a conflict in both files for a two-week window of enforcement. If #213 is close to landing, (b) is fine and the risk is small.

### 4. The two paths I left uncovered

**The finding.** After the 18 tests I added, a coverage run still shows two uncovered regions that a refactor slip could pass through silently.

- `_execute_attacks` lines 3491-3517: the `except (asyncio.CancelledError, KeyboardInterrupt)` block, including the `id(pt.memory_entity_ids)` dedupe against `all_at_cleanup_info`. That dedupe is the exact thing your PR body singles out as the reason `at_mem_ids` stayed in the caller, and nothing tests it. Swapping the `id(...)` set for the lists themselves would pass the whole suite.
- `_classify_capabilities` lines 2083-2091 and 2116-2126: the "no LLM client resolvable" branch and the "classifier raised" branch, the latter of which returns `AgentCapabilities(classification_failed=True)` and logs "Strategies will be included optimistically". That is a degraded path that changes which strategies get run.

**Options.**
- (a) Add both tests in this PR.
- (b) Add them in a follow-up ticket.
- (c) Leave them; they were uncovered before this PR too.

**My recommendation: (b).** They were uncovered before, so this PR does not make anything worse, and neither test is a two-liner — the cancellation one needs a fake `evaluatorq` that raises mid-flight plus two distinct memory-id lists to make the dedupe observable. But the cancellation path in particular is worth a ticket rather than a shrug: it is the one place in this file where getting it wrong leaves real attack data sitting in a customer's agent memory, and the PR body already identifies it as the fragile seam.

Baukebrenninkmeijer:
Decision for tracking: the `save=True/False` deprecation warning raised from `_resolve_save_mode` uses `stacklevel=3` (not 2) because the helper is one frame deeper than `red_team`, so the warning still points at the user's call site. Already in place with an inline comment; #213 pins it with a test.

PR #215, opened 2026-09-16
docs: state that eq redteam and eq sim register without their extras; drop the fictional CHANGELOG version

Baukebrenninkmeijer:
## Summary

Two documentation corrections surfaced while removing the Streamlit dashboards (#206).

`docs/cli-reference/overview.md`, `redteam.md` and `simulation.md` said `eq redteam` and `eq sim` are registered only when their extra is installed. They are registered unconditionally in `src/evaluatorq/cli.py` and have been since `openai` and `typer` became core dependencies; the `redteam` and `simulation` extras supply optional runtime dependencies for some commands, not the commands themselves. The three sentences now say that.

`CHANGELOG.md`'s unreleased section was headed `[1.3.0]` while the latest tag is `v1.39.0`. Releases are tag-driven (`hatch-vcs`), so the hand-written changelog cannot know the next version; the heading is now `[Unreleased]` with no version claim.

Stacked on #206 because that PR touched the same CLI reference pages.

## Test plan

- [x] `uv run --group docs mkdocs build --strict` passes


currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 5/5** — small, correct docs fix, both claims checked against actual source

Verified against head `d7106d167f` and the tip of `Baukebrenninkmeijer/remove-streamlit-ui`.

### ✅ CLI registration claim confirmed by reading cli.py directly
`src/evaluatorq/cli.py:164-165` inside `_register_subapps()` (`:154`):
```
app.add_typer(redteam_app, name='redteam', help='Red teaming commands.')
app.add_typer(sim_app, name='sim', help='Agent simulation pipeline.')
```
No conditional import guard, no `try/except ImportError`, no feature-flag `if` around either registration or the imports above them. `main()` calls `_register_subapps(app)` unconditionally. `pyproject.toml`'s core `dependencies` already include `typer` and `openai`; the `redteam`/`simulation` extras only add runtime deps (`huggingface-hub`, `vl-convert-python`, etc.), not the command registration itself. The corrected doc wording is accurate.

### ✅ Tag/CHANGELOG claim confirmed
`git tag --sort=-v:refname | head -3` → `v1.39.0, v1.38.0, v1.37.0`. The old `[1.3.0]` heading was fictional; `[Unreleased]` is correct for a tag-driven (hatch-vcs) version scheme. Diff is exactly the heading change, nothing else in that section altered — confirmed no changelog entries were dropped or reordered.

### ✅ Build
`uv run --group docs mkdocs build --strict` passes at head, no errors introduced by this diff.

### 🟡 Judgment call
This is stacked on `Baukebrenninkmeijer/remove-streamlit-ui` (#206), which is still **open**, not merged. The diff applies cleanly today, but merge order matters — merging #215 first would pull #206's Streamlit removal in with it. Land #206 first.

Baukebrenninkmeijer:
Decision for tracking: the CHANGELOG line that promised the `config=` alias 'still works in 1.3.0 and will be removed in 1.4.0' was rewritten to be version-free (e69d34e6, now 00f7aff4 after rebase), since versions come from git tags and those numbers were never guaranteed.

PR #224, opened 2026-09-20
refactor: split evaluatorq() into params, dataset and finish phases (C901 bucket 2)

Baukebrenninkmeijer:
`evaluatorq()` is the package's public evaluation entry point and was the worst single function in `src/evaluatorq/evaluatorq.py` at a ruff `C901` score of 32 against a limit of 15. This splits it along the phase boundaries the source already marked with comments, without changing behaviour, and takes the file off the `C901` debt register in `pyproject.toml`.

Two commits. The first hoists the two nested evaluation coroutines to module level — they were the bulk of the score, since a nested `def`'s branches count against its enclosing function. The second extracts the five remaining phases. Nothing moved into or out of the tracing session, the `AsyncExitStack`, `llm_concurrency_limit`, or `with_progress`; `_enter_single_trace` enters the run span on the caller's exit stack so the span's lifetime is unchanged. No log line, error message, span name or span attribute was touched, and no test assertion changed.

| Function | Before | After |
|---|---:|---:|
| `evaluatorq` | 32 | 3 |
| `_run_streaming_evaluation` (hoisted from `run_streaming_evaluation`) | — | 13 |
| `_normalise_params` | — | 5 |
| `_resolve_streaming_inputs` | — | 4 |
| `_finish_evaluation` | — | 4 |
| `_resolve_experiment_input` | — | 3 |
| `_run_in_memory_evaluation` (hoisted from `run_evaluation`) | — | 3 |
| `_enter_single_trace` | — | 2 |

The file's highest remaining score is `_run_streaming_evaluation` at 13 and `check_pass_failures` at 10, both pre-existing and both under the limit, so `"src/evaluatorq/evaluatorq.py" = ["C901"]` comes out of `[tool.ruff.lint.per-file-ignores]`. That is the only line this PR changes in `pyproject.toml`, which is shared with three sibling C901 PRs.

Checks run, all green:

```
uv run ruff check src/evaluatorq/evaluatorq.py --select C901 --config "lint.per-file-ignores={}"
uv run ruff check src
uv run basedpyright
```

The unit suite reports `5470 passed, 2 skipped, 10 deselected` both before and after the refactor, and `5471` with the regression test added below. `basedpyright` is clean across the whole repo.

## Early-return audit

Two sibling C901 splits hit the same defect: a value the original code had already recorded by the time control left a path stopped being recorded there, because a `break`/fall-through became an early `return`, or because a helper started returning a value the caller then appended only conditionally. I audited this split for it — 34 exit paths across all seven new functions (every `return`, every `raise`, every fall-off-the-end, and every exception that can propagate out of an `await`), each checked against `git show Baukebrenninkmeijer/c901-d:src/evaluatorq/evaluatorq.py` for what the original had written to a list, dict, counter, progress service, span or caller-visible local before that point. **Nothing is lost or moved.** The streaming coroutine's four exits still perform the same `total_datapoints[0] +=`, `processing_tasks.append`, `cancelled_for_fetch.add`, `progress_ref` increments, both `safe_update_progress` calls and `all_results.extend` on exactly the paths that performed them before, and its `finally` block is intact. `_normalise_params` still raises before `tracing_session(...)` opens, so a rejected call still produces no trace. `_finish_evaluation`'s sink append still mutates the caller's list — the frozen dataclass freezes the field binding, not the list object — under both unchanged guards. `validated`, `run_span` and `upload_response` became helper-locals and are provably never read in `evaluatorq()` afterwards.

The audit did turn up a **test** gap, now closed by the third commit. `dataset_id` is resolved inside the streaming branch and read after it, at upload time, making it the one value in `evaluatorq()` that crosses a branch boundary — and it had no coverage at all, because every streaming test runs with `_send_results=False` and every `send_results` test passes `dataset_id=None`. The `_experiment_url_out` sink had the same gap on the streaming path. A change that stopped propagating either would have shipped silently. The new test fails if either write is dropped; I verified that by injecting both mutations and confirming red, then restoring and confirming green.

Two rulings worth flagging for the reviewer:

The streaming phase takes a frozen `_StreamingInputs` dataclass rather than eight positional parameters, and the params phase takes `_EvaluationInputs` and returns `_ResolvedEvaluationInputs` for the same reason. That follows the plan's rule for phases needing more than about six inputs. The cost if it is the wrong call is verbosity, not behaviour — the alternative was an eight-argument call site.

The new `TYPE_CHECKING` import of `TracingContext` is relative (`from .tracing.context import ...`), matching the ten relative imports already at the top of this file and the relative import already inside `evaluatorq()`'s single-trace branch. `CONTRIBUTING.md` asks for absolute imports; converting only the new line would make the file internally inconsistent, and converting all eleven is drive-by reformatting in a PR whose whole point is that nothing else changed. Say the word and I will convert the file's imports in a separate commit.

Based on `Baukebrenninkmeijer/c901-d` rather than `Baukebrenninkmeijer/cicd-complexity-checks`, because that PR also edits this file and stacking avoids the conflict. Merge it first.


Baukebrenninkmeijer:
Replaces #208 (native-stack mergeability stuck after its stack base was deleted). Same branch and commits; #208 carried an approval.

Baukebrenninkmeijer:
Decision for tracking, carried over from #208: the seven helpers extracted from `evaluatorq()` got docstrings in 3a9ad517 to match the docstring rule D's reviewer set, and the one relative `TYPE_CHECKING` import in `_enter_single_trace` became absolute per CONTRIBUTING.md in b6ca56a8. Both commits are in this branch after the rebase onto the new c901-d.

currentlycodinng (approved):
## Verdict: APPROVE (re-review of the reopened #208)

**Rating: 5/5** on the code. Two small notes below, and two corrections to numbers in my own earlier review.

This is the reopen of #208, which I approved at `b14d587041`. That was commit 2 of 5, so I re-verified everything that landed after it against head `b9e2aaf6a2` and base `Baukebrenninkmeijer/c901-d` at `9845e14e12` (the tip of #223).

### What changed since my approval
`git range-diff` between the old and new series: commits 1, 3, 4, 5 are identical patches, commit 2 differs only in pyproject.toml context lines from the rebase. `src/evaluatorq/evaluatorq.py` and `tests/unit/test_evaluatorq.py` are byte-identical between #208's final head `cd266a08b4` and this head. Three commits landed after my approval:
- `d25195bd` adds the streaming `dataset_id` / `_experiment_url_out` regression test at `tests/unit/test_evaluatorq.py:656-712`. Mutation-tested: forcing `dataset_id=None` at `src/evaluatorq/evaluatorq.py:851` fails it with `assert None == 'ds_42'`, and deleting the sink append at `:478` fails it plus `tests/test_evaluatorq_upload_sink.py:55`. Not a vacuous test.
- `d00117e5` is +87 lines of docstrings. The AST hash with docstrings stripped is unchanged across the three commits, so zero code change.
- `b9e2aaf6` (head) is one line, `evaluatorq.py:334`: `from .tracing.spans import` became `from evaluatorq.tracing.spans import`.

### Corrections to my earlier review
- I wrote "32 → 15" for `evaluatorq()`. That was wrong: 15 is the threshold. Real numbers, run with `--isolated --select C901 --config 'lint.mccabe.max-complexity = 15' --config 'lint.per-file-ignores = {}'`: base reports `evaluatorq` at 32, head reports `All checks passed!`. At max-complexity 1 the head values are `evaluatorq` 3, `_normalise_params` 5, `_finish_evaluation` 4, `_run_streaming_evaluation` 13 (the highest). The conclusion stands, and the `["C901"]` entry for this file at `pyproject.toml:171` on base is gone at head, which is justified since nothing in the file is above 15.
- I wrote "5454 passed" with one pre-existing live-catalogue failure. On this base the suite is fully green: 5488 passed / 4 skipped on base, 5489 / 4 on head (the +1 is the new regression test), zero failures either side.

### ✅ Behaviour preservation
- The public signature of `evaluatorq()` (`:668` at head, `:185` on base) is identical across 25 AST facets: 17 params, every annotation and default, keyword-only markers, return annotation, decorators and the 3814-char docstring.
- 45 scenarios driven through the public entry point on base and head (list / dict-row / awaitable / empty / streaming `DatasetIdInput` datasets, jobs raising, evaluators raising, all parallelism options and the deprecated `parallelism` alias, `print_results`, `path`, send-to-orq with a fake client including upload raising, `ExperimentInput`, seven invalid-param cases). Compared returned objects, full stdout, logging, warnings, exception type plus MRO plus message, and an ordered trace of every job, scorer, fetch and upload call. 44 of 45 byte-identical. The one difference is a Rich spinner glyph that also varies between two runs of base alone.
- `raise` statements (9) and `except` handlers (4) are identical in `evaluatorq.py`. No public names added or removed.
- One observable difference outside the captured output: the loguru source-location field moves because the emitting function moved (`evaluatorq:316` becomes `_normalise_params:288`, `evaluatorq:391` becomes `_resolve_streaming_inputs:415`). Message text is identical. The PR body's "no log line was touched" is true of the text and not of that field. Cosmetic, but anything parsing it will shift.

### 🔒 Checks
`basedpyright` 0 errors. `ruff format --check src` clean. `ruff check src` shows 3 errors on base and on head, identical and all BLE001 at `tracing/setup.py:255, 298, 323`. That is the known issue fixed by #226, and it is the only reason the Lint job is red here. All 15 other checks pass.

### Still present from my first review (nit)
`_normalise_params` unpacks 11 inputs at `:244-253`, and six of them (`datapoint_parallelism`, `llm_parallelism`, `print_results`, `description`, `path`, `single_trace`) are unconditionally overwritten from `validated` at `:293-299`, so they only matter in the `elif` branch. `jobs` is also cast twice, at `:284` and `:303`. Harmless.

### 🟡 Judgment calls
- **Mixed import style, one line in this diff caused it.** Base had a relative import at `evaluatorq.py:335`, and the last commit converted just that one to absolute. The PR body argues against exactly this ("converting only the new line would make the file internally inconsistent"), and the TYPE_CHECKING import at `:39` plus ten others at `:12-25` are still relative. `CONTRIBUTING.md:73` asks for absolute imports. Either convert the whole file in its own commit or revert `:334`. Your call.
- **PR body is stale.** It quotes `5470 passed, 2 skipped` (real: 5488 / 4 skipped) and lists `uv run ruff check src` under "all green", which is red for the three BLE001 findings above until #226 lands.
- **Merge order:** the base is #223's branch, so #223 has to land first.

### 🔴 Needs live testing
Every upload and dataset-fetch path was exercised with fakes, and `tests/conftest.py` blocks outbound calls in unit tests. One real streaming run against a live Orq dataset with `_send_results=True` is the only thing not proven end to end. `send_results_to_orq` receives byte-identical arguments in both trees, so I rate the risk low, but it is worth doing once on the stack before it merges.

PR #225, opened 2026-09-20
refactor(redteam): name the attack error fields and enforce the turn outcome invariant

Baukebrenninkmeijer:
Two reviewer commits — `26f7fe1c` and `dd447749` — were dropped from #209 by a rebase that ran from a stale local checkout, and #209 merged without them. This PR lands them on top of the merged result, unchanged. Both cherry-picked onto `main` with no conflicts.

`26f7fe1c` replaces the positional `tuple[str, str, str, str, dict, int]` that carried the six error fields between `_attacker_turn` and `run_attack` with a frozen `_AttackError` of six named fields. Four of the six are `str`, so a transposition at any of the four construction sites produced no type error and no test failure — only a report whose error classification was quietly wrong. `_AttackerTurn.error_fields` becomes `failure`, and `__post_init__` now rejects a value that sets both `failure` and `retry`, the mutual exclusion the docstring previously only described. The ten helpers extracted by #209 gain docstrings. No behaviour change: same fields, same values, same order.

`dd447749` pins the two regressions #209 introduced and fixed but left unpinned — each could be reintroduced with the full suite staying green. `test_adversarial_empty_max_tokens` now asserts `truncated_turns`, so the truncation append cannot move below the early break. A new test drives `run_attack` with an empty `choices` list and asserts the clean abort still counts the billed attempt, so a dropped usage fold cannot report the attack as costing nothing. Both branches of `_unresolved_timeout_error` are covered too; no test reached either before.

Checks run verbatim on the branch tip: `ruff check src` clean, `ruff format --check src` clean, `basedpyright` (unscoped) 0 errors, `pytest -m 'not integration'` 5473 passed / 2 skipped.


currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 5/5** - a faithful re-landing of the two dropped commits, with a behaviour differential and mutation checks that all come out clean.

### ✅ Provenance: the commits are what the description says
I compared the added and removed lines of `e6ce9638` and `e0d22fcc` against the diffs of the original `26f7fe1c` and `dd447749` (242 and 78 changed lines respectively). They are identical line for line, so "unchanged" is accurate. The branch applies to `main` at 838c42ea.

### ✅ "No behaviour change" holds
I ran a mocked `run_attack` harness on base and head over 12 scenarios: happy path, objective achieved, timeout then retry, consecutive-timeout give-up, content-filter exception, generic exception, empty choices, empty `max_tokens`, empty stop, truncated but usable, and unresolved timeout with and without an earlier turn. After removing wall-clock durations and a stringified `model_fields_set` (a set printed in hash order, so it differs between any two runs), all 12 results are identical between base and head, including error text, error codes, error details, turn records and token usage.

### ✅ The new tests catch what they claim to
I made four mutants of head and ran `tests/redteam/test_orchestrator_coverage.py` and `test_adversarial_timeout_budget.py` (57 pass on head):
- Truncation append moved below the early break: `test_adversarial_empty_max_tokens` fails.
- Usage fold dropped on the empty-choices path: `test_adversarial_empty_choices_still_bills_the_attempt` fails.
- `_unresolved_timeout_error` guard forced off: both unresolved-timeout tests fail.
- Its `if not turns_record` branch forced true: `test_unresolved_timeout_after_a_turn_warns_but_keeps_the_transcript` fails.
Every mutation is killed by the test written for it. Full `pytest -m 'not integration'` at head: 5490 passed, 4 skipped, 10 deselected.

### 🟡 Smaller notes
- The description's "5473 passed" is out of date; I get 5490 at this head. The number reflects how far main moved, not a problem.
- `_AttackerTurn.__post_init__` (`src/evaluatorq/redteam/adaptive/orchestrator.py:469-471`) raises if both `failure` and `retry` are set, but I found no test that constructs that state, so the invariant itself is unpinned. Also, the `retry=True` construction at :890 sits inside a `try` whose handler at :895 is `except Exception`, so a future bug that set both would be caught there and reported as a generic adversarial LLM failure instead of surfacing as a `ValueError`. No current site sets both; I checked the constructions at :872, :890, :911, :962, :998 and :1016.
- `ruff check src` shows the three `BLE001` findings in `tracing/setup.py` that main already has; #226 fixes them, this PR does not touch them.

### 🔴 Needs live testing
None for this change. It is a refactor plus tests, and I did not run a live red-team attack against a real target or attacker model. The `open-code-review` CI check, if red, is mechanical and not a code finding.

PR #226, opened 2026-09-20
chore: restore the BLE001 register line for tracing/setup.py dropped in #204

Baukebrenninkmeijer:
#204 merged with the `tracing/setup.py` per-file-ignores line missing (a rebase conflict resolver dropped it instead of keeping the `["BLE001"]` version), so `ruff check src` on main now reports three BLE001 findings in that file and CI on main is red. This puts the line back. One line, no behaviour change.


currentlycodinng (approved):
## Verdict: APPROVE

**Rating: 5/5** — one-line fix, every claim in the body reproduced by running it

Verified against head `786cf3beb8` and main tip `838c42eaef`.

### ✅ Main is red for exactly the stated reason
At main `838c42eaef`, `uv run ruff check src` (ruff 0.15.0, the `uv.lock` pin) reports:
```
BLE001 ... src/evaluatorq/tracing/setup.py:255:12
BLE001 ... src/evaluatorq/tracing/setup.py:298:12
BLE001 ... src/evaluatorq/tracing/setup.py:323:16
Found 3 errors.
```
Nothing else. `ruff format --check src` is clean. At this PR's head the same command prints `All checks passed!`. The Lint job on main's tip failed with those same three lines; the typecheck and test jobs on 3.10 to 3.13 are green, so this is the only thing keeping main red.

### ✅ The line was dropped by accident in #204
`gh pr diff 204` removes `"src/evaluatorq/tracing/setup.py" = ["BLE001"]` along with eight C901-only entries. The eight match #204's stated intent (files that carried C901 alone). This ninth one is a BLE001 exemption for a file #204 never touched and never listed in its table, so the drop was unintentional. The restored line at `pyproject.toml:199` is byte-identical to the pre-#204 version and sits in the right place in the sorted register. No other entries were removed or altered by #204.

### ✅ Register is exactly minimal at head
With the PR applied there are 53 entries and all 53 are triggered (`ruff check src --select BLE001,C901 --config 'lint.per-file-ignores={}'`), so nothing stale and nothing missing.

### 🟡 Judgment call: the same line is being added by two other open PRs
#205 and #220 both add this exact line too (each diff has one `+` line for `tracing/setup.py`). Whichever lands first turns the others into a conflict or a no-op. Suggest landing this one first, since it is the only one where the change is the whole point, and having #205 and #220 drop the line on rebase.

Nothing here needs live testing.

